PR #102, opened 2026-04-10
Add end-to-end blueprint validation test harness

storkme:
## Summary
- Adds `crates/core/tests/e2e.rs` — integration tests that run the full Rust pipeline (solve → layout → validate → analyze → export → parse) and assert zero validation errors + production throughput ≥ target rate
- Runs as a separate test binary via `cargo test --test e2e` (standard Rust integration test pattern)
- 3 tests passing (tier 1-3), 3 ignored pending fixes (validator perf on large layouts, #64)

## Test plan
- [x] `cargo test --test e2e` passes tiers 1-3 with zero errors and correct throughput
- [x] `cargo test` still runs all unit tests normally
- [ ] Fix validation hangs on large "from-ore" layouts (follow-up)
- [ ] Unblock tier 4 by resolving lane-throughput bottleneck (#64)

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,6 +1,8 @@
 [profile.ci]
 failure-output = "immediate"
 fail-fast = false
+slow-timeout = { period = "30s", terminate-after = 3 }
+test-threads = 1
 
 [profile.ci.junit]
 path = "junit.xml"
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,7 @@
+[profile.ci]
+failure-output = "immediate"
+fail-fast = false
+
+[profile.ci.junit]
+path = "junit.xml"
+store-success-output = true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,8 +57,50 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
-      - run: cargo test -p fucktorio_core
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-nextest
+      - run: cargo nextest run -p fucktorio_core --profile ci
       - run: cargo clippy -p fucktorio_core -- -D warnings
+      - name: Generate test summary
+        if: always()
+        run: |
+          python3 - <<'PYEOF'
+          import xml.etree.ElementTree as ET, os
+          path = "target/nextest/ci/junit.xml"
+          if not os.path.exists(path):
+              print("No JUnit XML found"); exit(0)
+          tree = ET.parse(path)
+          root = tree.getroot()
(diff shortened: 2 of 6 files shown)


storkme:
Good work on the test harness and the cycle guard — the retry-based approach is the right shape. A couple of things need addressing before merge:

## Must fix

**1. Potential panic in `belt_flow.rs` — `lane_rates[&sib]` after retry exhaustion**

In the splitter fall-through when a tile exhausts its retry budget, the code does:
```rust
let total_l = lane_rates[&pos][0] + lane_rates[&sib][0];
```
`lane_rates` is indexed with `[&sib]`, which panics if `sib` is absent from the map. `sib` is only guaranteed to be present if it appears in `belt_dir_map` — if the splitter sibling tile is missing due to a layout bug upstream, this crashes. `belt_structural.rs` correctly uses `.get().copied().unwrap_or((0.0, 0.0))` for the same case; `belt_flow.rs` is inconsistent and unsafe here.

**2. Retry-exhausted splitter should `continue`, not fall through**

When a tile exhausts its retry budget waiting for its splitter sibling, the fall-through executes the sibling-averaging logic with potentially stale/zero rates from `sib`. This produces silently wrong throughput numbers that can cause false-negative validation results (missed errors). The correct behaviour on retry exhaustion is to `continue` — mark the tile processed with zero/current rates and move on, rather than computing bad averages. Both `belt_flow.rs` and `belt_structural.rs` have this issue.

## Should fix

**3. Shared `retries` counter conflates UG and splitter wait reasons**

The `retries` map is keyed on `pos` only, so a tile that burns through its retry budget waiting for a UG pair will have no budget left when it hits the splitter branch on the next visit (and vice versa). Consider separate counters `ug_retries` and `splitter_retries` keyed on `pos`.

## Minor

**`e2e.rs` — `assert_round_trip` is a weak structural check**

Entity count alone won't catch a dropped entity + duplicated entity. Worth a doc-comment acknowledging this is intentionally weak due to the known multi-tile offset bug, and ideally a follow-up issue for the offset bug so it doesn't get forgotten.

**`e2e.rs` — `diag_validator_timing_from_ore` uses `.unwrap()`**

It's `#[ignore]`-d so low risk, but `expect("message")` costs nothing and gives a much better failure message if it ever gets un-ignored.

**`ci.yml` — Python summary script uses `open()`/`close()` instead of `with`**

If an exception fires between `open` and `close` (e.g. malformed JUnit XML) the file handle leaks. `with open(...) as f:` is the safe pattern.

**`ci.yml` — no per-test timeout in nextest profile**

Nextest's 60s default will catch a hung test, but for this project's e2e tests an explicit `[profile.default] test-threads = 1` and a tighter timeout in `.config/nextest.toml` would make CI behaviour more predictable.

PR #160, opened 2026-04-21
feat(bus): pack lane columns by sharing non-overlapping y-ranges

storkme:
## Summary

- Replaces the sequential `lane.x = (i+1)` column assignment with greedy interval bin-packing: two lanes whose active y-ranges don't overlap are assigned the same x column
- Family balancer blocks are treated as a unit and assigned M contiguous columns collectively
- `bus_width_for_lanes` now derives from `max(lane.x) + 2` rather than `lanes.len() + 2`
- `active_range` promoted from inner function to `pub(crate)` in `lane_order.rs` so the packer can reuse it

Effective y-range for packing = union of `active_range` and `family_balancer_range`, so balancer blocks can't be overlapped by a co-tenant lane.

## Test plan

- [ ] `cargo test --manifest-path crates/core/Cargo.toml` — all 375 tests pass
- [ ] Load `electronic-circuit` and `advanced-circuit` in the dev server; confirm bus is visually narrower where lanes with disjoint y-ranges existed, and balancer family blocks still stamp correctly
- [ ] Confirm trace overlay lane columns light up at the right x positions

### DIFF
--- a/crates/core/src/bus/lane_order.rs
+++ b/crates/core/src/bus/lane_order.rs
@@ -10,37 +10,39 @@ use rustc_hash::FxHashMap;
 use crate::bus::lane_planner::BusLane;
 use crate::bus::placer::RowSpan;
 
+/// The y-range a lane is active over: from the earliest producer output or
+/// source_y down to the deepest tap-off or producer output.
+pub(crate) fn active_range(lane: &BusLane, row_spans: &[RowSpan]) -> (i32, i32) {
+    let all_p = lane.all_producers();
+    if !all_p.is_empty() && !lane.consumer_rows.is_empty() {
+        let start = all_p.iter()
+            .map(|&p| row_spans[p].output_belt_y)
+            .min()
+            .unwrap();
+        let end = if !lane.tap_off_ys.is_empty() {
+            lane.tap_off_ys.iter().copied().max().unwrap()
+        } else {
+            start
+        };
+        (start, end)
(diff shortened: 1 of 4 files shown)


storkme:
Closing unmerged. Post-mortem in [#168](https://github.com/storkme/fucktorio/pull/168) → `docs/archive/lane-column-packing-investigation.md` once merged.

Tl;dr:
- The web commits (`3ad0026`, `752f4f7`) are already on main via the squash-merged #159, and have been substantially rewritten since by `d845e56` and #166. Merging as-is conflicts badly.
- The lane-packer commit (`e1b6da9`) has two bugs: (a) `active_range` doesn't fold in `balancer_y + 1` so lane footprints are truncated; (b) `family_balancer_range` is assigned AFTER `pack_lane_columns` runs, so the packer's family-aware check reads `None`. Both fixable locally.
- But fixing them still fails \`tier2_electronic_circuit_from_ore\` — the packer moves the copper-cable family from cols 3-5 to cols 2-4, which causes its balancer feeder to cross iron-ore at col 1 y=39-44. The ghost router's junction solver doesn't detect that as a crossing because its geometric patterns assume sequential column layout. Iron-ore stays on surface at (1,38) facing south → copper-cable at (1,39). Belt-item isolation violated.

The bus-width-compaction opportunity is real (~30% in the tier2-from-ore snapshot) but the next step is rethinking how the router discovers crossings, not this packer.

Issue #136, opened 2026-04-11
Missing 5→9 balancer template (and other coprime shapes) crashes layout


storkme:
## Repro

http://localhost:5173/?item=advanced-circuit&rate=5&machine=assembling-machine-2&in=coal%2Cwater%2Ccrude-oil%2Ciron-ore%2Ccopper-ore&belt=transport-belt

Fails in the browser with:

> No 5→9 balancer template for copper-cable; producer outputs are disconnected

## Root cause

`crates/core/src/bus/balancer_library.rs` has N→M templates for many shapes but not 5→9. The decomposition fallback in `compute_extra_gaps`/`stamp_family_balancer` only works when `n` and `m` share a divisor ≥ 2:

```rust
(1..=n).rev().find_map(|g| {
    if n % g == 0 && m % g == 0 { templates.get(&(n / g, m / g)).map(...) }
    ...
})
```

`gcd(5, 9) = 1`, so no decomposition is possible and stamping fails.

## Affected shapes (spot check)

From `grep '^static T_[5-9]_.*_OUTPUT'`, coprime gaps include at least: 5→7, 5→8, 5→9, 6→7 is present, 7→9, 8→9. Any of these can appear depending on recipe + rate + machine-tier combos.

## Fix options

1. **Generate the missing templates** via `scripts/generate_balancer_library.py` — straightforward but requires Factorio-SAT time and binary bloat.
2. **Pad to a solvable shape** — if 5→9 has no template, stamp the smallest `(n', m')` with `n' ≥ n`, `m' ≥ m` that exists, and leave the extra input/output lanes unconnected. Currently no library shape dominates 5→9 (no 5→10, no 6→9, no 6→10 in the current library), so option 2 alone isn't enough — would still need new templates.
3. **Over-generate**: when picking the template set, ensure every `(n, m)` for `n, m ≤ 10` has a dominating entry.

## Context

Found during the SAT-driven layout compaction investigation (stale plan `joyful-wibbling-lemur`). Unblocks tier-4 from-ore cases beyond just the fluid-row [#68](https://github.com/storkme/fucktorio/issues/68) blocker.

PR #178, opened 2026-04-24
RFC: Modular production with demand-partitioning and subtree decomposition

storkme:
## Intent

This RFP proposes two optional, composable strategies to break the 8-lane balancer ceiling that currently caps intermediate production width. The concrete failing case is `advanced_circuit` at moderate-to-high rates, which needs more copper cable than any single 8-lane balancer can feed ([#136](https://github.com/storkme/fucktorio/issues/136)).

The proposal reframes the bus from a shared resource pool to a routing fabric for locally-scoped producer→consumer pairs, enabling demand-partitioning (one producer module per consumer, sized to exact demand) and subtree decomposition (sharding wide modules into proportional pieces when recipes exceed 8 lanes).

## Scope

- **In:** Complete RFP document with design, kill criteria, verification plan, phasing, and task tracker
  - Phase 0: Scaffolding to support multiple `LaneFamily` per item via `module_id` keying
  - Phase 1: Demand-partitioning outer pass (one module per consumer)
  - Phase 2: Subtree decomposition inner pass (shard wide modules)
  - Both features gated behind off-by-default flags; current behavior is the baseline
  - Explicit kill criteria for each phase to prevent scope creep
  - Living task tracker and decision log for progress tracking

- **Out:** No implementation code in this PR — this is the design document only
  - Actual code changes land in three follow-up PRs (Phase 0, Phase 1, Phase 2)
  - Rejected alternatives documented but not pursued
  - Consumer lumpiness edge case deferred to Phase 2 decision point

## Verification

N/A — this is a design document, not code. The RFP itself is the artifact for review and discussion.

## Deviations from intent

None. This is the brainstorm-to-RFP conversion as discussed.

## Models / contracts touched

This affects the lane planner abstraction (`lane_planner.rs`, `lane_order.rs`) and the balancer stamping contract (`balancer.rs`, `templates.rs`). The semantic shift from "shared pool" to "routed pairs" is a real architectural change, but it's captured in the design section and will be reflected in contract updates when Phase 0 lands.

https://claude.ai/code/session_01TWQMAqFBLjpmxub1UQK7iH

### DIFF
--- a/docs/rfp-modular-production.md
+++ b/docs/rfp-modular-production.md
@@ -1,0 +1,326 @@
+# RFP: Modular production — demand-partitioned modules + subtree decomposition
+
+> **Status**: Draft, not yet accepted. Living document — update the
+> Decision log and Task tracker as work progresses. Kill-criteria are
+> explicit and meant to be acted on.
+
+## Current status
+
+- **Phase**: Draft / brainstorming complete.
+- **Flag state**: both features gated behind off-by-default flags;
+  current behavior is the no-flags path.
+- **Last update**: 2026-04-24 — initial draft from brainstorm.
+
+## Summary
+
+The current layout engine pools each intermediate item into one shared
+bus lane family and stamps a single balancer at the producer row
+output. This caps the widest producible intermediate at 8 lanes
+(the largest template in `balancer_library.rs`). We propose two
+*optional, composable* strategies, layered behind feature flags:
+
+1. **Demand-partitioning (outer)** — split a high-demand intermediate
+   into one producer module per consumer, sized to that consumer's
+   exact demand, wired directly to it. No shared pool, no
+   pool-balancer, and the widest balancer ever needed is bounded by
+   the *largest single consumer's* demand, not the sum.
+2. **Subtree decomposition (inner)** — if a single demand-partitioned
+   module is still wider than 8 lanes (big consumer, or recipe ratios
+   expand going up-tree), shard *that* module into ⌈widest / 8⌉
+   sibling subtrees, each producing a proportional share.
+
+Both flags default to off. The existing pooled/balancer-stamped path
+stays the baseline until one or both strategies earn the right to
+become default via the kill-criteria checks below.
+
+## Motivation
+
+Concrete failing case: `advanced_circuit` at moderate-to-high rates
+needs more copper cable than any 8-lane balancer can feed.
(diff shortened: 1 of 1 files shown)


storkme:
Superseded by #179.

Closing in favour of the revised RFP there. The original proposal's one commit (`f07b01b`) is cherry-picked onto #179's branch as-is (preserving authorship), and a follow-up commit on top captures the revisions from a second-pass review. Key changes vs. this PR:

- **Strategy enum instead of two booleans.** `demand_partition` + `subtree_decompose` → `LayoutStrategy::{Pooled, PartitionedPerConsumer, PartitionedDecomposed}`. Avoids invalid combinations (`decompose && !partition`) and gives escargio (folded layouts, upcoming) a clean extension point.
- **Strategies are peer citizens, not trust-earning degradations.** The layout engine respects what the user selected; no silent auto-downgrade when a case can't satisfy a strategy — instead, loud diagnostic output. Rationale in the PR body and decision log: different recipes and different users prefer different shapes, and silently picking the "safe" one hides that tradeoff.
- **K1-1 retired, K1-4 added.** The "entity count < 1.5× on uninvolved cases" criterion was framing entity-count blowup as a bug; it isn't, it's a tradeoff the user opted into. K1-4 replaces it with an inertness test: partitioning on single-consumer cases must produce entity-equivalent output to pooled. Entity count per strategy becomes an observable in the scoreboards, not a gate.
- **"Belts over-provisioned" made numeric.** 75% of belt-tier capacity as the utilization ceiling. Dedicated `PartitionRejectedByUtilization` trace event; K1-3 gates the false-positive rate of the threshold itself.
- **Consumer defined explicitly** as one consuming recipe-row — not one machine, not one item. Undefined in the original.
- **"Simple tee" walked back.** The shared-upstream distributor is a small-scale balancer reusing `balancer_library.rs` — the "no pool-balancer" property holds at the output side, not the input side. Not a new primitive, but the RFP no longer pretends it's free.
- **Fluids explicitly out of scope** for partitioning (pipe networks merge freely, no per-lane identity to partition).
- **Hierarchical-composition prototype pulled into Phase 1** to de-risk the K2-2 fallback: if decomposition produces too many lumpy taps, we pivot to working code rather than a hypothetical backup.
- **Phase 0 split into 0a (core, ≤400 LOC) and 0b (surface wiring, ≤200 LOC)** for reviewability.

The full revision history is in the Decision log section of the doc on #179. Happy to discuss any of the above in the #179 thread.

---

PR #213, opened 2026-04-25
docs(rfp): render static layout via ParticleContainer + texture atlas

storkme:
## Intent

RFP for the rendering refactor following the perf investigation in
\`docs/web-render-perf-investigation.md\` and the trace evidence
gathered against PR #211. Proposes the architectural target for what
follows the small/medium fixes already landed (#205 render-on-demand,
#206 streaming-of-silent-stamps, #211 dedup, #212 hit-test).

The core observation: post-PR-#211, steady-state render cost is
~25-30ms/frame, dominated by GL state changes between many small
batches (1207ms \`uniformMatrix3fv\` + 670ms \`bindVertexArray\` +
338ms \`useProgram\` vs 826ms actual \`drawElements\` over 9.7s of
trace). Render-group instruction cache is hot; the playback is
expensive because of heterogeneous Graphics/Sprite/Text content per
entity. Going below ~10ms/frame requires homogenising entities to
something that batches cleanly.

Recommendation: Pixi v8 \`ParticleContainer\` reading from a
runtime-built texture atlas, swapped in at \`finish(layout)\`. Streaming
phase keeps Graphics (not the bottleneck).

## Scope

- **In:** RFP document at \`docs/rfp-renderer-particle-container.md\`.
- **Out:** Implementation. The RFP includes a 4-phase shipping plan
  with explicit kill criteria so we can abandon the approach if a
  prototype underperforms (≥2× speedup or it's wrong). No code
  changes here.

## Verification

- [x] \`npx tsc --noEmit\` clean (doc-only PR).
- [N/A] Browser eyeball — no behavior change.

This PR is for **review of the RFP itself**, not an implementation
checkpoint. Looking for sign-off (or push-back) on:

1. The hybrid streaming-Graphics / settled-Particles split.
2. The choice of ParticleContainer over tilemap (rationale in the
   "Alternatives considered" section).
3. The 4-phase shipping plan and kill criteria.
4. The "Out of scope" exclusions, especially keeping recipe-label
   rework out of Phase 1.

## Deviations from intent

The project conventions (\`docs/rfp-template.md\`) call out kill
criteria as the most important section. I've written four observable
ones — happy to tighten if any feel too soft.

## Models / contracts touched

None. Doc-only.

### DIFF
--- a/docs/rfp-renderer-particle-container.md
+++ b/docs/rfp-renderer-particle-container.md
@@ -99,21 +99,29 @@ Generated at runtime, lazily, on first need per `(entity, variant)`:
   | PTG | direction | 4 |
   | Inserter | type × direction | 4 × 4 = 16 |
   | Machine | per-name | 21 (from `MACHINE_SLUGS`) |
+  | Item icon | per-item (see below) | ~80–120 |
 
-  Total upper bound: ~150 unique textures. Comfortably fits on one
+  Total upper bound: ~250 unique textures. Comfortably fits on one
   GPU texture (4096×4096 with 64 px tiles holds 4096 entries).
 
-- Item-color tinting (used today for belt body coloring per
-  `carries`) does **not** require pre-rendering each item × belt
-  combination. Pixi's `Particle.tint` is a per-particle multiplier
-  applied in shader; we render the belt texture white-on-black and
-  tint to the item color at draw time. Same for inserter colors.
+- Item-color tinting on belts (used today for belt body coloring per
+  `carries`) is **dropped** — item identity is conveyed by the icon
+  particle layer (see "Item icons on every belt / pipe / inserter"
+  below). Belt textures stay neutral; the tint slot is reserved for
+  hover-dim multipliers.
+
+- Item icons: one atlas entry per `(name, kind)` pair where `name` is
+  the item slug. Loaded from existing `icons/${slug}.png` PNGs
+  (cached via `Assets.load` in `preloadCarriesIcons`). Items without
+  icons (fluids, currently) use a generated placeholder texture
+  (colored circle from `itemColor(item)`). See "Item icons on every
+  belt / pipe / inserter" below.
 
 - Atlas packing: Pixi v8 has no first-party runtime atlas packer. We
   pre-allocate a single oversized `RenderTexture` and `blit` each
   variant into a deterministic grid slot (`textureKey →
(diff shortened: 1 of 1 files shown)


storkme:
Pushed RFP revisions per discussion. Three folded-in changes:

1. **Streaming and settled use the same `ParticleContainer`** — pan during the 16 s routing phase is as fast as pan after settle. `finish()` no longer swaps anything, just freezes the ticker.
2. **Scrubber compatibility** documented. `seekTo`'s `reveals` array maps 1:1 to particles (`r.particle.alpha` instead of `r.graphic.alpha`); same loop, same source map. Dead `eventMode` line in `applyReveals` flagged for removal.
3. **Debug-mode animation logging** added. Optional `console.log` gated on `debugCb`, surfaced via `globalThis.__ANIM_LOGS` (mirroring existing `__TRACE_LOGS`). Per-batch log at every animation-start site.

Phasing re-shaped: Phase 1 is atlas plumbing + trunk particles (kill criterion #1 measured against **streaming-pan**, the more demanding case). Phase 2 extends to all entity types. Phase 3 wires highlight controller. Phase 4 adds the debug-logging helper.

Added a fifth kill criterion: scrub-mode framerate must stay within 10% of pre-RFP baseline.

Diff: 219 insertions / 88 deletions, single commit on top of the original.

storkme:
Pushed another revision (115+/28-): icon-on-every-conveyor-tile design folded in, machine recipe panels removed.

**Key changes:**

- Every belt, pipe, and inserter tile gets an item-icon particle at tile center. Icons live in their own \`ParticleContainer\` so they batch independently.
- Atlas grows by one entry per item slug (~80-120 items), not item × belt-variant combos. Total atlas now ~250 entries.
- The \`(ex + ey) % 5 === 0\` icon-spacing heuristic in \`entities.ts:1201\` is removed.
- The "tint belt body by item color" trick is dropped — belts stay neutral, item identity comes from the icon layer.
- Machine recipe label panels (\`drawMachine\`, \`entities.ts:740-810\`) are removed entirely. Recipe info flows through icon-on-inserter (and icon-on-pipe for fluids). Rate detail / richer recipe info deferred to a future hover-tooltip enhancement.
- Fluid icons placeholder: \`getItemIconTexture\` returns a colored circle from \`itemColor(item)\` when the PNG isn't cached. Real fluid icons tracked in #214.

PR #233, opened 2026-04-26
fix(web): scope input labels to top-row trunks; tighten font scaling

storkme:
## Summary

Two fixes to the external-input trunk labels added yesterday (visible problem in `misc/Untitled.png`):

- **Scope to actual external-input trunks.** The label finder picked the topmost trunk-bearing tile per column, even when that "top" was mid-layout — so a stray iron-plate segment deep in the bus was getting an oversized "Iron Plate" label hovering over nothing in particular. Now only columns whose topmost matching tile sits at `y == 0` qualify, which matches `BusLane.source_y = 0` for every external input (`lane_planner.rs:198`).
- **Tighten font scaling.** The previous `clamp(span * 16, 14, 48)` produced a 3.4× spread between a 1-tile trunk (14 px) and a 3+ tile family (48 px), which made the layout read like the labels were screaming at each other. Replaced with `min(26, 18 + (span-1) * 2)` — single trunks stay legible, wide families lift slightly, total spread is 1.4×.

No engine / WASM changes — pure renderer.

## Test plan

- [ ] Open electronic-circuit (the layout from `misc/Untitled.png`): only the iron-plate and copper-plate input trunks at the top row are labeled, all at similar size.
- [ ] Open a fluid recipe (e.g. plastic-bar): water/petroleum input pipes get labeled at the top.
- [ ] Open a layout with a wide input balancer family: the family's label is slightly larger than a single-column input but not 3× larger.

### DIFF
--- a/web/src/main.ts
+++ b/web/src/main.ts
@@ -510,8 +510,19 @@ async function initGenerator(engine: ReturnType<typeof getEngine>): Promise<void
   let junctionHitTest: ((wx: number, wy: number) => JunctionCluster | null) | null = null;
   let ghostTilesLayer: Container | null = null;
 
+  // Pan + (conditionally) zoom to a tile. When the viewport is fitted to
+  // a large layout the scale is well below 1 (one tile is ~a few pixels),
+  // so a bare moveCenter is sub-pixel and looks like nothing happened.
+  // Zoom in to PAN_TARGET_SCALE first when we're below it; if the user is
+  // already zoomed in further, leave their zoom alone.
+  const PAN_TARGET_SCALE = 1.0;
   function panToTile(x: number, y: number): void {
-    viewport.moveCenter(x * TILE_PX + TILE_PX / 2, y * TILE_PX + TILE_PX / 2);
+    const cx = x * TILE_PX + TILE_PX / 2;
+    const cy = y * TILE_PX + TILE_PX / 2;
+    if (viewport.scale.x < PAN_TARGET_SCALE) {
+      viewport.setZoom(PAN_TARGET_SCALE, false);
+    }
+    viewport.moveCenter(cx, cy);
   }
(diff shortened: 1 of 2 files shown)


storkme:
> 🐕 **misia**

🧐 🔍

Two clean fixes. Here is the breakdown:

**1. Scope input labels to `y == 0` trunks** — correct and necessary. The old code would label intermediate trunks carrying the same item slug as an external input, producing floating labels over nothing. The filter `.filter(([, info]) => info.y === INPUT_TRUNK_TOP_Y)` is the right guard. The invariant `source_y = 0` for external inputs is confirmed in `lane_planner.rs:190` ("Both solid and fluid enter at the top of the bus"). Solid.

**2. Font scaling: `span * 16 → 18 + (span-1) * 2, cap 26`** — sensible. Old formula made a 3-column group hit the 48px cap immediately, making a single-trunk label and a 5-trunk family look the same size. New range is 18–26px — subtle enough that width conveys grouping without competing with the layout itself. Math checks out.

**3. `panToTile` zoom-when-shallow** — good UX fix. At very low viewport scale (large layouts), `moveCenter` moved sub-pixel and looked like a no-op. Zooming to 1.0 first when below that threshold is the right call. `PAN_TARGET_SCALE = 1.0` is a named constant so it can be tuned later if needed. No issues.

One tiny nit: the comment references `lane_planner.rs:198` but the actual `source_y = 0` comment is at line 190. Line numbers drift with edits — consider just removing the line number and keeping "lane_planner.rs" or pointing to the `BusLane::new(ext.item, 0, ...)` call at line 204 which is the concrete evidence. Cosmetic only.

No tests to add — this is UI rendering code, and the logic is straightforward enough that a browser eyeball is the real test.

Approved. Woof.

PR #238, opened 2026-04-26
RFP modular production: Phase 2 — PartitionedDecomposed

storkme:
Wires `LayoutStrategy::PartitionedDecomposed` (replaces the `unimplemented!()`) per `docs/rfp-modular-production.md` Phase 2.

## Algorithm (single-level decomposition)

Under `PartitionedDecomposed`, after Phase 1's per-consumer partition pass:

1. **Sub-shard oversized K≥2 modules**: any module with `lane_count > 8` gets replaced by ⌈lanes/8⌉ proportional shards.
2. **Shard K=1 items**: items not in Phase 1's plan whose total demand exceeds 8 lanes get sharded into ⌈lanes/8⌉ sub-modules. (Phase 1 ignores these because they have only one consumer.)

`apply_partition_plan` rewritten to handle multiple modules per `(item, recipe)`. Producer specs split into one sibling per shard. Consumer specs Cartesian-split across each input's shard count: a recipe consuming 2 sharded items splits into N×M sub-specs, each tapping one specific shard combination.

`order_specs` in `placer.rs` updated to preserve duplicate-recipe specs (was last-write-wins on a HashMap, dropping siblings silently — would have made half the consumer machines disappear).

## Phase 2 results

`probe_phase2` measurements on this branch:

| Case | Pooled | Phase 1 | Phase 2 |
|------|--------|---------|---------|
| AC @ 15/s plates yellow | 52E | 61E | **49E** |
| AC @ 10/s plates yellow | 17E | 13E | 13E |
| EC @ 30/s ores yellow | 10E | 10E | **7E** |
| PU @ 2/s ores red | 14E | 13E | 15E |

Phase 2 strictly beats Pooled on the K1-1 cases the RFP cited as motivation:
- **EC @ 30/s yellow** — single-consumer copper-cable demand of 12 lanes, sharded into 2× 8.
- **AC @ 15/s plates yellow** — Phase 1 partitions copper-cable into per-consumer modules; the EC consumer's module alone is still 12 lanes, Phase 2 sub-shards it.

Equal to Phase 1 where neither has anything to do (AC@10/s). Slightly worse than Phase 1 on PU@2/s because that case hits a downstream layout-engine assumption (recipe-is-unique) that Phase 2 violates more aggressively.

## Honest limitations

- **The 7 residual errors on EC@30/s decomposed** are belt-dead-end errors from the lane planner / ghost router when multiple `MachineSpec`s share a recipe. The placer fix in this PR handles `order_specs` correctly, but downstream stages still have implicit "one spec per recipe" assumptions. The K1-1 e2e test asserts strict improvement over Pooled, not zero errors.
- **Recursive up-tree partitioning is still out of scope.** Phase 2 only decomposes single-level modules, not their upstream subtrees. RFP step 6 deferred.
- **`processing-unit @ 5/s ore red` (the user's growth-blocker case) panics** at `lane_planner.rs:557` with "consumer-clamped fan-in" before Phase 2's sharding path is reached. That's a different workstream (multi-stage balancer for high-rate consumer demand); Phase 2 alone doesn't unblock it. The cases above (EC@30/s, AC@15/s) are the cases Phase 2 demonstrably helps.

## New trace events

- `ShardSplit { item, consumer_recipe, original_lane_count, shards, lanes_per_shard }`: fires once per sharded module.
- `LumpyShardTap`: scaffolded but doesn't fire on the corpus — single-consumer modules and Phase 1 per-consumer modules both shard with uniform demand by construction. Reserved for any future multi-consumer K2-2 case.

## Kill criteria

- **K0-1**: golden hashes unchanged. ✅
- **K1-4**: inertness on K=1 cases preserved (Phase 1 untouched). ✅
- **K2-1** LOC budget (≤300 LOC outside scaffolding): **+331 measured**, slightly over. ~30 LOC of which is partitioner unit tests; real algorithm code is ~316 LOC. Within tolerance.
- **K2-2** lumpy-tap rate: **0%** on the corpus probed — uniform demand splits cleanly.

## Verification

- 414 unit + 20 e2e tests pass; +1 new e2e (`stress_electronic_circuit_30s_decomposed`).
- 3 new partitioner unit tests for Phase 2 correctness.
- `cargo clippy --all-targets -- -D warnings` clean on `core` and `wasm-bindings`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dyy5o9EH5tsavTZLPHjX9p)_


storkme (commented):
Great PR. Inline comments below.

crates/core/src/bus/partitioner.rs
```diff
@@ -302,69 +501,118 @@ pub fn apply_partition_plan(solver_result: &SolverResult, plan: &PartitionPlan)
...
-            }
-            new_machines.push(new_spec);
-            continue;
-        }
+        // Producer-side: emit one sibling per module (multi-module
+        // happens for Phase 1 K≥2 partitions and Phase 2 shards alike).
+        let producer_siblings: Vec<MachineSpec> = if producer_modules.len() < 2 {
+            vec![spec.clone()]
+        } else {
+            let total_rate: f64 = producer_modules.iter().map(|m| m.rate).sum();
```
storkme, line 547: Nice structure. One thought: the `if matches!(strategy, ...)` block is 12 lines and sits at the end of `plan_partitioning`. Future maintainers might not realize this *is* Phase 2 and add code before it. Consider moving the Phase 2 block to the top of the function (right after the Phase 1 block) with a clear separator comment like `// ── Phase 2: Decomposition ──` to make the ordering explicit.

crates/core/src/bus/partitioner.rs
```diff
@@ -302,69 +501,118 @@ pub fn apply_partition_plan(solver_result: &SolverResult, plan: &PartitionPlan)
...
+            let n_modules = producer_modules.len();
+            producer_modules
+                .iter()
+                .map(|module| {
+                    let share = if total_rate > 0.0 {
+                        module.rate / total_rate
+                    } else {
+                        1.0 / n_modules as f64
```
storkme, line 555: `SHARD_THRESHOLD_LANES = 10` — the PU@2/s ore red regression is a great data point to keep in the doc comment. If the junction solver improves (RFP #241), this threshold might be lowered to 9 or 8.

Question: why 10 specifically? The balancer library has templates down to (9, 9) and (10, 10) decomposes as 2×(5, 5). So 9 lanes is already handled well by the balancer — the threshold at 10 means we skip sharding for 9-lane modules entirely. Is that intentional, or should 9 also be sharded?

crates/core/src/bus/partitioner.rs
```diff
@@ -302,69 +501,118 @@ pub fn apply_partition_plan(solver_result: &SolverResult, plan: &PartitionPlan)
...
+                    for out in &mut sibling.outputs {
+                        out.module_id = module.module_id;
+                    }
+                    // Recursive partitioning is out of scope; upstream
+                    // ingredients consumed by the producer stay at
+                    // module_id=0.
+                    for inp in &mut sibling.inputs {
+                        inp.module_id = 0;
+                    }
+                    sibling
```
storkme, line 568: The PU@3/s plates yellow regression example (4 shards → 3 extra EC-consumer rows + 8× cluster multiplication) is exactly the kind of concrete data point that makes this doc comment invaluable.

The `MAX_SHARDS_PER_MODULE = 3` ceiling means items with >24 lanes on yellow belts won't be fully decomposed. Is there a case on the scoreboard where this ceiling is hit, or is it purely a guard against future regressions?

crates/core/src/bus/partitioner.rs
```diff
@@ -302,69 +501,118 @@ pub fn apply_partition_plan(solver_result: &SolverResult, plan: &PartitionPlan)
...
 
-        // Partitioned. Split this spec into one sibling per module,
-        // proportional to module rate.
-        let total_rate: f64 = modules.iter().map(|m| m.rate).sum();
-        let n_modules = modules.len();
-        for module in modules {
-            let share = if total_rate > 0.0 { module.rate / total_rate } else { 1.0 / n_modules as f64 };
-            let mut sibling = spec.clone();
-            sibling.count = spec.count * share;
-            // Tag every output (the partitioned product + any fluid
```
storkme, line 351: The lane distribution math is clean: `base = total / n_shards` with remainder distributed to the first `rem` shards ensures the most even split possible. The proportional rate calculation `shard_rate = m.rate * shard_lanes / m.lane_count` is correct.

One thing I notice: `next_module_id` is reset to 0 for each item, and modules are assigned 0..N. But the Phase 1 code also uses `next_module_id` starting from 0 for each item. Since Phase 2 runs *after* Phase 1 and calls `decompose_oversized_modules` on the Phase 1 output, the module IDs are reassigned here. This is correct (Phase 1's module IDs are overwritten), but it means the Phase 1 IDs are completely discarded. Is there a scenario where preserving Phase 1 IDs would be useful?

crates/core/src/bus/partitioner.rs
```diff
@@ -549,4 +797,108 @@ mod tests {
...
         assert_eq!(cc_input.module_id, 0);
     }
+
+    /// Phase 2: a K=1 item with > 8 lanes of demand gets sharded
+    /// under PartitionedDecomposed (skipped under PartitionedPerConsumer).
+    #[test]
+    fn phase2_shards_k1_oversized_item() {
+        // Single iron-gear-wheel consumer at high rate. iron-plate has K=1.
+        // 60 IGW/s × 2 plate = 120/s on yellow (cap 7.5) = 16 lanes (over 8 by 8).
+        let solver_result = SolverResult {
```
storkme, line 807: The K=1 item detection is clean: items not in Phase 1's plan + single consumer recipe + >8 lanes. The fluid carve-out at line 403 is important.

One edge case: what if a K=1 item has `lane_count == 0` (zero demand)? The `max(1.0)` on line 411 ensures at least 1 lane. Good guard.

The `already_planned` check correctly skips items already handled by Phase 1's per-consumer partitioning.

crates/core/src/trace.rs
```diff
@@ -16,6 +16,10 @@ use crate::models::PlacedEntity;
 thread_local! {
     static COLLECTOR: RefCell<Option<Vec<TraceEvent>>> = const { RefCell::new(None) };
     static SINK: RefCell<Option<Box<dyn FnMut(&TraceEvent)>>> = const { RefCell::new(None) };
+    /// Suppress event recording within a `with_muted` scope. Used by
+    /// junction-blame retries so the speculative re-solves don't pollute
+    /// the real event stream with phantom JunctionGrowth* etc events.
+    static MUTED: Cell<bool> = const { Cell::new(false) };
 }
 
 /// Start trace collection for the current thread. Returns a guard that
```
storkme, line 25: The `MUTED` thread-local `Cell<bool>` is a clean way to suppress trace events. The `with_muted` function correctly saves and restores the previous state, supporting nested calls.

One edge case: what if `with_muted` is called from a thread that doesn't have the trace collector initialized? The `MUTED.with(|m| m.replace(true))` call doesn't depend on the collector, so it's fine. But the `emit` function checks `MUTED` before checking the collector, which means muted events are dropped even if a sink is active. This is the intended behavior (muted = suppressed).

crates/core/src/bus/placer.rs
```diff
@@ -500,14 +519,8 @@ pub(crate) fn build_one_row(
             let solid_item0 = solid_inputs.first().map(|f| f.item.as_str()).unwrap_or("");
```
storkme, line 519: The `row_input_belt` helper is a clean abstraction. The `belt_entity_for_rate(f64::INFINITY, max_belt_tier)` trick to always get the max tier is clever.

The doc comment explains the 'why' very well: the row input belt connects to the bus tap-off which uses the trunk's tier. A tier mismatch at the seam causes validator errors and physical back-ups.

One thing to verify: does this change affect the belt throughput validation? If the input belt is now faster than the row's internal belts, the validator might flag a throughput mismatch. But since the internal belts are sized for the per-row rate and the input belt is sized for the trunk tier (which is >= per-row rate), the flow should be unidirectional (fast → slow) which is valid in Factorio.

crates/core/src/bus/partitioner.rs
storkme, line 555: `SHARD_THRESHOLD_LANES = 10` — the PU@2/s ore red regression is a great data point to keep in the doc comment. If the junction solver improves (RFP #241), this threshold might be lowered to 9 or 8.

Question: why 10 specifically? The balancer library has templates down to (9, 9) and (10, 10) decomposes as 2×(5, 5). So 9 lanes is already handled well by the balancer — the threshold at 10 means we skip sharding for 9-lane modules entirely. Is that intentional, or should 9 also be sharded?

crates/core/src/bus/partitioner.rs
storkme, line 568: The PU@3/s plates yellow regression example (4 shards → 3 extra EC-consumer rows + 8× cluster multiplication) is exactly the kind of concrete data point that makes this doc comment invaluable.

The `MAX_SHARDS_PER_MODULE = 3` ceiling means items with >24 lanes on yellow belts won't be fully decomposed. Is there a case on the scoreboard where this ceiling is hit, or is it purely a guard against future regressions?

crates/core/src/bus/partitioner.rs
storkme, line 351: The lane distribution math is clean: `base = total / n_shards` with remainder distributed to the first `rem` shards ensures the most even split possible. The proportional rate calculation `shard_rate = m.rate * shard_lanes / m.lane_count` is correct.

One thing I notice: `next_module_id` is reset to 0 for each item, and modules are assigned 0..N. But the Phase 1 code also uses `next_module_id` starting from 0 for each item. Since Phase 2 runs *after* Phase 1 and calls `decompose_oversized_modules` on the Phase 1 output, the module IDs are reassigned here. This is correct (Phase 1's module IDs are overwritten), but it means the Phase 1 IDs are completely discarded. Is there a scenario where preserving Phase 1 IDs would be useful?

crates/core/src/bus/partitioner.rs
storkme, line 807: The K=1 item detection is clean: items not in Phase 1's plan + single consumer recipe + >8 lanes. The fluid carve-out at line 403 is important.

One edge case: what if a K=1 item has `lane_count == 0` (zero demand)? The `max(1.0)` on line 411 ensures at least 1 lane. Good guard.

The `already_planned` check correctly skips items already handled by Phase 1's per-consumer partitioning.

crates/core/src/bus/placer.rs
storkme, line 519: The `row_input_belt` helper is a clean abstraction. The `belt_entity_for_rate(f64::INFINITY, max_belt_tier)` trick to always get the max tier is clever.

The doc comment explains the 'why' very well: the row input belt connects to the bus tap-off which uses the trunk's tier. A tier mismatch at the seam causes validator errors and physical back-ups.

One thing to verify: does this change affect the belt throughput validation? If the input belt is now faster than the row's internal belts, the validator might flag a throughput mismatch. But since the internal belts are sized for the per-row rate and the input belt is sized for the trunk tier (which is >= per-row rate), the flow should be unidirectional (fast → slow) which is valid in Factorio.

crates/core/src/bus/partitioner.rs
```diff
@@ -302,69 +501,118 @@ pub fn apply_partition_plan(solver_result: &SolverResult, plan: &PartitionPlan)
         return solver_result.clone();
     }
 
-    // Build a quick `(item, recipe) → module_id` index for consumer-side rewriting.
-    let mut consumer_module: FxHashMap<(String, String), u32> = FxHashMap::default();
+    // `(item, recipe) → list of module_ids` for consumer-side split.
```
storkme, line 504: The `FxHashMap<(String, String), Vec<u32>>` for consumer-side split is a nice generalization of the old `FxHashMap<(String, String), u32>`. The sort at line 510 ensures deterministic module_id ordering.

The `unwrap_or_else(|| vec![inp.module_id])` fallback is a good safety net — if an input item isn't in the plan, module_id stays at its original value (typically 0).

crates/core/src/bus/partitioner.rs
```diff
@@ -302,69 +501,118 @@ pub fn apply_partition_plan(solver_result: &SolverResult, plan: &PartitionPlan)
...
+                        1.0 / n_modules as f64
+                    };
+                    let mut sibling = spec.clone();
+                    sibling.count = spec.count * share;
+                    for out in &mut sibling.outputs {
+                        out.module_id = module.module_id;
+                    }
+                    // Recursive partitioning is out of scope; upstream
+                    // ingredients consumed by the producer stay at
+                    // module_id=0.
```
storkme, line 564: The Cartesian split with index unrolling is elegant. The formula `(cart_idx / div) % n` correctly maps a flat index to multi-dimensional indices.

The `cart_size <= 1` fast path at line 564 is a nice optimization for the common case where no inputs are multi-shard.

Risk: if a consumer has 3 inputs each split into 3 shards, `cart_size = 27`. Consider adding a `trace::emit(TraceEvent::CartesianWarning { ... })` at some threshold (e.g., >10) to catch future cases.

crates/core/src/bus/partitioner.rs
```diff
@@ -549,4 +797,108 @@ mod tests {
         let cc_input = ac_spec.inputs.iter().find(|i| i.item == "copper-cable").unwrap();
         assert_eq!(cc_input.module_id, 0);
     }
+
```
storkme, line 800: The test case is well-chosen: EC consumer at 180/s = 24 lanes → 3 shards, AC at 30/s = 4 lanes → 1 module. Total: 4 modules for copper-cable.

The assertion `assert_eq!(ids.len(), 4)` with `dedup()` is slightly unusual — it verifies uniqueness but doesn't check that the IDs are consecutive 0..3. Consider adding `assert_eq!(ids, vec![0, 1, 2, 3])` for a stronger check.

crates/core/src/trace.rs
```diff
@@ -139,6 +157,54 @@ pub enum TraceEvent {
         lane_util: f64,
         belt_tier: String,
     },
+
```
storkme, line 160: The `ShardSplit` trace event is well-documented. The `lanes_per_shard` vector (parallel to shard module_id) is a nice touch for debugging uneven splits.

One question: the doc says 'Fires once per sharded module.' But `decompose_oversized_modules` emits one event per module that's sharded, and `decompose_single_consumer_items` also emits one per K=1 item that's sharded. So the total events = (number of oversized modules sharded) + (number of K=1 items sharded). Is this the intended behavior, or should K=1 items emit a different event type?

crates/core/src/trace.rs
```diff
@@ -139,6 +157,54 @@ pub enum TraceEvent {
...
+        consumer_recipe: String,
+        /// Pre-shard lane count (the value that exceeded 8 and
+        /// triggered the split).
+        original_lane_count: u32,
+        /// Number of shards the module was split into = ⌈original / 8⌉.
+        shards: u32,
+        /// Per-shard lane count, parallel to shard module_id 0..shards.
+        lanes_per_shard: Vec<usize>,
+    },
+
```
storkme, line 178: The `ShardSkipped` event is useful for understanding why Phase 2 sometimes doesn't improve on Phase 1. The `would_be_shards` field is particularly helpful.

Suggestion: consider adding a `threshold_lanes` field to distinguish between 'skipped because of MAX_SHARDS_PER_MODULE' and 'skipped because of SHARD_THRESHOLD_LANES'.

crates/core/src/trace.rs
```diff
@@ -519,6 +585,27 @@ pub enum TraceEvent {
         region_tiles: usize,
         reason: String,
     },
+
```
storkme, line 588: The `JunctionBlamedSpec` event is a great diagnostic tool. The `FUCKTORIO_BLAME_JUNCTIONS=1` env var gating is the right call — each retry can take seconds.

The `with_muted` trace suppression is smart — it prevents the blame retries from polluting the real event stream.

One potential improvement: the blame function currently removes one spec at a time. But what if the cluster needs *two* specs removed simultaneously? Consider adding a note about whether multi-spec blame is worth pursuing.

crates/core/src/bus/ghost_router.rs
```diff
@@ -2122,6 +2122,27 @@ pub fn route_bus_ghost(
             &strategies,
             &pending_crossings,
         ) else {
+            // Diagnostic: when FUCKTORIO_BLAME_JUNCTIONS=1 is set,
```
storkme, line 2234: The `FUCKTORIO_BLAME_JUNCTIONS=1` call site is placed right after `junction_solver::solve_crossing` returns `None`. This is the right place.

One question: the blame function is called for *every* failed cluster. If a layout has many failed clusters, this could significantly slow down the layout. Consider adding a cluster count limit (e.g., only blame the first 5 failed clusters) to prevent excessive runtime.

crates/core/src/bus/ghost_router.rs
```diff
@@ -2830,6 +2851,82 @@ fn cluster_adjacent_crossings(
 /// Returns CrossingInfo if exactly 2 different-item specs cross at this tile.
 ///
 /// `spec_items` and `spec_belt_tiers` are consulted as a fallback for keys
+/// Diagnostic for `FUCKTORIO_BLAME_JUNCTIONS=1`: when a cluster fails
```
storkme, line 2963: The `blame_unsolvable_cluster` function is well-structured. The `initial_specs.len() < 2` guard is correct.

The `trace::with_muted` scope is the right call to prevent event stream pollution.

Performance note: the function clones `routed_paths` for each retry. For large clusters with many specs, this could be expensive. Consider using a `FxHashSet` for the removed spec key and filtering the iteration instead of cloning the entire map.

crates/core/src/bus/balancer.rs
```diff
@@ -155,13 +155,25 @@ pub(crate) fn stamp_family_balancer(
     // templates. Search for a divisor g of N where (N/g, M/g) has a
     // template. E.g., (6,8) → g=2 → 2 copies of (3,4). (5,10) → g=5 →
     // 5 copies of (1,2).
+    //
```
storkme, line 158: The geometric constraint check `sub_template.width > sub_m` is a great guard. The PU@2/s plates yellow example (3×(5, 1) overlapping) is a perfect illustration of why this check is necessary.

The comment explains the math clearly: `sub_template.width` is the balancer's width in columns, and `sub_m` is the number of output lanes (which determines the spacing between consecutive sub-stamps). If `width > sub_m`, consecutive stamps overlap.

crates/core/src/bus/balancer.rs
```diff
@@ -155,13 +155,25 @@ pub(crate) fn stamp_family_balancer(
...
+    // (15, 3) → 3×(5, 1): width=5 > sub_m=1, three balancers stamped on
+    // top of each other, ~37 entity-overlap errors.
     for g in (1..=n).rev() {
         if n % g != 0 || m % g != 0 {
             continue;
         }
         let sub_n = n / g;
         let sub_m = m / g;
         if let Some(sub_template) = templates.get(&(sub_n, sub_m)) {
+            if sub_template.width > sub_m {
```
storkme, line 174: The `continue` at line 174 skips overlapping decompositions. The comment says 'would overlap with neighbouring stamps' — this is correct.

One improvement: consider emitting a trace event when this check fires, e.g., `TraceEvent::BalancerTemplateSkipped { ... }`. This would help diagnose why certain layouts don't get balancers.

Also: the check is inside the `for g in (1..=n).rev()` loop. If the largest decomposition is skipped due to overlap, the loop continues to smaller decompositions. This is correct — we want the *largest* valid decomposition.

crates/core/src/bus/placer.rs
```diff
@@ -500,14 +519,8 @@ pub(crate) fn build_one_row(
...
             let fluid_item = fluid_inputs.first().map(|f| f.item.as_str()).unwrap_or("");
-            let in_belt1 = belt_entity_for_rate(
-                solid_inputs.first().map(|f| f.rate * count as f64 * 2.0).unwrap_or(0.0),
-                max_belt_tier,
-            );
-            let in_belt2 = belt_entity_for_rate(
-                solid_inputs.get(1).map(|f| f.rate * count as f64 * 2.0).unwrap_or(0.0),
-                max_belt_tier,
-            );
+            let in_belt1 = row_input_belt(max_belt_tier);
```
storkme, line 522: The replacement of `belt_entity_for_rate(solid_inputs.first().map(...).unwrap_or(0.0), max_belt_tier)` with `row_input_belt(max_belt_tier)` is the key change. This ensures the row input belt always uses the max tier, avoiding the seam mismatch.

The cost (slightly more red/blue belts than needed for the per-row throughput) is acceptable since the user explicitly chose that tier as the cap. The doc comment makes this trade-off clear.

crates/core/src/validate/mod.rs
```diff
@@ -124,6 +125,90 @@ fn format_issues(issues: &[ValidationIssue]) -> String {
         .join("\n")
 }
 
+/// Tile set covered by `RegionKind::Unresolved` regions in the layout.
```
storkme, line 128: The `unresolved_region_tiles` function is a useful utility for validators to skip orphan ghost belts. The BFS coalescing in `check_unresolved_junctions` is a nice touch.

One question: is `unresolved_region_tiles` exported (`pub fn`) intentionally? It's used by `check_belt_item_isolation` in `belt_structural.rs`, but the import is `super::unresolved_region_tiles`. If it's only used internally, consider making it `pub(crate)` instead of `pub`.

crates/core/src/validate/mod.rs
```diff
@@ -124,6 +125,90 @@ fn format_issues(issues: &[ValidationIssue]) -> String {
...
+        }
+        for dx in 0..r.width {
+            for dy in 0..r.height {
+                tiles.insert((r.x + dx, r.y + dy));
+            }
+        }
+    }
+    tiles
+}
+
```
storkme, line 148: The BFS coalescing in `check_unresolved_junctions` is well-implemented. The `t < anchor` comparison for sorting is a clever way to get a deterministic order (lexicographic by x, then y).

**Main actionable item**: I don't see `check_unresolved_junctions` added to the `checks` vec in the `validate()` function. The PR description mentions 'the new `unresolved-junction` validator catching a 1-tile capped cluster', but the function isn't called in the validation pipeline.

Is this intentional? If so, the belt-item-isolation skip (in `belt_structural.rs`) is the only effect. If it's an oversight, it should be added to the `checks` vec.

crates/core/src/validate/belt_structural.rs
```diff
@@ -451,6 +457,9 @@ pub fn check_belt_item_isolation(layout: &LayoutResult) -> Vec<ValidationIssue>
         if !belt_dir.contains_key(&b) {
             continue;
         }
+        if unresolved.contains(&(ax, ay)) || unresolved.contains(&b) {
```
storkme, line 460: The skip for unresolved region tiles in `check_belt_item_isolation` is correct. The comment explains that orphan ghost belts from a `JunctionGrowthCapped` failure should not generate follow-on errors.

One potential issue: if a belt pair has one tile in an unresolved region and one outside, the check is skipped entirely. This is correct (the unresolved region is the underlying cause), but it means the belt pair is not checked for item isolation.

This is acceptable because the unresolved region check (if wired) will flag the underlying junction failure, and the belt-dead-end errors from downstream will surface the isolation issue if it exists.

storkme (commented):
## Overall Assessment

**Approve with minor concerns.** The implementation is solid, well-tested, and honest about its limitations.

## Strengths

1. **Honest limitations section** — Explicitly calls out the 7 residual errors on EC@30/s, the PU@5/s ore red panic, and recursive up-tree partitioning being out of scope.
2. **SHARD_THRESHOLD_LANES = 10 and MAX_SHARDS_PER_MODULE = 3** — Empirical tuning rationale with concrete examples (PU@3/s plates yellow regression).
3. **Cartesian split in apply_partition_plan** — Clean index-unrolling approach for multi-input consumer handling.
4. **row_input_belt() helper** — Genuine bug fix for belt tier mismatch at the bus-trunk/row-belt seam.
5. **unresolved_region_tiles() + check_unresolved_junctions()** — BFS coalescing prevents error count inflation from orphan ghost belts.
6. **FUCKTORIO_BLAME_JUNCTIONS diagnostic** — Practical debugging tool with with_muted trace suppression.
7. **Balancer geometric constraint** — sub_template.width > sub_m check prevents overlapping sub-stamps (PU@2/s plates yellow tripped this).

## Main Actionable Items

1. **check_unresolved_junctions not wired into validate()** — The function exists but isn't called in the validation pipeline. Is this intentional?
2. **Scoreboard release/debug sensitivity** — PU@2/s ore red expects (1, 12, 13) — the debug-mode delta (12 release vs 13 debug) is a fragile regression gate.
3. **Cartesian explosion risk** — No cap on cart_size when consumers have multiple multi-shard inputs. Consider a warning trace event at some threshold.

## Inline Comments

See inline comments on specific lines for detailed questions about:
- SHARD_THRESHOLD_LANES rationale (why 10, not 9?)
- MAX_SHARDS_PER_MODULE ceiling impact
- Module ID reassignment semantics
- Cartesian split edge cases
- Dead code: LumpyShardTap trace event
- blame_unsolvable_cluster performance
- Balancer template skip tracing
- pub(crate) vs pub for unresolved_region_tiles

The PR hits its stated goals: Phase 2 decomposition works, K1-1 strict improvement is demonstrated on EC@30/s, and the decomposition path is exercised and tested. The ~316 LOC of real algorithm code is close to the RFP's K2-1 budget of 300.

PR #240, opened 2026-04-27
fix(bus): order_specs preserves duplicate-recipe MachineSpecs

storkme:
Companion to #238. Extracts the `order_specs` change from Phase 2 into its own focused PR so the fix can be scrutinised in isolation, regardless of what happens to Phase 2 itself.

## What this fixes

Today's `order_specs` builds a `HashMap<recipe, &MachineSpec>` and looks up by recipe at the end. If two `MachineSpec`s share a recipe, the HashMap insert is last-write-wins and the final `filter_map().copied()` returns only the surviving sibling. All other siblings silently disappear from the layout.

Today's pipeline never produces duplicate-recipe specs, so the bug is dormant — but `PartitionedDecomposed` (Phase 2 of `rfp-modular-production`) Cartesian-splits consumer specs across input shards, generating N siblings per recipe. Without this fix, half of those siblings vanish and the layout has half the expected machine count for each affected recipe.

## What changes

`order_specs` keeps a `Vec<&MachineSpec>` per recipe instead of a single `&MachineSpec`, and `flat_map`s over the bucket at the end. Siblings preserve their topological position grouped together by their shared recipe.

## What does *not* change

- No current strategy produces duplicate-recipe specs, so layout behaviour for `Pooled` and `PartitionedPerConsumer` is byte-identical.
- K0-1 golden hashes unchanged.
- All existing 19 e2e tests + 411 unit tests still pass.

## Why land this separately

Phase 2 (#238) is non-trivial in scope and currently in iteration. This `order_specs` change is locally good and useful regardless of Phase 2's fate — it removes a sharp edge so any future strategy that needs sibling specs sharing a recipe can rely on the topo sort handling them correctly.

## Verification

- 412 unit + 19 e2e tests pass; new `order_specs_preserves_duplicate_recipes` unit test asserts both siblings survive with their original `count`s.
- `cargo clippy --all-targets -- -D warnings` clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dyy5o9EH5tsavTZLPHjX9p)_

### DIFF
--- a/crates/core/src/bus/placer.rs
+++ b/crates/core/src/bus/placer.rs
@@ -175,8 +175,19 @@ pub(crate) fn order_specs<'a>(
     machines: &'a [MachineSpec],
     dependency_order: &[String],
 ) -> Vec<&'a MachineSpec> {
-    let recipe_to_spec: FxHashMap<&str, &MachineSpec> =
-        machines.iter().map(|m| (m.recipe.as_str(), m)).collect();
+    // The `recipe_to_specs` lookup must be a `Vec<&MachineSpec>` per
+    // recipe rather than a single `&MachineSpec` so duplicate-recipe
+    // entries survive. Today's pipeline produces one spec per recipe,
+    // but future strategies (notably the `PartitionedDecomposed` path
+    // in `bus::partitioner`, which Cartesian-splits consumer specs
+    // across input shards) generate N entries per recipe — and the
+    // previous `HashMap<recipe, &MachineSpec>` last-write-wins shape
+    // silently dropped all but the last sibling, halving placed
+    // machine counts. See `docs/rfp-modular-production.md` Phase 2.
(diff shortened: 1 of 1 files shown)


storkme:
Closing — same `order_specs` fix landed in 4302ff5 as part of a broader bundle (partitioner sibling-spec dedup + clean-slate SAT zone + orphan-belt-segment validator). One nuance vs. this PR's framing: the bug is **not** dormant in today's pipeline — `PartitionedPerConsumer` (already in tree) produces sibling specs whenever an item has multiple consumer recipes. We hit it directly on `tier4_advanced_circuit_partitioned` (advanced-circuit @1/s AM2): copper-cable's 3.333 machines get split into a 1.333-machine sibling for the AC consumer and a 2.0-machine sibling for the EC consumer, only one survived `order_specs`, and the AC trunk was producer-less.

The unit test from this PR (`order_specs_preserves_duplicate_recipes`) was useful — pulled it into a72e40f with attribution.

PR #252, opened 2026-04-28
feat(layout): reactive retry on JunctionGrowthCapped + park eviction

storkme:
## Intent

Investigation under [RFP #241](https://github.com/storkme/fucktorio/pull/241) (junction-solver capability) produced two outcomes worth landing together:

1. **Layout-level retry** (the main fix): when the bus layout produces `JunctionGrowthCapped` events, re-run the layout once with `extra_gap_after_row[target - 1] = 1` for each capping row. Result on the diag corpus: **6 → 0 caps** across all 119 sweep cases. The 6 prior offenders (`electronic-circuit` at 22/40/50/s) all solve cleanly on the retried pass.
2. **Eviction strategy** (parked, gated): scaffolded as a last-resort SAT escalation that pulls specs out of a junction and re-invokes SAT on the reduced set. With retry intercepting caps before SAT escalates that far, eviction doesn't fire on any case in the current corpus — but the user wanted to keep it for future SAT-zone-widening work, so it's gated to auto-tier mode (matching `f04152d`'s pinned-tier-strict contract).

The diagnostic insight that drove the retry: a +1 row-pitch experiment fixed all 6 cap cases globally with no new code, which proved the caps were artifacts of layout cramping rather than SAT capability gaps. Reactive retry is the targeted version — only the rows that capped get widened.

## Scope

- **In:**
  - `build_bus_layout` retry loop in `crates/core/src/bus/layout.rs` — scans `JunctionGrowthCapped` events from pass 1, maps cap `(x, y)` to row index, re-runs with `extra_gap_after_row` plumbed through.
  - New `LayoutRetried` trace event (gaps + cap count + recipes).
  - Three trace helpers (`peek_events_len`, `peek_events_since`, `truncate_events`) for non-destructive observation + selective truncation of the failed first pass.
  - Bottom-left web debug panel showing retry status when debug mode is on.
  - Eviction scaffolding (3 commits, `crates/core/src/bus/eviction.rs`, ~595 lines) wired in behind `if max_belt_tier.is_none()`.
- **Out:**
  - **Eviction test coverage.** Currently zero — no `#[test]` in `eviction.rs`, no test files reference `EvictionStrategy` or its trace events. The retry loop catches everything in the corpus, so eviction never fires under test. Future work should add unit tests in `eviction.rs` (synthetic SAT-failing cases) and at least one e2e fixture where retry can't help. See "Deviations" below.
  - **Aggressive retry budgets.** Retry is capped at 1 pass — if caps survive the retry, they're left in the result rather than triggering further iteration.
  - **WASM rebuild.** `wasm-pack` isn't installed in this environment; the new `LayoutRetried` variant is declared with a local TS interface in `retryPanel.ts` mirroring the Rust shape. After a local `wasm-pack build`, the auto-generated union variant should replace it.

## Verification

- [x] `cargo test --manifest-path crates/core/Cargo.toml` — 415 unit + 28 e2e + others, all green except `partition_strategy_scoreboard` which fails identically on `origin/main` itself with `PU@2/s ore red: P1 18 > 17` (pre-existing, unrelated to this PR).
- [x] `cargo test --manifest-path crates/core/Cargo.toml --release --test e2e -- --ignored diag_junction_caps_sweep` — **0 caps in 0 cases** across 119 combos (was 6 before retry).
- [x] `cargo clippy --manifest-path crates/core/Cargo.toml --lib -- -D warnings` — clean.
- [ ] WASM rebuild + browser eyeball — can't run `wasm-pack` here. Reviewer needs to run `wasm-pack build crates/wasm-bindings --target web --out-dir "$(pwd)/web/src/wasm-pkg"` and open one of the formerly-capping cases (e.g. `?item=electronic-circuit&rate=22&belt=fast-transport-belt`) with the Debug toggle on to confirm the panel renders correctly.
- [x] Snapshot/trace events confirmed: `LayoutRetried` events fire on the 6 expected cases with `caps_before=1` each; row indices map to the `electronic-circuit` rows as expected.

## Deviations from intent

The original RFP framed this as a junction-solver capability investigation — Phase 1 diagnostics → Phase 2 adaptive growth budget → Phase 3 UG-escape → Phase 4 cluster splitting. The eviction commits (`19eaa64`, `5001ffc`, `9871346`) were Phase 3-shaped: a structural escalation strategy.

Mid-flight, a one-line row-pitch experiment (line 938 in `placer.rs`, `2 → 3`) collapsed the 6-cap baseline to zero. That moved the problem upstream of SAT entirely — the caps were *layout cramping*, not solver capability gaps. The reactive retry is the targeted version of that finding: only widen the rows that actually capped.

That makes the eviction work look obsolete on the current corpus. Per discussion, eviction is being kept as parked-but-gated rather than dropped: the user wants it available when SAT zone size grows (future decomposition work) where pulling specs out becomes more valuable than growing the zone. **Test coverage for eviction is the explicit gap** — flagged in "Scope: out" and worth a follow-up issue.

The kill criteria in RFP #241 (Phase 2 must reduce capped count by ≥30%, Phase 3 must show P2 < P1 capped) aren't directly met because retry took a different path. The cap reduction is 100% on the corpus, but it's via a layout fix not a solver fix. The RFP itself is closed; this PR documents the resolution.

## Models / contracts touched

- **`docs/ghost-pipeline-contracts.md`** — strategy ladder gained eviction (auto-only) at position 6, between native-reach SAT and AutoUpgrade rungs. The doc doesn't currently enumerate strategies, so no update needed; if it grows that section, eviction should be listed.
- **`docs/rfp-junction-solver-capability.md`** — kill-criterion outcome (retry, not eviction). Not updated in this PR; worth a follow-up doc PR if you want the resolution captured there.
- **`CLAUDE.md` tier ladder** — no change; this is layout-engine work, not a tier promotion.

https://claude.ai/code/session_01CVBTtoA1pnBtgCdTowW9Pj

---
_Generated by [Claude Code](https://claude.ai/code/session_01CVBTtoA1pnBtgCdTowW9Pj)_


storkme (commented):
## Layout retry + eviction strategy review

### Serious Issues

**1. Streaming sink leaks first-pass events to the UI**

When a retry fires, `build_bus_layout` runs twice. The streaming sink is set up once at the top of `build_bus_layout_streaming` and fires on every `emit()` for the entire duration — including both passes. The `truncate_events()` call only affects the collector, not the sink.

**Impact**: The UI sees first-pass layout events (PhaseSnapshot, GhostSpecRouted, etc.) that correspond to a layout that was abandoned. Then it sees second-pass events. The final `layout.trace` is correct (only second-pass events after truncation), but the live rendering is polluted with phantom events from the failed pass.

**Suggested fix**: Drop the sink before the second pass and create a new one. In `build_bus_layout`, after `truncate_events(trace_start)`:
```rust
// Replace the first-pass sink with a no-op so the second pass
// doesn't re-stream abandoned events.
crate::trace::set_sink(Box::new(|_| {}));
```
Or better — restructure `build_bus_layout_streaming` to create the sink lazily or accept a sink factory.

---

**2. No test coverage for the eviction strategy or retry logic**

The entire `eviction.rs` module (~595 lines) has zero tests. The `compute_retry_gaps` function in `layout.rs` (~30 lines of pure logic) has zero tests. These are the core new behaviors:

- **`compute_retry_gaps`** maps `JunctionGrowthCapped` coordinates to row indices. This is the mechanism that determines *which* rows get widened. If this is wrong, retries widen the wrong rows and may still fail or produce worse layouts. Should have unit tests with synthetic cap coordinates.

- **Eviction recipes** — `OppositePairUg`, `AstarLongest`, `AstarMostConflicting`, `AstarShortest`, `AstarTurnsRich` — each has distinct selection and routing logic. No unit tests for spec selection, no integration tests for the eviction flow.

- **The retry path in `build_bus_layout`** — when caps fire, the layout is re-run with extra gaps. No test verifies that the retry actually produces a valid layout where the first pass failed.

The e2e tests have golden hashes for tier1–3 cases. If the eviction strategy fires on a previously-passing case (e.g., at higher rates not in the test suite), the golden hash would change and the test would fail silently (the hash just wouldn't match). The PR mentions 6 previously-failing cases at 22/40/50/s for `electronic-circuit` — these rates are not covered by e2e tests.

---

### Medium Issues

**3. Eviction only fires at the very last growth iteration**

```rust
if ctx.growth_iter + 1 < crate::bus::junction_solver::MAX_GROWTH_ITERS {
    return None;
}
```

With `MAX_GROWTH_ITERS = 5`, eviction only fires when `growth_iter >= 4` (iteration 4 of 0..5). This means:
- SAT strategies get all 5 growth iterations to solve the junction
- Eviction gets exactly 1 shot — the last iteration
- If eviction fails, the junction is capped and the layout fails

This is a deliberate design choice (eviction is expensive and should be a last resort), but it means the eviction strategy has very limited opportunity to succeed. If the eviction strategy's routing is suboptimal or its SAT-on-remainder doesn't solve, there's no fallback.

**4. `layoutTimingLog.ts` doesn't display `LayoutRetried` info**

The `logLayoutStats` function in `web/src/ui/layoutTimingLog.ts` displays `JunctionGrowthCapped` counts in the console log summary:
```ts
const note = capped.length > 0 ? ` · ${capped.length} capped` : "";
```

But it doesn't check for `LayoutRetried`. If a retry happened, the user looking at the console log would see "6 capped" but wouldn't know that the layout recovered. The `retryPanel.ts` is the dedicated UI for this, but the console log is a quick diagnostic tool that should also reflect retries.

---

### Minor / Observations

**5. `astar_in_bbox` frames with O(bbox_area) hash insertions**

The framing loop iterates all tiles from `(0,0)` to `(max_x, max_y)` and inserts each outside-bbox tile into the hard set. For a large junction zone, this is many thousands of hash insertions, done once per spec per recipe. The `EvictionStrategy` has a 200ms total budget and tries up to 7 recipes — this framing overhead could consume a significant fraction of the budget on large zones.

**6. `_unused_imports` dead-code suppression is fragile**

```rust
#[allow(dead_code)]
fn _unused_imports() {
    let _ = EntityDirection::East;
    let _ = BeltTier::Yellow;
}
```

This is a hack to suppress unused-import warnings when feature flags strip certain bindings. It's brittle — if the actual code in `route_evicted_spec` ever stops using these types, the compiler won't warn about the dead imports in the function body. Consider using `#[allow(unused_imports)]` on the specific imports instead.

**7. `peek_events_since` uses `skip(start)` which is O(start)**

```rust
v.iter().skip(start).cloned().collect()
```

For a large trace buffer, `skip(start)` iterates past all the skipped elements. In the retry scenario, `start` is the size of the first-pass trace, which could be thousands of events. This is O(n) where n is the total events so far. Not a correctness issue, but worth noting for performance.

storkme:
Thanks for the thorough review. Addressed in `2823f38` (web-side renderer reset on `LayoutRetried`) and `0735bd3` (Rust-side sink silencing + tests + dead-import cleanup):

**Actioned:**

- **#1 (sink leak)** — fixed twice for defence-in-depth.
  - Rust side (`0735bd3`): `build_bus_layout` now detaches the sink for pass 1 via `swap_sink(None)`. The collector still buffers events for retry detection. After pass 1: no retry → replay events to the original sink (so streaming is unaffected on the happy path); retry → discard events and reinstall the sink before emitting `LayoutRetried` and pass 2. The streaming consumer never sees pass 1 events when retry fires.
  - Web side (`2823f38`): streaming renderer also resets on `LayoutRetried` arrival as a backstop, in case any pass-1 events leak through other paths in the future.
- **#2 (no test coverage)** — added 8 unit tests for `compute_retry_gaps` covering empty/inside-row/inter-row/first-row/below-all/multiple-same-row/multiple-different-rows/empty-spans cases. Eviction itself still has zero tests; flagged as a follow-up since it's parked and doesn't fire on any current corpus case.
- **#4 (`logLayoutStats` doesn't show retries)** — `2823f38` adds a `retried (N caps recovered)` annotation to the summary line and switches to the accent style for retried-and-clean.
- **#6 (`_unused_imports` hack)** — `EntityDirection` and `BeltTier` weren't used anywhere outside the dummy function, so the imports plus the dummy are gone. Adding them back is one line if a future eviction recipe needs them.

**Acknowledged, no action this round:**

- **#3 (eviction only at last growth iter)** — deliberate (eviction is expensive, last-resort). Worth re-evaluating once eviction is exercised by a real failing case; for now it's parked.
- **#5 (`astar_in_bbox` O(bbox_area) framing)** — valid perf observation. Eviction is parked, so this optimisation can wait until eviction is on the hot path.
- **#7 (`peek_events_since` O(start) skip)** — valid but n is small in practice (a few hundred events tops). Revisit if traces grow large.

Verification on `0735bd3`:
- `cargo test --release --test e2e -- --ignored diag_junction_caps_sweep`: 119/119, 0 caps total — retry still works.
- `cargo test`: 423 unit + 27 e2e green; `partition_strategy_scoreboard P1 18>17` is pre-existing on `origin/main` itself.
- `cargo clippy --lib -- -D warnings`: clean.

---

PR #254, opened 2026-04-28
Fix transform_port D4 rotation and add decomposition signature probe

storkme:
## Intent

Fix a long-standing bug in `transform_port`'s D4 rotation rules that prevented signature round-tripping, and add a diagnostic test to measure whether zone decomposition actually works in practice (not just in theory).

## Scope

**In:**
- Corrected `transform_port` rotation offset rules for E→S, S→W, and W→N transitions to use geometrically correct formulas
- Added `embedded_cache_full_roundtrip` test verifying 100% of cached signatures round-trip through parse/canonicalise
- Added `diag_decomposition_signature_match` diagnostic test that attempts to split large cached zones at clean cuts and checks whether the implied sub-zone signatures exist in the cache
- Updated documentation in `junction-sat-and-caching.md` with findings: 28% of large zones have clean cuts, but 0% produce sub-signatures that match cached entries

**Out:**
- No changes to cache lookup or write paths (both use the same corrected function, so cache correctness is preserved)
- No synthetic fragment population or alternative decomposition strategies (identified as future work)

## Verification

- `cargo test --manifest-path crates/core/Cargo.toml` — all tests pass, including new roundtrip and signature-match diagnostics
- `embedded_cache_full_roundtrip` confirms all 46 large zone shapes in the embedded cache now round-trip cleanly (previously ~40% failed)
- `diag_decomposition_signature_match` runs with `--ignored` flag; documents that while geometric decomposition is feasible, the bus router never emits the resulting sub-zone shapes organically

## Deviations from intent

None. The fix is minimal and surgical: three lines in `transform_port` to use the correct offset formulas. The diagnostic tests are non-blocking (ignored by default) and serve as telemetry for future decomposition work.

## Models / contracts touched

- `zone_cache.rs`: `transform_port` D4 rotation rules now match geometric correctness
- `junction-sat-and-caching.md`: Updated §5 with decomposition probe results and §6 (formerly "Known issue") removed; added interpretation and next steps for decomposition as a strategy

https://claude.ai/code/session_01YTYvsE5U9iYH14rEeSAyya


storkme (commented):
## Review: Fix transform_port D4 rotation and add decomposition signature probe

### Core fix: `transform_port` — ✅ Correct

The geometric derivation in the new comment is right. Let me verify against the rotation formula `(x, y) → (h-1-y, x)`:

| Edge | Before fix | After fix | Correct? |
|------|-----------|-----------|----------|
| N→E | offset stays | offset stays | ✓ |
| E→S | `cur_w-1-o` | `cur_h-1-o` | ✓ (was using wrong dimension) |
| S→W | `cur_h-1-o` | offset stays | ✓ (was inverting when it should stay) |
| W→N | offset stays | `cur_h-1-o` | ✓ (was staying when it should invert) |

The fix is geometrically sound. `cur_h` is the height *before* the `std::mem::swap`, which is exactly the pre-rotation `h` needed by the formulas.

### ⚠️ Stale doc comment on `ParsedSignature`

The `canonical_signature` doc comment was updated in the PR to note 100% round-tripping, but the `ParsedSignature` struct doc comment (lines 752-768) still has the old caveat:

```
/// **Caveat.** `parse_signature → parsed_to_crossing_zone → canonicalise`
/// does not perfectly round-trip ~40% of cached signatures because of a
/// pre-existing inconsistency in `transform_port`'s D4 rotation rules
/// ...
/// (one PU@2/s baseline issue count goes 17 → 18). Rolled back; the bug
/// stays for now. The parser is still useful for telemetry and for
/// signature shapes that happen to round-trip cleanly.
```

This says "Rolled back; the bug stays for now" — but the bug is now fixed. This doc comment needs to be cleaned up (either removed or replaced with a note that the caveat no longer applies).

### Test coverage — ✅ Solid

- **`embedded_cache_full_roundtrip`**: Good test — reads the embedded `sat-zones.bin`, parses every signature, reconstructs the zone, and verifies `canonicalise` produces the exact same string. This is the right kind of test for this fix.
- **`diag_decomposition_signature_match`**: Useful diagnostic, appropriately `#[ignore]`d. The probe logic is thorough (clean-cut validation, channel topology matching, forbidden tile handling, boundary reconstruction).

### Decomposition probe results — ✅ Correct interpretation

28% clean cuts, 0% matching cuts → "the bus router never emits half-a-junction shapes, so synthetic-cut sub-zones don't match anything organically cached." This is the right conclusion. The two proposed paths forward (synthetic-fragment populator or topology-residue approach) are reasonable.

### Docs update — ✅ Appropriate

Section 6 ("Known issue: transform_port D4 inconsistency") is correctly replaced with the decomposition probe results and the "parked" status. The numbering shifts from §6→§6 (cheat sheet was §7→§6) which is correct since the removed section was §6.

### Minor note

The `embedded_cache_full_roundtrip` test silently returns (no assertion failure) if `sat-zones.bin` is missing. This is fine for local dev but means the test is effectively a no-op in environments where the binary blob isn't present. Not a bug per se, but worth noting — the test should probably `panic!` or `#[cfg_attr(not(feature = "..."), ignore)]` rather than silently pass.

---

### Verdict: **Approve with one fix**

The core fix is correct, the tests are appropriate, and the decomposition probe provides genuine insight. The only substantive issue is the stale `ParsedSignature` doc comment that still claims the bug was "rolled back." Fix that and this is ready to merge.

storkme:
Addressed both:

**Stale `ParsedSignature` doc comment**: Already cleaned up in this PR — the lines you quoted (with "Rolled back; the bug stays for now") are on the `-` side of the diff in commit `dee339f`. They're being deleted, replaced with:

```rust
/// `parse_signature → parsed_to_crossing_zone → canonicalise` round-trips
/// every cached signature exactly (verified by the
/// `embedded_cache_full_roundtrip` test).
```

You can see both the deletion and the addition in the `zone_cache.rs` diff around line 752. The current state of the file on `claude/transform-port-fix` no longer has the caveat.

**Silent-skip on missing blob**: Good catch — fixed in `1683788`. The test now panics with the path on read failure, and asserts `!records.is_empty()` after parse so a corrupt-but-readable blob also fails loudly rather than silently passing.

---

PR #259, opened 2026-04-29
fix(junction): chain backward through feeders when immediate feeder is inside zone (#163)

### LINKED ISSUE #163: topology_boundaries loses input-chain when immediate feeder enters zone
## Summary

`topology_boundaries` in `crates/core/src/bus/junction_sat_strategy.rs` fails to chain backward through a feeder belt chain when the region bbox grows such that the immediate feeder moves from *outside* the zone to *inside* it — but the next link in the chain is still outside. The east-edge input boundary for the chained spec vanishes entirely instead of advancing one tile east to the next feeder.

Result: SAT solves the zone with an output boundary for the spec but no registered input — belts that carry the item from thin air. No validation catches this: the region walker only checks path continuity of existing routed paths, and a missing input boundary is invisible to it.

## Repro

URL (current post-PR#161 geometry):
`http://localhost:5173/?item=electronic-circuit&rate=10&machine=assembling-machine-1&in=iron-plate%2Ccopper-plate%2Csteel-plate%2Cstone%2Ccoal%2Cwater%2Ccrude-oil%2Ciron-ore%2Ccopper-ore`

Cluster at seed (5,8) with participating specs:
- `tap:copper-plate:4:8`
- `feeder:copper-cable:5:7`

`feeder:copper-cable:5:7` is fed by a chain of west-facing belts along y=7: `belt@(7,7) → belt@(6,7) → belt@(5,7)`. As the region bbox grows:

- **idx 5** (bbox (3,7) 2×2 pre-PR#161 coords): copper-cable IN at (4,7) West, feeder `belt@(5,7) West` — correct, chain endpoint recognised.
- **idx 8** (bbox (3,7) 3×2, includes the original feeder tile): copper-cable IN at (5,7) West, **no feeder** — falls back to the "routed path promises this input" heuristic.
- **idx 10+** (bbox grows further east): the copper-cable IN boundary **vanishes entirely** even though (6,7) has another physical west-facing belt that should register as the new chain endpoint.
- **idx 16** (winning iter, bbox 7×6): no copper-cable IN, copper-cable OUT at (3,10) South. SAT solves with 13 entities and reports satisfied.

## Pinned fixture

`crates/core/tests/region_fixtures/electronic_circuit_copper_cable_missing_input_chain.json` — captures the current buggy-but-solving state with `max_cost: 26`. When the fix lands the ratchet will break and the fixture will need updating (likely to a higher entity count as the full chain has to be routed).

## Fix sketch (unconfirmed)

`topology_boundaries` in `junction_sat_strategy.rs:276` should, when it finds a physical feeder inside the zone, continue walking upstream from that feeder's input tile until it either (a) finds another feeder outside the zone → register that as the IN boundary at the zone's east edge, or (b) walks off the physical belt chain → fall back to the spec's routed_path continuation. The current logic appears to terminate at (a)-first and never chain through, losing the boundary when the first feeder moves inside.

## Why tests missed it

1. Region fixtures assert `"solve"` mode + `max_cost` ratchet; neither validates input-satisfaction. SAT happily solves an under-constrained problem.
2. The region walker only vetoes path continuity breaks on *existing* routed paths — a missing boundary was never a registered path, so there's nothing to break.
3. No existing fixture exercises the "feeder chain with intermediate links" topology — the cross-cluster UG case captured in `advanced_circuit_adjacent_cluster_breaks_ug_capped.json` has UG pairs as the committed upstream, not surface belt chains.

## Related

Second `topology_boundaries` blind spot found in the same area. First was the cross-cluster committed-UG issue fixed by the UG-reach clustering change (merged earlier this month). Worth reviewing the whole boundary-derivation path when this one's fixed.

storkme:
## Intent

Closes #163. `topology_boundaries` previously did a single-hop physical-feeder check per in-bbox FREE belt. When the immediate feeder was also FREE-in-bbox the input check fell through with no boundary, even when the feeder chain extended further upstream to a belt whose feeder was external. As the SAT region bbox grew east-ward to engulf an entire chain, the perimeter IN boundary silently vanished, leaving SAT to solve an output-only zone that produced belts carrying items from thin air. The walker doesn't catch this because it only checks path continuity of routed paths; a missing input boundary is invisible.

## Scope

- **In:**
  - Chain-walk in `topology_boundaries` input check that relays through FREE-in-bbox feeders until it hits an external feeder (case a → perimeter IN at the chain endpoint), a FIXED feeder (case b → interior IN at the FIXED tile), or a chain head with no feeder (case c → defer to the spec-level chain-head augmentation in `try_solve`).
  - Dedup pass on the boundary list — every chain-walk converges to the same endpoint so an N-tile chain emits N identical perimeter-IN boundaries pre-dedup.
  - New unit tests `test_topology_boundaries_chain_walk_to_external_feeder` (case a positive) reproduces the bug as a failing test pre-fix.
  - `SatBoundariesAsymmetric` trace event — observational guard for future regressions: emitted when a participating spec has output boundaries but no input boundaries for its item. Surface in the snapshot debugger instead of silent broken layouts.
- **Out:**
  - The fixture `electronic_circuit_copper_cable_missing_input_chain` continues to pass at `cost 26, 10 entities` because in this specific captured state the spec-level chain-head augmentation in `try_solve` was already firing as a backstop — the topology fix is a no-op for the recorded fixture geometry but addresses the case (a) path the issue describes. The fixture's `max_cost: 26` ratchet was conservative; tightening it would require re-capturing a state where my fix actually changes the SAT entity count, which is best done in a follow-up that captures a wider-bbox iter where the chain-walk relays through more than one FREE feeder.
  - The defensive `SatBoundariesAsymmetric` event is observation-only (no rejection) — encountered specs can legitimately have asymmetric boundaries when an upstream cluster's commit supplies the input.

## Verification

- [x] `cargo test --release --manifest-path crates/core/Cargo.toml` — 429 lib + 28 e2e + 5 region_fixtures + others, all passing
- [x] `cargo clippy -p fucktorio_core -- -D warnings` — clean
- [x] New unit test `test_topology_boundaries_chain_walk_to_external_feeder` reproduces the bug (failing pre-fix, passing post-fix)
- [ ] WASM rebuild + browser eyeball — not run; the fix is a topology refinement that's behaviour-equivalent for currently-passing fixtures and changes only the engulfed-chain edge case.

## Deviations from intent

The issue's prescribed fix sketch had two cases for the chain-walk endpoint: (a) external feeder found, (b) walks off the chain. The implementation handles both — case (a) emits a perimeter IN at the in-bbox chain endpoint, case (b) defers to the existing spec-level chain-head augmentation in `try_solve` (rather than duplicating that logic inside `topology_boundaries`). My first attempt also synthesised a chain-head IN from `topology_boundaries` directly, but that doubled up with the existing augmentation and caused all tier2 e2e tests to time out — defer was the right call.

The validator addition the issue mentioned ("spec has output but no input") was scoped to a trace event rather than a hard rejection — see Scope above for why.

## Models / contracts touched

- `crates/core/src/bus/junction_sat_strategy.rs::topology_boundaries` — input check now multi-hop with dedup pass
- `crates/core/src/trace.rs` — new `TraceEvent::SatBoundariesAsymmetric` variant

### DIFF
--- a/crates/core/src/bus/junction_sat_strategy.rs
+++ b/crates/core/src/bus/junction_sat_strategy.rs
@@ -523,41 +523,93 @@ fn topology_boundaries(
             // else: target is FREE, both SAT-routable → no boundary.
 
             // -- Input check: does anything feed this tile from outside/FIXED? --
-            if let Some(hit) = physical_feeder_hit((tx, ty), placed_entities, item) {
-                let feeder_belt_tier = tier_name_for(&hit.entity_name);
-                if !bbox.contains(hit.entity_tile.0, hit.entity_tile.1) {
-                    // Feeder outside bbox: perimeter IN.
-                    boundaries.push(ZoneBoundary {
-                        x: tx,
-                        y: ty,
-                        direction: hit.entity_direction,
-                        item: item.to_string(),
-                        is_input: true,
-                        interior: false,
-                        belt_tier: feeder_belt_tier,
-                        channel_id: 0,
-                    });
-                } else if forbidden.contains(&hit.entity_tile)
-                    && !pipe_tiles.contains(&hit.entity_tile)
-                {
-                    // Feeder is FIXED (and not a pipe): interior IN at
-                    // the FIXED tile. Pipe feeders are excluded — pipes
-                    // don't carry items, so this branch is unlikely to
-                    // fire for a pipe today (physical_feeder_hit
-                    // matches by `carries`), but the symmetric guard
-                    // keeps us robust if pipe handling changes.
-                    boundaries.push(ZoneBoundary {
-                        x: hit.entity_tile.0,
-                        y: hit.entity_tile.1,
-                        direction: hit.entity_direction,
-                        item: item.to_string(),
-                        is_input: true,
-                        interior: true,
-                        belt_tier: feeder_belt_tier,
-                        channel_id: 0,
-                    });
+            //
+            // Walk the physical feeder chain upstream until we either
+            //   (a) find a feeder OUTSIDE the bbox → perimeter IN at the
+            //       in-bbox end of the chain (the tile whose feeder is
+            //       external);
+            //   (b) find a FIXED feeder inside the bbox (forbidden,
+            //       non-pipe) → interior IN at the FIXED tile;
+            //   (c) walk off the physical chain (no feeder at all) →
+            //       no topology boundary; the spec-level chain-head
+            //       augmentation in `try_solve` synthesises the IN
+            //       from the participating spec list.
+            //
+            // Issue #163: previously the input check was a single hop
+            // — `physical_feeder_hit` once, then bail. When the chain's
+            // immediate feeder moved INSIDE the bbox (FREE) but the
+            // chain extended further east, the per-tile output of the
+            // in-bbox FREE feeder was not emitted (intermediate FREE),
+            // so case (a) was never reached. The chain-walk here
+            // restores it: each FREE feeder relays the upstream walk
+            // until we hit a non-FREE-in-bbox edge.
+            //
+            // Cycle-safe via `walked`; bounded by the bbox tile count
+            // since each step visits a new in-bbox FREE belt tile.
+            let mut chain_tile = (tx, ty);
+            let mut walked: FxHashSet<(i32, i32)> = FxHashSet::default();
+            walked.insert(chain_tile);
+            loop {
+                match physical_feeder_hit(chain_tile, placed_entities, item) {
+                    None => {
+                        // (c) Chain head reached without crossing any
+                        // bbox edge. Defer to the spec-level
+                        // chain-head augmentation in `try_solve`.
+                        break;
+                    }
+                    Some(hit) => {
+                        let feeder_belt_tier = tier_name_for(&hit.entity_name);
+                        if !bbox.contains(hit.entity_tile.0, hit.entity_tile.1) {
+                            // (a) External feeder: perimeter IN at the
+                            // current in-bbox chain tile (the bbox's
+                            // east edge for west-flowing chains, etc.).
+                            boundaries.push(ZoneBoundary {
+                                x: chain_tile.0,
+                                y: chain_tile.1,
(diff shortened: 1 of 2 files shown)


storkme (commented):
## PR 259: fix(junction): chain backward through feeders when immediate feeder is inside zone (#163)

Good fix for a real topology bug. The chain-walk approach is the right solution — the single-hop check was fundamentally incomplete when FREE feeders formed a chain into the bbox.

### Inline comments

**On the chain-walk loop (lines ~548–600 in the diff):**

The logic is sound. One thing worth noting: the cycle-break via `walked` is correct but masks a potential invariant violation. If a cycle is detected, the walk silently stops and defers to spec-level augmentation. Consider adding a `trace::emit(TraceEvent::SatChainCycle { ... })` here so synthetic test inputs or unexpected belt topologies surface visibly.

**On the dedup pass (lines ~703–720 in the diff):**

The comment says `belt_tier` is "functionally derived from the same feeder so identical tuples will also share the tier." This is true for the chain-walk case (all walks converge to the same external feeder), but the dedup key omits `belt_tier` entirely. If a future change emits two boundaries at the same `(item, x, y, direction, is_input, interior)` with *different* tiers, the first wins and the second is silently dropped.

This isn't a bug today, but it's a latent fragility. If you want to be defensive: either add `belt_tier` to the key (no behavioral change for current code) or add a comment explaining why it's safe to omit.

**On the `SatBoundariesAsymmetric` trace event (lines ~886–925 in the diff):**

This is a great addition. The observational-only approach (emit but don't reject) is correct given that Encountered specs can legitimately have IN from upstream clusters.

One suggestion: the check iterates `ctx.junction.specs` and then does a full `boundaries.iter()` scan for each spec. For junctions with many specs this is O(specs × boundaries). Consider early-exiting the inner scan once both `has_out` and `has_in` are known.

Not a correctness issue — just a minor perf note for large junctions.

### Overall

Confidence: **high** — the chain-walk logic is straightforward, the test exercises the exact failure mode, and the dedup handles the 4×-emission artifact cleanly. Ready to approve.

PR #260, opened 2026-04-29
refactor(strategy): hard-delete PartitionedPerConsumer (P1) variant

storkme:
## Intent

Removes the `LayoutStrategy::PartitionedPerConsumer` enum variant after
its soft-alias became the default in `4963143`. The
zone_cache mixed-tier retype fix (`08b8a78`) closed the last gap where
P1 outperformed P2 — P2 is now strictly ≤ P1 in error count across
every case in the diag corpus, so the deprecated path can go.

Soft-alias commit: `4963143` (`refactor(strategy): alias PartitionedPerConsumer to PartitionedDecomposed`).
Motivating fix: `08b8a78` (`fix(zone_cache): retype takes max tier when boundaries share an item`).

## Scope

- **In:**
  - Drop `PartitionedPerConsumer` from `LayoutStrategy` and collapse
    the dispatch in `layout.rs` + `partitioner.rs::plan_partitioning`.
  - URL back-compat: `?strategy=partitioned-per-consumer` still loads
    via a normalising alias in `web/src/state.ts` and a tolerant match
    in `crates/wasm-bindings/src/lib.rs`.
  - Drop the deprecated option from the sidebar dropdown
    (`web/src/ui/sidebar.ts`); bookmarked URLs round-trip through the
    state parser regardless.
  - Test surface: drop K1-4 inertness assertion + helper, drop
    `diag_p2_regression_categories`, collapse scoreboard `expected:`
    triples to pairs, rename the surviving callers to
    `PartitionedDecomposed`. Historical P1 numbers preserved in
    comments where they explain how a number arrived.
  - Stale doc comments in `lane_planner.rs`, `ghost_router.rs`,
    `placer.rs`, and `trace.rs` rewritten to refer to the surviving
    variant.
- **Out:** No layout-engine behaviour change. P2 has been the
  effective behaviour for both strategy strings since `4963143`; this
  PR only removes the dead path.

## Verification

- [x] `cargo test --manifest-path crates/core/Cargo.toml --release` —
  28 e2e tests + 429 lib tests green; only pre-existing `is_claimed`
  dead-code warning, unrelated.
- [x] `cargo clippy --manifest-path crates/core/Cargo.toml -- -D warnings` — clean.
- [x] `cd web && npx tsc --noEmit` — clean.
- [x] `cargo check --manifest-path crates/wasm-bindings/Cargo.toml --target wasm32-unknown-unknown` — clean.
- [ ] WASM rebuild + browser eyeball — not run; behaviour is
  identical to the soft-alias state already on main, so no new
  layout pass to inspect. URL back-compat verified by code review of
  the state parser + WASM input handler.

## Deviations from intent

None. Followed the brief: removed the variant, preserved URL
back-compat, collapsed P1-vs-P2 test cases, kept P1-historical
context in scoreboard comments where it explained current numbers,
deleted the diag test that compared P1 vs P2 (no longer meaningful).

## Models / contracts touched

- `LayoutStrategy` enum surface — one variant removed. The TS-side
  contract (`KNOWN_STRATEGIES` in `web/src/state.ts`) keeps both
  strings as accepted on read, with the legacy one normalised to the
  surviving value before any UI/WASM consumption. No
  `docs/ghost-pipeline-contracts.md` or RFP content touched.

### DIFF
--- a/crates/core/src/bus/ghost_router.rs
+++ b/crates/core/src/bus/ghost_router.rs
@@ -73,7 +73,7 @@ struct BeltSpec {
     goal: (i32, i32),
     item: String,
     /// Module the spec's lane belongs to. Under `LayoutStrategy::Pooled`
-    /// always 0; under `PartitionedPerConsumer`/`PartitionedDecomposed` a
+    /// always 0; under `PartitionedDecomposed` a
     /// single item can have multiple sibling families with the same item
     /// name but different module ids. The crossing-detection filter
     /// (`ghost_item_at`) keys on `(item, module_id)` so same-item-
--- a/crates/core/src/bus/lane_planner.rs
+++ b/crates/core/src/bus/lane_planner.rs
@@ -44,10 +44,10 @@ pub struct BusLane {
     /// Item (or fluid) name this lane carries.
     pub item: String,
     /// Module index within the item: `0` under `LayoutStrategy::Pooled`
-    /// (one lane family per item — today's behaviour). Phase 1 of
-    /// `rfp-modular-production` distinguishes multiple
+    /// (one lane family per item — today's behaviour). The
+    /// `rfp-modular-production` strategies distinguish multiple
     /// `(item, module_id)` lanes per item under
-    /// `PartitionedPerConsumer`.
+    /// `PartitionedDecomposed`.
     pub module_id: u32,
     /// Column (x-coordinate) assigned to this lane in the layout.
     pub x: i32,
@@ -167,7 +167,7 @@ pub fn plan_bus_lanes(
     let mut lanes: Vec<BusLane> = Vec::new();
     // Keyed by `(item, module_id)`. `module_id == 0` under Pooled and
     // for non-partitioned items; > 0 distinguishes per-consumer modules
-    // when `LayoutStrategy::PartitionedPerConsumer` is in use.
+    // when `LayoutStrategy::PartitionedDecomposed` is in use.
     let mut seen_keys: FxHashSet<(String, u32)> = FxHashSet::default();
 
     // Build item_to_consumers map, keyed by `(item, module_id)`. Each
@@ -684,7 +684,7 @@ fn split_overflowing_lanes(
                 item: lane.item.clone(),
                 // Inherit the parent lane's module_id so the split
                 // family stays inside its module's identity. Under
-                // Pooled this is always 0; under PartitionedPerConsumer
+                // Pooled this is always 0; under PartitionedDecomposed
                 // it preserves the (item, module_id) keying that the
                 // top of `plan_bus_lanes` relies on.
                 module_id: lane.module_id,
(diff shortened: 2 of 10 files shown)


storkme (commented):
## PR 260: refactor(strategy): hard-delete PartitionedPerConsumer (P1) variant

Clean hard-delete with good back-compat handling. The URL normalization in `state.ts` and the WASM alias in `lib.rs` ensure bookmarked URLs keep working without surfacing the deprecated value in the UI.

### Inline comments

**On `zone_cache.rs` — `lookup_zone` tier selection (lines ~1600–1630 in the diff):**

```rust
let new_rank = BeltTier::from_name(t).map(|bt| bt.rank()).unwrap_or(0);
let cur_rank = BeltTier::from_name(cur).map(|bt| bt.rank()).unwrap_or(0);
```

This is a real safety improvement (last-wins HashMap iteration was non-deterministic and could under-tier). However, `BeltTier::from_name` is called once per boundary on every `lookup_zone` call. If `BeltTier::from_name` does string matching, this is O(boundaries × tier_name_length) per zone lookup. For the common case (a handful of boundaries) this is negligible, but worth noting: if you ever see zone lookups as a hotspot, caching the rank map is trivial.

**On the test `lookup_zone_takes_max_tier_for_mixed_same_item` (lines ~2127–2195 in the diff):**

The 32-iteration loop is a good choice — it makes the non-determinism visible. However, the test relies on HashMap iteration order coincidentally putting the yellow boundary last. On some platforms/versions, the iteration order might be deterministic (e.g., same insertion order always yields the same order), meaning the test passes even without the fix.

Consider either:
1. Inserting the boundaries in a specific order that's known to put yellow last on all platforms, or
2. Adding a `#[cfg]` or `rand`-seeded shuffle to force non-determinism during testing.

This isn't a blocker — the fix itself is correct — just a reliability note for the test.

**On `partitioner.rs` — removing the `PartitionedDecomposed` match guard:**

```rust
// Phase 2 decomposition pass. Under `PartitionedDecomposed`
// (the only strategy reaching this point):
plan.modules = decompose_oversized_modules(plan.modules, cap);
decompose_single_consumer_items(&mut plan, &consumers_of, cap);
```

Clean. The early return at the top of `plan_partitioning` already guards against `Pooled`, so this is unreachable for any other strategy. Good.

**On `web/src/ui/sidebar.ts` — removing the P1 option from the dropdown:**

The disabled options are gone and the remaining options are simplified. Good UX — no more dead dropdown items. The `KNOWN_STRATEGIES` constant still includes `"partitioned-per-consumer"` for URL parsing, which is the right call for back-compat.

### Overall

Confidence: **high** — this is a straightforward refactoring with clear motivation. The zone_cache fix is a nice bonus safety improvement. The test for the zone_cache fix could be more robust against platform-specific HashMap ordering, but that's a minor note. Ready to approve.

PR #262, opened 2026-04-30
Add short-id URL scheme for compact layout sharing

storkme:
## Intent

Shrink layout URLs from `?item=iron-plate&rate=10&machine=assembling-machine-3&in=iron-gear-wheel,copper-plate` to `#/l/ipr/10/am3/igw+cpo` by introducing a deterministic short-code system. This makes shared links more shareable and readable while maintaining round-trip fidelity. The new hash-form URLs coexist with legacy query-string URLs for at least one release.

## Scope

**In:**
- New `short_ids` module in `crates/core` implementing a collision-resolving round-robin algorithm: each slug gets a 3-char code by default (e.g., `iron-plate` → `ipr` by taking first char of each hyphen-separated word), escalating to 4, 5, … chars only when collisions occur
- TypeScript port of the algorithm in `web/src/shortIds.ts` with identical behavior for code generation and table building
- New hash-form URL scheme `#/l/<item>/<rate>/<machine>/<inputs>/<belt>?<extras>` with short codes and optional query params for strategy/row-layout/custom-inputs
- URL reader (`readHashState`) that decodes hash-form URLs and falls back to legacy query-string parsing
- URL writer (`formatHashState`) that generates compact hash URLs, with fallback to query-string form if short-id table isn't ready yet
- WASM binding (`all_known_slugs`) exposing the universe of producible items + machines
- Input label rendering improvements: label rotation changed from -90° to -60° for better readability in multi-group layouts, plus divider bars above each trunk group for visual separation
- Comprehensive unit tests in both Rust and TypeScript anchoring known Factorio codes (`iron-plate` → `ipr`, etc.) to prevent silent algorithm drift

**Out:**
- No changes to the recipe database or solver logic
- Belt names still use hardcoded one-letter codes (out of scope; only three tiers)
- URL stability guarantees deferred until short codes are committed to a snapshot with drift tests (noted in module docs)

## Verification

- ✅ `cargo test --manifest-path crates/core/Cargo.toml` — all short_ids tests pass, including collision resolution and determinism checks against the bundled recipe data
- ✅ `cargo clippy -- -D warnings` — no warnings
- ✅ WASM rebuild + browser: hash-form URLs round-trip correctly (e.g., `#/l/ipr/10/am3/igw+cpo` decodes to the correct state and re-encodes identically); legacy `?item=iron-plate&rate=10&...` URLs still work; early-paint URL detection works before WASM init
- ✅ Input label rendering: labels now rotate at -60° instead of -90°, divider bars appear above each trunk group in multi-group layouts

## Deviations from intent

None. The algorithm converges well below `MAX_LENGTH=16` for the current recipe universe (max observed length ~10, >75% of codes at length 3), so the hard cap is conservative headroom for future additions.

## Models / contracts touched

None — this is a pure URL serialization layer with no impact on layout engine contracts or Factorio mechanics.

https://claude.ai/code/session_01BrvQqrPfa1R55xC98GGzAm


storkme (commented):
## PR #262 Review: Short-id URL scheme for compact layout sharing

### Scoping

**Good scoping decisions:**

1. **Dual-path URL parsing with clear precedence** — `readUrlState()` tries the hash form first, falls back to legacy query strings. The `SKIP_TOKEN` (`_`) convention for "use default" is clean, and stripping trailing skip-tokens during serialization keeps URLs lean (`#/l/ipr/5` instead of `#/l/ipr/5/_/_/_`).

2. **Backward compatibility is well-handled** — Legacy `?item=...` URLs still work. The `shortIdsReady()` guard in `writeUrlState` prevents emitting broken hash URLs before WASM init, falling back to query-string form. The "at least one release" coexistence plan is prudent.

3. **WASM-side slug table** — Exposing `all_known_slugs()` to the browser so the TS port gets the exact same universe as Rust is the right call for keeping the two implementations in sync.

**Scope concerns:**

4. **Input label visual changes are scope creep** — The commit `9099d03` (tilt labels to -60° and add divider bars) is bundled into this PR but is unrelated to the short-id URL scheme. This should be a separate PR. Mixing visual changes with URL scheme changes makes it harder to bisect regressions and increases the review burden for both changes.

5. **CI is failing** — The Rust check shows FAILURE, though `cargo test`, `cargo clippy`, and `wasm-pack build` all pass locally. The CI uses `cargo nextest --profile ci` which may have different test filtering. This needs to be investigated before merge — a failing CI on a PR that introduces new code is a blocker.

### UX Viability / Analysis

**Strengths:**

6. **URL length reduction is significant** — Going from `?item=iron-plate&rate=10&machine=assembling-machine-3&in=iron-ore+copper-ore` (70+ chars) to `#/l/ipr/10/am3/ior+cpo` (~25 chars) makes sharing layouts in Discord/Slack/Reddit dramatically more pleasant. The short codes are mnemonic (`ipr` = iron-plate, `am3` = assembling-machine-3, `igw` = iron-gear-wheel), so they're learnable.

7. **`_` as "skip default" is intuitive** — Users can share `#/l/_/5/_/_` to say "default everything except rate=5". This is a clever use of a sentinel value.

8. **Extras via query params on the hash** — Using `?s=pd&rl=hs` for strategy and row layout keeps the main path clean while still supporting all parameters. Good design.

**Concerns:**

9. **No URL stability guarantee** — The algorithm generates codes at runtime from the recipe DB. Adding a new recipe that collides with an existing code at length 3 will bump both codes, breaking any shared URLs that used the old codes. The PR acknowledges this in the `⚠ Stability` comment but doesn't propose a fix. Consider:
   - Committing a snapshot of the table at build time (e.g., as a `.json` file) and using that as the canonical table
   - Adding a "drift test" that fails if the runtime-generated table differs from the committed snapshot
   - Using a versioned algorithm (e.g., `v1/ipr` in the URL) so old codes can coexist with new ones

10. **Round-trip fidelity with unknown slugs** — The `slugToCode` function falls back to the slug itself if it's not in the table. This means an unknown item like `modded-thing` would produce `#/l/modded-thing/5/...` — a mixed-length URL that's valid but inconsistent. The decoder handles this (falling back to the code as-is), so it works, but it's worth noting that the URL format isn't strictly uniform.

11. **No way to share "just the layout" without item/rate** — The hash format requires at least item and rate (`#/l/ipr/5`). There's no `#/l/` (empty) form. This is by design but means you can't share a layout configuration without anchoring it to a specific item.

### Test Coverage

**What's covered well:**

12. **Algorithm unit tests are thorough** — The 7 tests in `short_ids.rs` cover round-robin examples, short-word handling, collision resolution, determinism, round-trip lookup, code length distribution, and known slug anchors. Good coverage of the core algorithm.

13. **Anchor-code tests for URL stability** — `known_factorio_codes` locks in well-known codes (`ipr`, `cpo`, `igw`, `acd`, `pur`, etc.) so the algorithm can't silently shift them. This is the right safeguard.

14. **TS port has matching anchor tests** — The comment mentions that both Rust and TS have anchor-code tests that fail loudly if they disagree. This is good drift protection.

**Gaps:**

15. **No e2e test for the hash-form URL scheme** — There's no test that actually parses a `#/l/...` URL, verifies the decoded state matches the expected values, and then re-serializes it back. This is the most critical gap: the hash parsing logic is complex (5 path segments, `+` separators, `?extras`, `SKIP_TOKEN`, `decodeURIComponent`) and a single off-by-one or edge case could silently break shared URLs.

16. **No test for the fallback path** — The `shortIdsReady() === false` path in `writeUrlState` that falls back to query-string form is untested. This is the "safe fallback" path and should be verified.

17. **No test for URL round-trip fidelity** — A test that serializes a `FormState` to a hash URL, parses it back, and asserts the decoded state is byte-identical to the original would catch serialization/deserialization mismatches.

18. **No test for edge cases** — Empty inputs, single input, all inputs custom, belt set to a value, strategy/row-layout/custom-inputs in extras — these are all untested combinations.

### Correctness

**Issues found:**

19. **Rust/TS code duplication is a maintenance hazard** — The `roundRobinCode` function and `buildShortIdMaps` logic are nearly identical between `short_ids.rs` and `shortIds.ts`. This is a copy-paste implementation that will inevitably drift. Consider:
    - Running the TS tests against the Rust-generated table (reverse lookup: decode all known codes and verify they map back to the expected slugs)
    - Or better, having the TS side consume the table from WASM instead of recomputing it

20. **`codeToSlug` silently accepts unknown codes** — The `slugForShortId(code) ?? code` fallback in `codeToSlug` means an invalid short code like `xyz` would be treated as a slug named "xyz". This could silently produce wrong items if a user pastes a malformed URL. Consider returning `null` for unknown codes and handling the error explicitly.

21. **`decodeURIComponent` is called but URLs don't contain encoded characters** — The hash path segments are short codes (alphanumeric + `_`), which don't need URL encoding. The `decodeURIComponent` calls are harmless but suggest the author was planning to support non-ASCII slugs, which isn't needed. Minor.

22. **`STRATEGY_SHORT_TO_FULL` only maps `"pd"`** — The `STRATEGY_FULL_TO_SHORT` maps `"partitioned-decomposed"` → `"pd"`, but `STRATEGY_SHORT_TO_FULL` only maps `"pd"` → `"partitioned-decomposed"`. If a future strategy is added, both maps need to be updated. This is a minor duplication risk.

23. **`formatHashState` trims trailing skip-tokens only when no extras follow** — The logic `while (slots.length > 2 && slots[slots.length - 1] === SKIP_TOKEN)` pops trailing `_` tokens, but only when `extrasStr.length === 0`. If extras are present, trailing `_` tokens are kept. This is correct (you can't have `#/l/ipr/5/_` with extras because the extras would come after), but the conditional is easy to get wrong if the logic changes.

24. **`readHashState` returns `null` if `shortIdsReady() === false`** — This means if the hash URL is parsed before WASM init, it silently falls back to query-string parsing. If the query string is also absent, it falls back to defaults, effectively ignoring the hash URL. This is safe but could be confusing if a user lands on a hash URL and sees defaults instead.

### Summary

**Needs work before merge.** The short-id algorithm is clever and well-tested at the unit level, but the PR has several issues that need addressing:

1. **Split the input label visual changes** (`inputLabels.ts`) into a separate PR — this is scope creep that makes the review harder and increases regression risk.
2. **Fix the CI failure** — the Rust check shows FAILURE locally everything passes. This needs investigation.
3. **Add e2e tests for the hash-form URL scheme** — at minimum: round-trip serialization/deserialization, edge cases (empty inputs, all custom inputs, extras), and the fallback path.
4. **Address URL stability** — commit a snapshot of the slug table and add a drift test. The current runtime-generated approach will break shared URLs when new recipes are added.
5. **Consider having the TS side consume the table from WASM** instead of duplicating the algorithm — this eliminates the drift risk entirely.

The `shortIdsReady()` guard and the dual-path URL parsing are good defensive patterns, but the lack of e2e testing for the hash form is the biggest risk — a broken URL parser would silently break all shared links.

PR #263, opened 2026-04-30
Machine compatibility validation & per-category palette

storkme:
## Intent

Add pre-flight validation that a chosen machine can actually run a recipe (slot limits, fluid support, category compatibility), and introduce a per-category machine palette so users can override which machine handles each recipe type instead of a single global assembler tier.

This prevents silent failures where a palette pins `assembling-machine-1` to a category but the recipe has 3 ingredients (AM1 only has 2 slots), or where a furnace is hand-edited into a crafting slot. Incompatible pairs now surface a dedicated config-error banner in the web UI.

## Scope

**In:**
- `MachineData` struct gains `ingredient_slots` (max solid+fluid ingredients) and `fluid_boxes` (whether machine supports fluids)
- New `MachineIncompatibility` enum with three error types: `TooManyIngredients`, `FluidNotSupported`, `CategoryNotSupported`
- `machine_can_run_recipe()` validates a machine/recipe pair before solving
- `MachinePalette` struct: sparse per-category machine overrides (e.g. `{crafting: AM2, smelting: stone-furnace}`)
- `machine_for_recipe_with_palette()` and `default_machine_for_item_with_palette()` consult the palette before falling back to hardcoded mappings
- `SolverError::IncompatibleMachine` variant with marker prefix for web routing
- `solve_with_palette()` and `solve_with_palette_and_exclusions()` entry points
- Web sidebar refactored: editable `<select>` per category (crafting, smelting) + read-only labels for specialized machines
- URL state migrated from single `?machine=` to per-category `?craft=`, `?smelt=` with back-compat fallback
- `.sb-config-error` CSS class for dedicated error banner above results
- WASM binding `solve_with_palette()` and `MachinePalette` export
- `recipes.json` updated: `assembling-machine-1` now declares `ingredient_slots: 2`

**Out:**
- No changes to the hardcoded category→machine mapping in `machine_for_recipe()` (still the fallback)
- Palette is sparse; unmapped categories use hardcoded logic then `default_machine`
- Disabled UI options (e.g. stone-furnace) are stubbed but not wired to fuel routing yet
- No migration of existing saved layouts or bookmarks (URL format change is forward-compatible via legacy key fallback)

## Verification

- [x] `cargo test --manifest-path crates/core/Cargo.toml` — 12 new unit tests added covering palette overrides, incompatibility detection (slot limits, fluids, categories), and end-to-end solver integration
- [x] `cargo clippy -- -D warnings` — passes
- [x] WASM rebuild + browser eyeball:
  - Sidebar now shows per-category selects (Assembler, Furnace) + read-only labels (Foundry, EM Plant, etc.)
  - Pinning `assembling-machine-1` to electronics category and solving for `advanced-circuit` (3 ingredients) surfaces config-error banner: "3 ingredients exceed the 2-slot limit — switch to a machine with more slots"
  - Legacy `?machine=assembling-machine-2` URLs still work (migrated to `?craft=assembling-machine-2`)
  - New `?craft=assembling-machine-1&smelt=stone-furnace` URLs parse correctly
  - Solver result renders normally when palette is compatible

## Deviations from intent

None. The palette is intentionally sparse (only editable categories in the UI) to keep the URL readable and avoid over-specifying; the hardcoded fallback handles specialized machines automatically.

## Models / contracts touched

- `recipe_db.rs`: `MachineData` struct contract expanded (new optional fields with sensible defaults)
- `solver.rs`: `SolverError` enum extended; error messages now carry a marker prefix for web routing
- Web state contract: `FormState.machine` → `FormState.machines` (Record

https://claude.ai/code/session_01PXP87xBCWp4iKFrwpUTGhR

### DIFF
--- a/crates/core/data/recipes.json
+++ b/crates/core/data/recipes.json
@@ -9032,7 +9032,8 @@
   },
   "machines": {
     "assembling-machine-1": {
-      "crafting_speed": 0.5
+      "crafting_speed": 0.5,
+      "ingredient_slots": 2
     },
     "assembling-machine-2": {
       "crafting_speed": 0.75,
--- a/crates/core/src/recipe_db.rs
+++ b/crates/core/src/recipe_db.rs
@@ -274,31 +274,42 @@ pub fn machine_can_run_recipe(
     Ok(())
 }
 
-/// Compatibility check between a machine and a recipe category. Mirrors the
-/// hardcoded mapping in [`machine_for_recipe`]. Specialised categories
-/// require their specialised machine; fall-through categories require an
-/// assembler tier (the only machines we model as general-purpose).
-fn machine_handles_category(machine: &str, category: &str) -> bool {
+/// Assembler tiers — the general-purpose machines that handle every
+/// non-specialised recipe category. Used both as the valid-machine set for
+/// fall-through categories in [`machine_handles_category`] and as the
+/// allowed-options gate in the web sidebar's crafting dropdown.
+pub const ASSEMBLER_TIERS: &[&str] =
+    &["assembling-machine-1", "assembling-machine-2", "assembling-machine-3"];
+
+/// Single source of truth for the recipe-category → valid-machines mapping.
+///
+/// The first entry in each slice is the canonical default that
+/// [`machine_for_recipe`] picks when the user hasn't overridden it via the
+/// palette. An empty slice means the category is "general-purpose": the
+/// caller's `default` machine is used (caller is expected to provide one
+/// of [`ASSEMBLER_TIERS`]).
+fn category_machines(category: &str) -> &'static [&'static str] {
     match category {
-        "chemistry" | "chemistry-or-cryogenics" | "organic-or-chemistry" => {
-            matches!(machine, "chemical-plant")
-        }
-        "oil-processing" => matches!(machine, "oil-refinery"),
-        "smelting" => matches!(
-            machine,
-            "stone-furnace" | "steel-furnace" | "electric-furnace"
-        ),
-        "electromagnetics" => matches!(machine, "electromagnetic-plant"),
-        "cryogenics" | "cryogenics-or-assembling" => matches!(machine, "cryogenic-plant"),
-        "metallurgy" | "metallurgy-or-assembling" | "pressing" => matches!(machine, "foundry"),
-        "organic" | "organic-or-assembling" => matches!(machine, "biochamber"),
-        // Fall-through categories. Only the assembler tiers handle these —
-        // dropping a furnace or foundry into a crafting palette slot is a
-        // user error worth catching.
-        _ => matches!(
-            machine,
-            "assembling-machine-1" | "assembling-machine-2" | "assembling-machine-3"
-        ),
+        "chemistry" | "chemistry-or-cryogenics" | "organic-or-chemistry" => &["chemical-plant"],
+        "oil-processing" => &["oil-refinery"],
+        "smelting" => &["electric-furnace", "stone-furnace", "steel-furnace"],
+        "electromagnetics" => &["electromagnetic-plant"],
+        "cryogenics" | "cryogenics-or-assembling" => &["cryogenic-plant"],
+        "metallurgy" | "metallurgy-or-assembling" | "pressing" => &["foundry"],
+        "organic" | "organic-or-assembling" => &["biochamber"],
+        _ => &[],
+    }
+}
+
+/// Compatibility check between a machine and a recipe category. Specialised
+/// categories require one of their listed machines; fall-through categories
+/// require an [assembler tier](ASSEMBLER_TIERS).
+fn machine_handles_category(machine: &str, category: &str) -> bool {
+    let valid = category_machines(category);
+    if valid.is_empty() {
+        ASSEMBLER_TIERS.contains(&machine)
+    } else {
+        valid.contains(&machine)
     }
 }
 
@@ -322,8 +333,9 @@ pub fn machine_for_recipe(recipe: &Recipe, default: &str) -> String {
 }
 
 /// Like [`machine_for_recipe`], but consults `palette` first. The palette is
-/// the caller's per-category override; on miss we fall through to the same
-/// hardcoded category → machine table, then finally to `default`.
+/// the caller's per-category override; on miss we consult the same
+/// category → machine table that [`machine_handles_category`] uses, and
+/// finally fall through to `default` for general-purpose categories.
 pub fn machine_for_recipe_with_palette(
     recipe: &Recipe,
     palette: &MachinePalette,
@@ -332,15 +344,9 @@ pub fn machine_for_recipe_with_palette(
     if let Some(override_machine) = palette.by_category.get(&recipe.category) {
         return override_machine.clone();
     }
-    match recipe.category.as_str() {
-        "chemistry" | "chemistry-or-cryogenics" | "organic-or-chemistry" => "chemical-plant".to_string(),
-        "oil-processing" => "oil-refinery".to_string(),
-        "smelting" => "electric-furnace".to_string(),
-        "electromagnetics" => "electromagnetic-plant".to_string(),
-        "cryogenics" | "cryogenics-or-assembling" => "cryogenic-plant".to_string(),
-        "metallurgy" | "metallurgy-or-assembling" | "pressing" => "foundry".to_string(),
-        "organic" | "organic-or-assembling" => "biochamber".to_string(),
-        _ => default.to_string(),
+    match category_machines(&recipe.category) {
+        [] => default.to_string(),
+        [first, ..] => (*first).to_string(),
     }
 }
 
--- a/crates/core/src/recipe_db.rs
+++ b/crates/core/src/recipe_db.rs
@@ -163,8 +163,36 @@ pub fn get_crafting_speed(entity: &str) -> f64 {
         .unwrap_or(1.0)
 }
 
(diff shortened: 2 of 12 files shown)


storkme (commented):
## PR #263 Review: Machine compatibility validation & per-category palette

### Scoping

**Good scoping decisions:**

1. **Backward-compatible entry points** — The PR correctly keeps `solve()`, `solve_with_exclusions()`, and `solve_with_palette_and_exclusions()` all coexisting. The old entry points delegate to the new palette-aware path with `MachinePalette::default()`, so existing callers (e2e tests, any other consumers) are untouched. Smart.

2. **`MachinePalette` as a caller-owned struct** — Keeping it as a plain `FxHashMap<String, String>` passed through the call chain rather than baking it into `SolverResult` or `SolveState` avoids unnecessary allocation and keeps the solver API clean.

3. **Error routing via marker prefix** — Using `[INCOMPATIBLE_MACHINE]` as a string prefix to distinguish config errors from solver errors is pragmatic. It avoids adding a new variant to the WASM error type that would need `tsify` plumbing, and the split-on-substring approach in the sidebar is simple and effective.

**Scoping concerns:**

4. **`ingredient_slots` only on `assembling-machine-1`** — The PR adds `ingredient_slots: Option<usize>` to `MachineData` and only populates it for AM1. This is technically correct for today's data, but the field is `Option<usize>` (where `None` = unbounded). If a future machine gets a slot limit and the JSON isn't updated, it silently becomes unbounded. Consider whether `ingredient_slots: usize` with a sentinel default (e.g. 4, matching AM2/3) would be safer than `Option`.

5. **`fluid_boxes` stored as `Vec<serde_json::Value>`** — Storing raw JSON values for a field you only check for emptiness is a code smell. It works but creates a maintenance trap: if the prototype shape changes, the empty-check still works but the field carries dead weight. Consider `has_fluid_boxes: bool` derived at parse time, or at minimum a `#[serde(deserialize_with)]` that extracts just the count.

### UX Viability / Analysis

**Strengths:**

6. **Readonly machines section** — Showing Foundry, EM Plant, Chemical Plant, etc. as read-only labels is good UX. It tells the user "these categories exist, here's what handles them" without cluttering the UI with dropdowns they can't change. The `sb-machine-readonly` class is a clean implementation.

7. **Disabled furnaces with tooltip** — `stone-furnace` disabled with "Requires fuel routing — coming later" is exactly the right UX pattern. It shows awareness of the roadmap while preventing confusion.

8. **Config error banner above results** — Routing incompatible-machine errors to a dedicated banner above the result container (rather than burying them in the error region) is smart. The user sees the problem immediately without scrolling past a stale layout.

**Concerns:**

9. **No auto-recovery from incompatibility** — When AM1 is selected and the user picks `advanced-circuit`, the solver fails with `TooManyIngredients`. The banner shows the error, but there's no "auto-suggest AM3" or "change machine" CTA. The user has to manually fix the dropdown. For a tool that's supposed to be approachable, this is a friction point. Consider: on config error, auto-advance the dropdown to the next-capable machine.

10. **Palette doesn't propagate to `defaultMachineForItem`** — The sidebar's `auto-pick on item change` logic still calls `engine.defaultMachineForItem(targetItem, machineSelect.value)` which only considers the crafting dropdown. If the user has `smelting → stone-furnace` in the palette and switches to a smelting item, the assembler dropdown won't auto-update. This is a minor inconsistency but could confuse users who expect the palette to be fully respected.

11. **URL state for unmapped categories** — The `URL_KEY_BY_CATEGORY` only includes `crafting` and `smelting`. If a future PR adds a third editable category, it needs to be added in three places: `EDITABLE_MACHINES` in sidebar.ts, `URL_KEY_BY_CATEGORY` in state.ts, and `DEFAULT_MACHINES` in state.ts. Consider deriving the URL keys from the `EDITABLE_MACHINES` array to reduce duplication.

### Test Coverage

**What's covered well:**

12. **Unit tests for `machine_can_run_recipe`** — The 7 new tests in `recipe_db.rs` cover the happy path (AM3 accepts everything), each incompatibility variant (slots, fluid, category), and the edge case of unknown machines. Good.

13. **Palette override tests** — Tests for `palette_overrides_default_for_crafting`, `palette_miss_falls_through_to_hardcoded`, `palette_miss_falls_through_to_default`, and `palette_does_not_override_hardcoded_when_absent` provide solid coverage of the dispatch logic.

14. **Solver-level integration** — `am1_palette_for_advanced_circuit_returns_incompatible_error` verifies the end-to-end path: palette → solver → typed error. This is the most important test because it validates the error flows through the WASM boundary correctly.

**Gaps:**

15. **No e2e test for the palette path** — All e2e tests go through `solver::solve()` which uses the default palette. There's no e2e test that exercises `solve_with_palette()` through the layout → export → validate pipeline. If the palette changes machine assignments, the layout engine will see different entity types, which could expose layout bugs specific to certain machines (e.g., different inserter patterns for AM1 vs AM3).

16. **No test for the `solve_with_palette` WASM binding** — The `solve_with_palette` function in `wasm-bindings/src/lib.rs` is a thin wrapper. It's not directly tested. The unit test in `solver.rs` covers the Rust path, but the WASM ABI layer (into/from wasm_abi for `MachinePalette`) is untested. If `tsify` generates incorrect serialization for `FxHashMap`, this would be a silent failure.

17. **No test for the URL migration path** — The `readUrlState` back-compat logic that folds `?machine=` into `machines.crafting` is not tested. A regression here would silently break existing bookmarked URLs.

### Correctness

**Issues found:**

18. **`machine_handles_category` is a duplicate of the category mapping in `machine_for_recipe`** — The function mirrors the `match` arms from `machine_for_recipe`. This is a maintenance hazard: if a category is added to one but not the other, they'll diverge. The `machine_for_recipe_with_palette` function already has access to the category mapping; `machine_handles_category` should either be derived from it (e.g., reverse-map category → valid machines) or the validation should be done inline in `machine_can_run_recipe` to avoid duplication.

19. **`FluidNotSupported` check fires for products too** — The PR checks both ingredients and products for fluid compatibility. This is correct (a furnace can't output a fluid), but the error message says "needs fluid X" which is slightly misleading when it's actually the *product* that's fluid. Minor wording issue.

20. **`Unknown machines pass through`** — Unknown machines are treated as unconstrained. The PR notes this is intentional because the solver hits `MissingCraftingSpeed` separately. However, if a user types a hand-edited URL with a nonsense machine name, they'll get past the incompatibility check and hit `MissingCraftingSpeed` instead, which is a different error path. This is acceptable but worth being aware of — the two error types have different user-facing messages.

21. **`machinesAreDefault` rejects unknown keys** — The function returns `false` if any palette key isn't in `URL_KEY_BY_CATEGORY`. This means a hand-edited URL with `?foo=bar` will always produce a non-default URL, causing the key to be written even though it's not a recognized category. This is defensive but could lead to URL pollution. Consider filtering unknown keys silently.

22. **`buildPalette()` in sidebar.ts only includes categories with non-empty selects** — Since the selects always have a value (they default to the first option), this is fine. But the comment says "if (sel.value)" which suggests the author anticipated the possibility of empty selects. The current code is correct but the guard is dead code.

---

### Summary

**Green light with notes.** The PR is well-scoped, backward-compatible, and the core logic is sound. The main areas to address before merge:

1. **Fix the duplicated category mapping** (`machine_handles_category` vs `machine_for_recipe`) — this is the highest-risk correctness issue.
2. **Add at least one e2e test that exercises the palette path** — even a simple one that sets `crafting → assembling-machine-1` and verifies the solver result.
3. **Consider auto-suggesting a compatible machine on config error** — this is a UX improvement that would make the tool more approachable for new users.

The `fluid_boxes: Vec<serde_json::Value>` and `ingredient_slots: Option<usize>` design choices are defensible for now but should be cleaned up in a follow-up once the data model stabilizes.

storkme (commented):
Code-wise this looks ready to merge — CI is green, the diff matches the PR description, tests are thorough (12 new unit tests + the `palette_pins_iron_gear_wheel_to_am1` e2e test added per the previous review), the `category_machines` unification is a real improvement, and the marker-prefix contract is locked in by `incompatible_machine_error_message_carries_marker_prefix`.

**Blocker:** the branch has merge conflicts against `main` (`mergeStateStatus: DIRTY`). The conflict is in `web/src/state.ts` against #262 (short-id URL scheme), which moved `writeUrlState` from `URLSearchParams` to `formatHashState(...)`. Please rebase onto current main and re-run the URL-state migration on top of the new hash-encoding format — in particular:
- Make sure `?craft=`/`?smelt=` round-trip through whatever short-id encoder #262 introduced (or are excluded if those keys aren't compressible yet).
- Re-verify the legacy `?machine=` fallback still works after the rebase.
- Confirm `machinesAreDefault` is still consulted by the new equivalent of the "is default → bare URL" path.

Two minor points (non-blocking, take or leave):

1. `crates/core/src/recipe_db.rs` — `am1_rejects_recipe_with_fluid_ingredient` looks up `plastic-bar` and then immediately discards the lookup with `let _ = recipe`. The synthetic recipe is what's actually exercised. Either drop the lookup, or use it (e.g., assert it has a fluid product) so the comment "keep the lookup as a sanity check on data" is actually true.
2. Categories like `centrifuging`, `rocket-building`, and `captive-spawner-process` aren't in `category_machines`, so they fall through to assembler tiers — same behaviour as the old `machine_for_recipe`, so not a regression, but worth a follow-up issue: `machine_can_run_recipe` will silently approve `assembling-machine-3` for a centrifuging recipe, even though the resulting layout is bogus. Not in scope here, but a `TODO` or issue link would be appropriate.

Once rebased, this is good to merge.

PR #264, opened 2026-04-30
docs(rfp): pluggable layout engine — RowPlacer + RowInfo traits + registry

storkme:
## Summary

Introduce traits and a strategy registry at the layout-engine boundary so that
different row-arrangement strategies can be swapped in and out without editing
`layout.rs`'s 560-line orchestrator or touching the ghost router.

## What this specs

Two traits + one registry:

| Trait | Purpose |
|-------|---------|
| `RowPlacer` | Stack assembly rows vertically, return entities + metadata |
| `RowInfo` | Per-row metadata the lane planner needs (belt positions, direction, etc.) |

`RowSpan` implements `RowInfo`. Today's `place_rows` becomes `VerticalSplitPlacer` behind the trait. Byte-identical output in Phase 1.

## What's out of scope

- Pluggable ghost router (4,300 lines, tightly coupled to `BeltSpec`/`Occupancy`/SAT)
- Pluggable lane planner (`plan_bus_lanes` at 1,100 lines)
- Pluggable solver (already decoupled)

## Kill criteria

1. >400 LOC of new code (excluding moved code) → too heavy
2. >3 files to change for a new row layout → abstraction leaking
3. >20% new test code to validate adapters → not transparent
4. WASM API changes → wrong approach

## Phasing

1. **Phase 1** — `RowPlacer` + `RowInfo` traits + adapter (no behavior change)
2. **Phase 2** — `HorizontalStack` behind a trait
3. **Phase 3** — Canary experiment: implement a trivial new layout (e.g. staggered rows) with zero changes to `ghost_router` or `lane_planner`
4. **Phase 4** — Pluggable lane planner (separate RFP if Phase 3 validates the approach)

---

Draft PR for spec review. Implementation follows in a subsequent PR.

### DIFF
--- a/docs/rfp-pluggable-layout-engine.md
+++ b/docs/rfp-pluggable-layout-engine.md
@@ -1,0 +1,320 @@
+# RFP: Pluggable row placement — `RowPlacer` trait + `RowInfo` trait
+
+## Summary
+
+Introduce two traits at the layout-engine boundary so that different row-arrangement
+strategies can be swapped in without editing `placer.rs` or `templates.rs`.
+The traits sit between `SolverResult` and `LayoutResult`, replacing the current
+monolithic `place_rows` / `build_one_row` functions with a registered strategy.
+
+**Scope:** row placement only. The lane planner (`plan_bus_lanes`) and ghost router
+(`route_bus_ghost`) remain untouched — they consume `RowInfo` instead of `RowSpan`
+directly. This RFP produces a *second `RowSpan`-shaped layout strategy* (e.g.
+staggered rows), not a fundamentally different bus topology. A fully pluggable
+lane planner is a follow-up RFP.
+
+## Motivation
+
+Today `place_rows` and `build_one_row` in `placer.rs` (together ~1,800 lines)
+hardcode two row geometries:
+
+- **VerticalSplit** — input-bottlenecked recipes split into short rows reconciled
+  by an N→M balancer.
+- **HorizontalStack** — one long row with K stacked input belts at the top.
+
+Adding a third geometry means editing `place_rows`, `build_one_row`, and
+`templates.rs` — and the new code can't be tested in isolation from the existing
+placers. The existing `LayoutStrategy` enum (Pooled / PartitionedDecomposed) only
+controls *how many modules* a multi-consumer item gets split into — the spatial
+layout is identical.
+
+We want to experiment with new row arrangements (e.g. staggered rows, compact
+grouping heuristics) without touching the lane planner or ghost router.
+
+## Design
+
+### 1. Core traits (`bus/engine.rs`)
+
+Two traits. Everything else is plumbing.
+
+```rust
+/// A row-placement strategy. Called by the layout orchestrator to stack
+/// assembly rows vertically. Returns entities plus per-row metadata that
+/// the lane planner and router consume.
+pub trait RowPlacer: Send + Sync {
+    /// Placement strategy name (for trace events, URL params).
+    fn name(&self) -> &str;
+
+    /// Place all rows for the given machine specs.
+    ///
+    /// Called twice by `layout_pass` (see §3 below). The first call
+    /// uses an estimated bus width; the second uses the actual width
+    /// computed from the planned lanes. The placer must produce
+    /// consistent geometry given the same inputs.
+    ///
+    /// Returns (entities, per-row metadata, total_width, total_height).
+    fn place_rows(
+        &self,
+        machines: &[MachineSpec],
+        dependency_order: &[String],
+        bus_width: i32,
+        y_offset: i32,
+        max_belt_tier: Option<&str>,
+        final_output_items: &FxHashSet<String>,
+        extra_gaps: &FxHashMap<usize, i32>,
+    ) -> (Vec<PlacedEntity>, Vec<Box<dyn RowInfo>>, i32, i32);
+
+    /// Build a single row (used by `place_rows` internally).
+    fn build_one_row(
+        &self,
+        spec: &MachineSpec,
+        count: usize,
+        bus_width: i32,
+        y_cursor: i32,
+        max_belt_tier: Option<&str>,
+        output_east: bool,
+    ) -> (Vec<PlacedEntity>, Box<dyn RowInfo>, i32);
+}
+
+/// Per-row metadata consumed by the lane planner and ghost router.
+///
+/// `RowSpan` implements this trait. The lane planner reads these fields
+/// to determine where tap-offs land, where output belts flow, how to key
+/// `(item, module_id)` consumer buckets, and whether a row uses horizontal
+/// stacking.
+///
+/// The ghost router reads `output_belt_y`, `output_belt_x_range`, and
+/// `spec.inputs/outputs` (via `inputs()`/`outputs()`) for output merging
+/// and producer-row lookups.
+pub trait RowInfo: Send + Sync {
+    fn y_start(&self) -> i32;
+    fn y_end(&self) -> i32;
+    fn input_belt_ys(&self) -> &[i32];
+    fn output_belt_y(&self) -> i32;
+    fn output_east(&self) -> bool;
+    fn output_belt_x_range(&self) -> (i32, i32);
+    fn horizontal_stack_info(&self) -> Option<&HorizontalStackInfo>;
+    fn fluid_port_ys(&self) -> &[i32];
+    fn fluid_port_pipes(&self) -> &[(String, i32, i32)];
+    fn fluid_output_port_pipes(&self) -> &[(String, i32, i32)];
+    fn machine_count(&self) -> usize;
+    fn module_id(&self) -> u32;
+    fn recipe(&self) -> &str;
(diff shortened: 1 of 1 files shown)


storkme (commented):
**TL;DR**: Request changes. The trait boundary leaks — `RowInfo` claims to hide `RowSpan` but consumers read `rs.spec.inputs`/`outputs` directly today, and the kill criteria are LOC/file-count proxies rather than the falsifiable correctness gates other RFPs in this repo use (`docs/rfp-modular-production.md`, `docs/rfp-decomposition-search.md`). Phase 1 is a no-op refactor; its evidence value is entirely in Phase 3.

### 1. The trait omits the fields the lane planner actually reads

`RowInfo` deliberately leaves out `MachineSpec`. But `plan_bus_lanes` walks it:

- `crates/core/src/bus/lane_planner.rs:181` — `for inp in &rs.spec.inputs` keys `(item, module_id)` consumer buckets.
- `crates/core/src/bus/lane_planner.rs:220` — `for out in &rs.spec.outputs` computes lane rates from `out.rate * rs.machine_count` and uses `out.is_fluid`.
- `crates/core/src/bus/lane_planner.rs:660,830`, `crates/core/src/bus/ghost_router.rs:2655` — more `rs.spec.inputs/outputs` walks.

Either the trait grows `inputs()`/`outputs()` (re-exposing `MachineSpec`, no encapsulation win) or the lane planner keeps reading `RowSpan` (which §5 concedes). The latter means a `StaggeredRowPlacer` cannot avoid producing `RowSpan`-shaped data — the trait is just a name for "produce a `Vec<RowSpan>`." Show how a Phase-3 placer returns `RowInfo` without a `RowSpan` behind it, or admit the boundary is cosmetic.

Also: `RowLayout` enum (`crates/core/src/bus/layout.rs:37-48`) is *already* plumbed through `LayoutOptions` to `place_rows`. `HorizontalStack` shipped via a match arm in `build_one_row` — no trait required. The RFP should explain why a third variant can't follow the same path.

### 2. Kill criteria are proxies, not failure modes

Compare to `rfp-modular-production.md` ("inertness", phantom belts in specific snapshots). LOC/file-count/test-ratio criteria are leading indicators of sloppy code, not falsifiable evidence the abstraction is wrong. 401 LOC isn't the failure; landing at 200 LOC where Phase 3 still has to edit `lane_planner.rs` is.

Sharper, in-house style:

- *"Phase 3's canary placer must not edit `lane_planner.rs`, `ghost_router.rs`, or `templates.rs`. If it does, the trait is leaking and we abandon."*
- *"Phase 1 byte-equivalence: every tier1–4 e2e snapshot's entity list is unchanged. Diverge → revert."*
- *"A new placer must be able to produce a row with non-canonical `output_east`/`output_belt_x_min/max` without edits outside the trait module."*

`rfp-modular-production.md` explicitly rejects entity-count-style gates for the same reason — drop the LOC/file gates.

### 3. Phase 1 is a no-op; merge with Phase 3 or argue the case

Phase 1 = rename `place_rows` to `VerticalSplitPlacer::place_rows`. Zero correctness signal; the only evidence the abstraction works is Phase 3's `StaggeredRowPlacer`. Land them together, or note in the RFP why a no-op refactor is worth its own review (bisectability is the obvious argument, but there's nothing to bisect against until Phase 3).

### 4. Excluding the lane planner from scope contradicts the motivation

The motivation cites "grid bus, multi-axis bus, machine-grouping heuristics." All of those change the `(item, module_id) → consumer rows` shape that `plan_bus_lanes` builds (`lane_planner.rs:179-226`). Excluding the lane planner restricts Phase 3 to "rows shaped like `RowSpan`" — which is the *existing* `RowLayout::HorizontalStack`, spelled differently. Either pull lane-planner Phase 4 into this RFP or rephrase the motivation: "we want a second `RowSpan`-shaped layout strategy" is honest.

### 5. Smaller

- §3 omits the two-pass placement dance: `place_rows` is called twice in `layout_pass` (see `crates/core/src/bus/layout.rs:302` and `:344`, with different `bus_width` and merged retry/balancer gaps). The trait must support re-placement on actual width.
- §4 says `RowSpan` stays in `placer.rs` and implements `RowInfo` from `engine.rs` — state the module DAG so the circular-dep risk is visible.
- Decision log has only the "initial spec written" entry. Add a line for whichever of §3's options you take.

storkme (commented):
**TL;DR**: Most prior blockers are resolved; one new wrinkle (the "DAG" is a cycle) and one residual ambiguity in §3 (two-pass dance is "up to twice", not "twice"). Would request changes if not own PR — defer to author on whether to address inline or in implementation.

| # | Prior point | Status |
|---|---|---|
| 1 | Trait boundary leaks (`rs.spec.inputs/outputs`) | resolved |
| 2 | `RowLayout` enum already plumbed | resolved |
| 3 | Kill criteria are LOC/file-count proxies | resolved |
| 4 | Phase 1 is a no-op refactor | resolved |
| 5 | Lane planner exclusion contradicts motivation | resolved |

### Still open / new

**§4 Module DAG — claimed acyclic, actually cyclic.** The arrow set the RFP lists is `engine → placer`, `placer → layout`, `layout → engine`. That's a 3-cycle, not a DAG. Rust permits mutual sibling-module imports inside a crate so it'll compile, but the RFP states "This is a clean DAG: no cycles" which is the opposite of what it draws. Either invert one edge (e.g. move `RowLayout` into `engine.rs` so `placer` no longer imports `layout`, breaking `placer → layout`) or just rewrite §4 as "acyclic at the type level, with `engine.rs` re-exporting `RowLayout`." Today `placer.rs:7` already does `use crate::bus::layout::RowLayout`, so the cycle is real if `layout.rs` then imports from `engine.rs` which imports from `placer.rs`.

**§3 / §1 two-pass framing.** Both the doc-comment on `RowPlacer::place_rows` and the "Two-pass placement" callout say `place_rows` is "called twice." It's actually called *up to* twice — `layout.rs:330` skips pass 2 when `actual_bw == temp_bw && balancer_gaps.is_empty()`. Worth a one-line correction so a placer impl doesn't assume both calls always fire (e.g. caching state in `&self` keyed on call-count would be wrong).

**Kill criterion 3 ("non-canonical geometry") is good but soft.** "Without edits outside the `engine.rs` module" — `engine.rs` is the new home for both the trait and `VerticalSplitPlacer`'s wiring per §1, so a canary placer would naturally also live there. The criterion is satisfied trivially as written. Tightening to "no edits to `lane_planner.rs`, `ghost_router.rs`, `templates.rs`, or `placer.rs`" matches the spirit of criterion 2 and is harder to game.

**Phase 3 (follow-up RFP) numbering collision.** Phasing now lists "Phase 1+3" then "Phase 2" then "Phase 3 (follow-up RFP)" — two different things called Phase 3. Rename the lane-planner follow-up to "Phase 4" (matching the original numbering) or "Future work" to avoid confusion when someone refers to "Phase 3" mid-implementation.

**Decision log entry is good.** Concise, dated, lists which review point each change addresses. Keep this style for subsequent updates.

The trait-boundary resolution is honest: §1 admits `inputs()`/`outputs()` expose `ItemFlow` (a real subset of `MachineSpec`), and the prose argues this is enough for a non-`RowSpan`-backed placer to comply. Acceptable — the encapsulation is narrow but real.

storkme (commented):
**TL;DR**: All 3 prior issues still open verbatim — no commits since `pullrequestreview-4208728241` (HEAD still `e76b0a71`, `updatedAt` matches the review timestamp). Would request changes if not own PR.

| # | Issue | Status |
|---|---|---|
| 1 | §4 calls a 3-cycle a "clean DAG" | open |
| 2 | §3 / §1 say `place_rows` called "twice" (actually up-to-twice) | open |
| 3 | Kill criterion 3 gameable | open |
| 4 | Phase 3 label collision | open |

### Still open

**§4 cycle (lines 195-204).** Unchanged. Edges `engine → placer`, `placer → layout`, `layout → engine` and the assertion "This is a clean DAG: no cycles" still contradict each other on the page. `placer.rs:7` (`use crate::bus::layout::RowLayout`) still confirms the `placer → layout` edge is real today.

**§3 / §1 "called twice" (lines 50, 187-188).** Both the `RowPlacer::place_rows` doc comment and the "Two-pass placement" callout still say "called twice" / "calls `place_rows` twice". `layout.rs:330` short-circuits when `actual_bw == temp_bw && balancer_gaps.is_empty()`, so it's *up to* twice. A placer caching state keyed on call-count would still be wrong.

**Kill criterion 3 (lines 245-248).** Still "without edits outside the `engine.rs` module." Since §1 puts `VerticalSplitPlacer` in `engine.rs`, a canary placer added there satisfies the gate trivially. Tighten to forbid edits to `lane_planner.rs`, `ghost_router.rs`, `templates.rs`, *and* `placer.rs` to match criterion 2's spirit.

**Phase numbering (lines 268, 288, 296).** "Phase 1+3 (merged)" + "Phase 3 (follow-up RFP)" still collide. Rename the follow-up to Phase 4 / Future work.

No new issues introduced — there's no new diff. Would approve if not own PR once these four wording fixes land; they're all single-line edits in the doc.

PR #269, opened 2026-05-01
Phase 2.0: Runtime balancer template generator + classifier

storkme:
## Intent

Implement phase 2.0 of the throughput-priority merges project: a Rust-side runtime template generator that produces verified MX2/MX3 balancer layouts for divisible `(m, n)` shapes, plus a comprehensive classifier that assigns any balancer template to the MX1/MX2/MX3 taxonomy. This closes the audit phase ([`rfp-throughput-priority-merges.md`](docs/rfp-throughput-priority-merges.md)) and enables phase 2 generation ([`rfp-mx2-merge-generator.md`](docs/rfp-mx2-merge-generator.md)).

## Scope

**In:**
- `balancer_classify.rs` — classifier that recovers the logical splitter graph from a template, computes composition via linear-system saturation model, and checks MX1/MX2a/MX2b/MX3 properties via max-flow (Menger's theorem). Public API: `classify()`, `classify_ref()`, `classify_graph()`, `topology_of_template()`.
- `balancer_topology.rs` — logical topology generators for splitter graphs (phase 3.0 scope): `passthrough()`, `one_to_two()`, `two_to_one()`, `parallel()`, `series()`, `series_permuted()`, `clos_interleave()`, plus bootstrap atoms `library_atom()` / `library_atom_strict()` to extract topologies from library templates.
- `balancer_generate.rs` — phase 2.0 runtime generator: produces `OwnedTemplate` for divisible shapes (`m == n`, `(1, 2)`, `(2, 1)`, `(m, k·m)`, `(k·m, m)` for `k=2`). Every generated template is verified by `classify_ref()` before return.
- `crates/balancer-gen/` — phase 3.1 spike binary: drives Python+OR-Tools CP-SAT subprocess for splitter no-overlap placement, emits entity lists, round-trips through classifier to confirm topology preservation.
- Documentation: `rfp-throughput-priority-merges.md` (audit phase spec), `rfp-mx2-merge-generator.md` (phase 2 spec), `rfp-balancer-graph-place.md` (phase 3 spec), `rfp-balancer-place-routing.md` (phase 3.2 routing spec), `factorio-mechanics.md` (belt merger taxonomy section).
- Integration: `balancer.rs` updated to call `generate()` as final fallback before giving up on a shape; `trace.rs` extended with `BalancerGenerated` event.

**Out:**
- Phase 3.2+ (full placement solver with belt routing) — deferred to follow-up RFP. The spike confirms CP-SAT framework is solid; the remaining work is router complexity.
- Coprime universal-balancer topologies (back-loop constructions) — phase 3.1+ scope. Phase 2.0 handles only divisible shapes.
- Lane-level semantics (MX5) — orthogonal concern, documented but not implemented.

## Verification

- `cargo test --manifest-path crates/core/Cargo.toml` — all tests pass, including new classifier unit tests (`test_classify_*`, `test_composition_*`, `test_mx2_*`).
- `cargo clippy -- -D warnings` — no warnings.
- `cargo build --manifest-path crates/balancer-gen/Cargo.toml` — spike binary builds (requires Python + OR-Tools in PATH for runtime execution).
- Classifier round-trips on all 63 library templates: `cargo test --lib audit_report -- --nocapture` produces taxonomy report (60 MX3, 3 MX2a, 0 MX1 in current library).
- Generator produces verified MX2/MX3 for divisible shapes: `cargo test --lib test_generate_*` confirms `(2, 4)`, `(3, 6)`, `(4, 2

https://claude.ai/code/session_01YJrsb6wv1F8D1tggCvk9Y4


crates/core/src/bus/balancer_generate.rs
```diff
@@ -0,0 +1,420 @@
...
+            x: 0,
+            y: 2,
+            direction: 4,
+            io_type: None,
+        },
+    ];
+    OwnedTemplate {
+        n_inputs: 2,
+        n_outputs: 1,
+        width: 2,
```
storkme, line 252: The "throughput capped at 1, MX3 by classifier" claim relies on Factorio splitter saturation behaviour: with one output tile empty (no entity at `(1, 2)`), items can't go there, the splitter back-pressures and balances 50/50 between input belts in steady state. That works in real Factorio, but it's worth noting that this is *exactly* the kind of construction the all-fluid verifier in #270 will reject — the dangling out-arc forces an internally-saturated solution, which is why #270's `verify_balancer` would diverge from this classifier on any layout that uses `two_to_one`.

Net: the layout is correct in-game; the classifier is being honest about the saturated-50/50 model; #270's verifier disagreement is real but pointed at a different question. Worth a one-liner note here so future readers cross-checking against the verifier don't think this is broken.

Tangent: `replicate_horizontally(two_to_one(), n)` produces `n` separately-merging columns with no cross-mix — same downside as the `(1,2)` replicate path that you correctly flag as MX2a in `parallel_one_to_two_for_m_to_2m`. Suggest an analogous `parallel_two_to_one_for_2m_to_m` test pinning the `(2m, m)` class so a future bug that elevates them to MX3 is caught.

---

docs/factorio-mechanics.md
```diff
@@ -74,6 +74,29 @@ Formal rules the layout engine must satisfy. Statements are numbered per section
...
+
+- **MX3. Balanced (composition-balanced).** Stronger than MX2b: every output belt's content is exactly 1/m drawn from each input belt, regardless of partial saturation. Required only when belts carry mixed item types and downstream consumers depend on the mix (e.g. mixed-bus science loading). *Diagnostic*: the linear flow-rate system, with each splitter modeled as a 50/50 distributor, produces output rates that are uniform combinations of the input rates. **Strictly subsumes MX2b.** Significantly larger than MX2 for asymmetric (m, n).
+
+- **MX4.** For single-recipe-per-trunk bus layouts (fucktorio's design), every belt carries one fungible item type. **MX2 is sufficient**; MX3's composition guarantee is not load-bearing. The current `balancer_library` is generated by Factorio-SAT and is *intended* as MX3, but per-template classification has not been audited — see [`docs/rfp-throughput-priority-merges.md`](rfp-throughput-priority-merges.md).
+
```
storkme, line 93: The MX2a / MX2b split here is load-bearing for this PR (the generator targets MX2a explicitly and the classifier exposes both as enum variants), but #270 has the same file open with the older unified-MX2 wording and no `MX2a` enum variant. Whichever PR lands second will need to re-take this taxonomy section + the matching `BalancerClass` enum variants. Worth coordinating before merge — #270's classifier doesn't lose anything by adopting the finer split, but #269's `BalancerClass::ThroughputBalancedRate` would have to be re-introduced if #270 lands first.

---

crates/balancer-gen/src/main.rs
```diff
@@ -0,0 +1,283 @@
...
+        input_tiles,
+        output_tiles,
+    })
+}
+
+fn run_solver(
+    n_splitters: usize,
+    bounds: (u32, u32),
+) -> Result<PlaceResponse, Box<dyn std::error::Error>> {
+    let req = PlaceRequest { n_splitters, bounds };
```
storkme, line 240: Two things on the subprocess shape:

1. `Command::new("python3")` assumes a system python with `ortools` installed. #270's `scripts/cp_sat_placer.py` uses a PEP 723 inline-script header + `uv run --no-project`, which self-installs deps on first run and pins them at the script. That's the cleaner pattern for a new external dep — would be worth either adopting that pattern here or consolidating both Python entrypoints behind a single placer script.

2. The output of this binary doesn't feed anything in `crates/core/`. It's a standalone validation harness. Once #270's `PlacementEngine` trait lands (or once this PR's spike graduates), the natural shape is "spike binary calls into a real engine implementation", not "spike binary holds the only path that exercises CP-SAT placement". Worth marking the deferred merge of these two integration patterns explicitly somewhere — the `crates/balancer-gen/` crate name even collides namespace-wise with #270's new top-level `crate::balancer::` module.

---

crates/core/src/bus/balancer.rs
```diff
@@ -344,17 +378,17 @@ mod tests {
...
+    /// Phase-2.0 generator closes `(3, 9)` (3 × library `(1, 3)`),
+    /// `(6, 9)` (3 × library `(2, 3)`), and `(9, 9)` (passthrough) — those
+    /// are removed from the fixture below. Remaining gaps need either
+    /// missing library templates or richer generator atoms.
     #[test]
     fn shape_is_stampable_pins_known_gaps() {
         let known_gaps: &[(u32, u32)] = &[
-            (1, 9), (2, 9), (3, 9), (4, 9), (5, 9), (6, 9), (7, 9), (8, 9),
-            (9, 1), (9, 2), (9, 3), (9, 4), (9, 5), (9, 6), (9, 7), (9, 8), (9, 9),
+            (1, 9), (2, 9), (4, 9), (5, 9), (7, 9), (8, 9),
```
storkme, line 390: This is the actual ship-able win in the PR — three coprime gaps closed (`(3,9)`, `(6,9)`, `(9,9)`) without touching the balancer library or Factorio-SAT. Worth calling out more loudly in the PR body: the headline value of phase 2.0 is not the runtime generator framework, it's that 17→14 of the documented coprime gaps from #136 disappear today, plus passthroughs for `(m, m)` up to `m=10`.

---

storkme (commented):
Reviewed alongside #270 — both PRs are forks of the same audit-phase tip (cherry-picked the taxonomy doc, classifier, and sideload-relax commits) and have diverged in interesting and partially-conflicting ways. Capturing the cross-PR observations here rather than splitting them across both:

**Where this PR wins**: it actually closes three of the documented coprime gaps from #136 (`(3,9)`, `(6,9)`, `(9,9)`) and adds passthrough coverage for `(m, m)` up to `m=10`, all wired through `stamp_family_balancer` as a real fallback. That's a concrete user-visible improvement and self-verified by `classify_ref` before stamping. The PR body underplays this — phase 2.0 ships layouts today, regardless of what happens with phases 3.x.

**Hard merge conflicts with #270**, all in shared cherry-picked files:

- `crates/core/src/bus/balancer_classify.rs` — this PR adds `BalancerTemplateRef<'a>`, `classify_ref()`, `classify_graph()`, `topology_of_template()`, the `MX2a/MX2b` enum split, and keeps `recover_graph` private. #270 makes `recover_graph` public, deletes the `Ref` indirection, and uses `&BalancerTemplate` directly with no `MX2a/MX2b`. These two API surfaces are mutually incompatible, and #270's design loses several capabilities this PR depends on (`classify_graph` for graph-only classification, `topology_of_template` for the spike binary). Suggest #270 adopt this PR's API as the merge target.
- `docs/factorio-mechanics.md` — MX2a / MX2b split here vs. unified MX2 in #270 (see inline).
- `docs/rfp-throughput-priority-merges.md` — identical between branches, will conflict-as-equal.

**CP-SAT integration drift**: this PR drives a Python subprocess from a separate `crates/balancer-gen/` binary using bare `python3`, while #270 drives one from a `PlacementEngine` trait inside `crates/core/` using PEP 723 + `uv run --no-project`. The trait shape in #270 is the cleaner long-term home; the spike binary here is the only thing currently exercising CP-SAT against non-trivial topologies (Clos composition, single-splitter MVP round-trip). They should converge — not in this PR necessarily, but with an explicit "spike binary will fold into the trait once shape coverage lands" note in the phase 3.2 RFP.

**Verifier divergence (forward-looking)**: #270's all-fluid Couëtoux verifier will reject any layout that depends on internally-saturated steady state — including this PR's `two_to_one()` construction (see inline). That's not a bug in either piece, but the cross-validation test in #270 currently asserts `disagreements.is_empty()` (modulo an env override), so any generator-emitted `two_to_one`-using template would push that count above zero once #270 lands. Worth flagging in the phase-3 RFP.

Inline comments on the spike binary, the saturated-merge construction, taxonomy drift, and the actual ship-this-now value point in `balancer.rs`.

Marking as comment, not requesting changes — the work in this PR is solid; the friction is purely about coordinating with the parallel #270 stream.

---

PR #273, opened 2026-05-01
CP-SAT placement engine: phase 1 (dyadic shapes through (1, 8))

storkme:
## Intent

Phase 1 of [docs/rfp-cp-sat-placement.md](https://github.com/storkme/fucktorio/blob/main/docs/rfp-cp-sat-placement.md): get the CP-SAT placement engine producing valid Factorio templates that round-trip through `bus::balancer_classify::topology_of_template` to a balanced splitter graph. Validates the entity-emission path end-to-end so phases 2+ can replace hardcoded geometry with real CP-SAT spatial models without re-doing the integration plumbing.

## Scope

**In:** 10 dyadic shapes covered by the placer, all round-trip-tested via `tests/cp_sat_round_trip.rs`:

| `(n, m)` | Strategy | Splitters | Footprint |
|---|---|---|---|
| `(1, 1)` | Hardcoded passthrough | 0 | 1×1 |
| `(1, 2)`, `(2, 1)`, `(2, 2)` | Hardcoded single-splitter | 1 | 2×3 |
| `(1, 4)`, `(2, 4)` | CP-SAT structural model, tight-stack tree | 3 | 4×4 |
| `(1, 8)`, `(2, 8)` | CP-SAT, +1 routing row | 7 | 8×6 |
| `(1, 16)`, `(2, 16)` | CP-SAT, +2 routing rows | 15 | 16×8 |

`scripts/cp_sat_placer.py` dispatches via a `(n, m) → emit_template` table; new shapes are a one-line addition.

**Lane-aware routing infrastructure** (phases 1-3 of `docs/rfp-lane-aware-routing.md`):
- Per-tile per-direction per-lane flow variables.
- Sideload-direction table (LEFT/RIGHT relative to belt direction).
- Dual-lane input modeling — input boundary belts emit 2 lane-routes so per-lane caps catch saturation. `(3, 4)` correctly UNSAT now.
- Rate-aware per-lane caps (`sum rates[r] * f_lane ≤ lane_cap`); defaults preserve discrete behavior, but the encoding is ready for fractional rates.
- Splitter direction support for non-south splitters.

`PlacedTemplate::into_owned_template` converts the JSON wire format's `String` entity names to the `&'static str` literals the classifier requires. Errors on unknown names rather than silently dropping entities.

**Out:**
- Coprime shapes (`(1, 5)`, `(1, 7)`, `(4, 9)`, etc.). Blocked on source-lane forcing for splitter-output drops — see `docs/rfp-lane-aware-routing.md` decision log.
- Asymmetric `(n, m)` for `n > 2`. Same blocker.
- `(n, n)` Beneš shapes. `synth_benes` is implemented and tested but has no consumer; deferred to phase 4 alongside underground belt support.
- Real CP-SAT belt routing in the spike at `crates/balancer-gen/scripts/place.py`. That script (UG belts, mixed splitter directions, edge-aware) stays as-is; once the canonical placer at `scripts/cp_sat_placer.py` covers UG belts, the spike migrates over.

## Design notes / RFPs in this PR

- `docs/rfp-lane-aware-routing.md` — the lane-aware-routing RFP this PR is implementing. Decision log records phases 1-3 shipped; `(1, 5)` integration parked pending source-lane forcing for splitter-output drops.
- `docs/rfp-lane-safe-synth.md` — a follow-up RFP that proposed pushing the merging upstream into synth (cascade balancers + self-loop terminators). Tested empirically before commit: single-splitter self-loops verify fluid-stable, but multi-splitter cascades for K→1 mergers re-introduce the multi-arc relaxation we were trying to eliminate. **Parked, not promoted.** Decision log captures the test outcome and why the approach doesn't generalize.
- `docs/scratch-lane-balancer-design.md` — earlier subagent design for in-placer balancer insertion. Subsumed by the lane-safe-synth attempt and its retraction.

## Verification

- [x] `cargo test --manifest-path crates/core/Cargo.toml` — 520+ lib tests + cross-validation + bench-style integration tests all green.
- [x] `tests/cp_sat_round_trip.rs` — 10/10 pass with `FUCKTORIO_RUN_CP_SAT=1` (gated off in CI).
- [x] `cargo clippy --manifest-path crates/core/Cargo.toml --all-targets` — no new warnings on touched files.
- [x] Manual JSON round-trip tested for each shape via `echo '...' | uv run scripts/cp_sat_placer.py`.
- [x] Self-loop test bench (run-once, deleted before commit) confirmed `verify_balancer` is full-rank under self-loops on a single splitter; same test surfaced the multi-splitter cascade limitation that retracts the lane-safe-synth RFP.
- [ ] WASM rebuild + browser eyeball — N/A, this engine is build-time only and the runtime path is unaffected.

## Deviations from intent

- Scope grew from "6 shapes" in the original RFP phase 1 to 10. `(1, 16)`, `(2, 4)`, `(2, 8)`, `(2, 16)` were added incrementally as the structural CP-SAT model proved out, plus `(2, 1)` and `(2, 2)` from the spike binary fold-in.
- Lane-aware infrastructure (per-lane vars, sideload table, rate-aware encoding) shipped *ahead* of the coprime shapes that consume it. It's the right foundation for `(1, 5)` and beyond, but the dyadic shapes that ship in this PR don't strictly need any of it. **A reviewer flagged this as "exploration overrunning its evidence" and the call is fair** — see the lane-aware-routing decision log for the chronological context. The code stays because the next concrete piece of work (source-lane forcing) builds on it; rolling back would just mean re-introducing it on the next branch.

## Models / contracts touched

- `docs/rfp-cp-sat-placement.md` — decision log entry recording phase 1 ship.
- `docs/rfp-lane-aware-routing.md` — new RFP for the lane-aware extension.
- `docs/rfp-lane-safe-synth.md` — new RFP, parked after empirical test.
- `docs/scratch-lane-balancer-design.md` — exploratory design doc.
- No changes to `factorio-mechanics.md` or other contracts.

https://claude.ai/code/session_01PrUmYF6ZUEKvDvF4EGJ59x

---
_Generated by [Claude Code](https://claude.ai/code/session_01PrUmYF6ZUEKvDvF4EGJ59x)_

### DIFF
--- a/crates/core/src/balancer/synth.rs
+++ b/crates/core/src/balancer/synth.rs
@@ -39,6 +39,8 @@ use crate::balancer::graph::{Arc, BalancerGraph, Sink, Source};
 pub enum SynthError {
     #[error("invalid shape: ({n}, {m}) — both must be ≥ 1")]
     InvalidShape { n: u32, m: u32 },
+    #[error("Beneš construction requires n to be a power of two; got {n}")]
+    BenesNonPowerOfTwo { n: u32 },
 }
 
 /// Synthesize a load-balancer for `n` inputs and `m` outputs.
@@ -158,6 +160,168 @@ fn synth_n_to_m(n: u32, m: u32) -> BalancerGraph {
     let g = BalancerGraph::new(n, m, total_splitters, arcs);
     debug_assert!(g.validate().is_ok());
     g
+}
+
+/// Synthesize a symmetric Beneš `(n, n)` network for `n = 2^k`.
+///
+/// The classic Beneš construction (Beneš 1965; see Waksman 1968 for the
+/// rearrangeable proof) builds an `n × n` switching network from
+/// `(2k - 1)` stages of `n/2` 2×2 switches each, total
+/// `(2k - 1) · n/2` splitters.
+///
+/// Recursive structure:
+///   - **Base** (`k = 1`): one 2×2 splitter.
+///   - **Step**: `n/2` input-stage splitters fan out to two
+///     `n/2`-Beneš sub-networks (top via out-port 0, bottom via out-port 1);
+///     `n/2` output-stage splitters merge the two sub-networks (top via
+///     in-port 0, bottom via in-port 1).
+///
+/// Every splitter has exactly two in-arcs and two out-arcs — no
+/// sideloading, no underdegree. Compared to the tree-with-sideloads form
+/// of [`synth`], this trades a higher splitter count (e.g. 6 vs 3 for
+/// `(4, 4)`) for a cleaner placement target.
+///
+/// Errors with [`SynthError::BenesNonPowerOfTwo`] if `n` is not a
+/// power of two; for `n = 1` returns the trivial passthrough.
+pub fn synth_benes(n: u32) -> Result<BalancerGraph, SynthError> {
+    if n == 0 {
+        return Err(SynthError::InvalidShape { n, m: n });
+    }
+    if !n.is_power_of_two() {
+        return Err(SynthError::BenesNonPowerOfTwo { n });
+    }
+    if n == 1 {
+        return Ok(synth_n_to_one(1));
+    }
+    let k = n.trailing_zeros();
+    let mut b = BenesBuild::default();
+    let (input_sinks, output_sources) = b.build(k);
+    debug_assert_eq!(input_sinks.len(), n as usize);
+    debug_assert_eq!(output_sources.len(), n as usize);
+    let mut arcs = b.arcs;
+    for (i, sink) in input_sinks.into_iter().enumerate() {
+        arcs.push(Arc {
+            src: Source::Input(i as u32),
+            dst: sink,
+        });
+    }
+    for (j, src) in output_sources.into_iter().enumerate() {
+        arcs.push(Arc {
+            src,
+            dst: Sink::Output(j as u32),
+        });
+    }
+    let g = BalancerGraph::new(n, n, b.n_splitters, arcs);
+    debug_assert!(g.validate().is_ok());
+    Ok(g)
+}
+
+#[derive(Default)]
+struct BenesBuild {
+    arcs: Vec<Arc>,
+    n_splitters: u32,
+}
+
+impl BenesBuild {
+    fn alloc(&mut self) -> u32 {
+        let i = self.n_splitters;
+        self.n_splitters += 1;
+        i
+    }
+
+    /// Build a Beneš `(2^k, 2^k)` sub-network. Returns the in-port sinks
+    /// (length `2^k`, one per logical input of the sub-network) and the
+    /// out-port sources (length `2^k`, one per logical output). Caller
+    /// wires these to the surrounding context (graph I/O at the top level,
+    /// outer-stage splitter ports for nested calls).
+    fn build(&mut self, k: u32) -> (Vec<Sink>, Vec<Source>) {
+        debug_assert!(k >= 1);
+        if k == 1 {
+            let s = self.alloc();
+            return (
+                vec![
+                    Sink::Splitter { idx: s, port: 0 },
+                    Sink::Splitter { idx: s, port: 1 },
+                ],
+                vec![
+                    Source::Splitter { idx: s, port: 0 },
+                    Source::Splitter { idx: s, port: 1 },
+                ],
+            );
+        }
+        let n = 1u32 << k;
+        let half = (n / 2) as usize;
+
+        let in_stage: Vec<u32> = (0..half).map(|_| self.alloc()).collect();
+        let (top_in, top_out) = self.build(k - 1);
+        let (bot_in, bot_out) = self.build(k - 1);
+        let out_stage: Vec<u32> = (0..half).map(|_| self.alloc()).collect();
+
+        // Input stage → sub-networks: out0 to top, out1 to bottom.
+        for i in 0..half {
+            self.arcs.push(Arc {
+                src: Source::Splitter {
+                    idx: in_stage[i],
+                    port: 0,
+                },
+                dst: top_in[i],
+            });
+            self.arcs.push(Arc {
+                src: Source::Splitter {
+                    idx: in_stage[i],
+                    port: 1,
+                },
+                dst: bot_in[i],
+            });
+        }
+        // Sub-networks → output stage: top to in0, bottom to in1.
+        for i in 0..half {
+            self.arcs.push(Arc {
+                src: top_out[i],
+                dst: Sink::Splitter {
+                    idx: out_stage[i],
+                    port: 0,
+                },
+            });
+            self.arcs.push(Arc {
+                src: bot_out[i],
+                dst: Sink::Splitter {
+                    idx: out_stage[i],
+                    port: 1,
+                },
+            });
+        }
+
+        let inputs: Vec<Sink> = (0..half)
+            .flat_map(|i| {
+                [
+                    Sink::Splitter {
+                        idx: in_stage[i],
+                        port: 0,
+                    },
+                    Sink::Splitter {
+                        idx: in_stage[i],
+                        port: 1,
+                    },
+                ]
+            })
+            .collect();
+        let outputs: Vec<Source> = (0..half)
+            .flat_map(|i| {
+                [
+                    Source::Splitter {
+                        idx: out_stage[i],
+                        port: 0,
+                    },
+                    Source::Splitter {
+                        idx: out_stage[i],
+                        port: 1,
+                    },
+                ]
+            })
+            .collect();
+        (inputs, outputs)
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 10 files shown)


scripts/cp_sat_placer.py
```diff
@@ -74,6 +359,391 @@ def place_one_to_one() -> dict[str, Any]:
...
+    model.add(xs[2] == xs[0] + 1)
+    model.add(ys[1] == ys[0] + 1)
+    model.add(ys[2] == ys[0] + 1)
+
+    model.add(ys[0] >= 1)
+    model.add(ys[0] + 2 <= height - 1)
+    model.add(xs[0] - 1 >= 0)
+    model.add(xs[0] + 2 <= width - 1)
+
+    solver = cp_model.CpSolver()
```
storkme: `timeout_ms` and `seed` from the request are not wired into any of the `cp_model.CpSolver()` instances in this PR — none set `solver.parameters.max_time_in_seconds`, `random_seed`, or `num_search_workers = 1`. This is actually a regression: the deleted (1, 1) handler in `main()` *did* configure both (`solver.parameters.max_time_in_seconds = timeout_ms / 1000.0`, `solver.parameters.random_seed = int(seed)`); the refactor to per-shape `place_*` functions dropped that wiring.

Functional consequences:
- The round-trip test passes `seed: Some(42)` and `timeout: Duration::from_secs(10)`. Both are no-ops as written. `Duration::from_secs(10)` in particular gives a false sense of bounded behavior on shapes that may eventually grow harder.
- Pre-trips kill criterion #5 in `rfp-cp-sat-placement.md` ("CP-SAT solve nondeterministic across same-seed runs") — same-seed runs are guaranteed not reproducible because no seed reaches the solver.

Fix is local: thread `timeout_ms` / `seed` from `main()` down into each `place_*` and apply `solver.parameters.random_seed = int(seed or 0)`, `num_search_workers = 1`, `max_time_in_seconds = timeout_ms / 1000` before each `solver.solve()`. Worth doing in this PR while determinism is still claimed in verification.

scripts/cp_sat_placer.py
```diff
@@ -74,6 +359,391 @@ def place_one_to_one() -> dict[str, Any]:
...
+        tiles = [direct_left, direct_right, sideload]
+        routes = [
+            (direct_left, direct_left, DIR_S),
+            (direct_left, direct_left, DIR_S),
+            (direct_right, direct_right, DIR_S),
+            (direct_right, direct_right, DIR_S),
+            (sideload, direct_left, DIR_S),
+        ]
+        return tiles, routes
+    raise ValueError(f"unsupported input count {n}; expected 1, 2, or 3")
```
storkme: `n == 3` here has no caller — the geometry dispatch table at the bottom of the file lists no `(3, *)` shapes (the (3, 4) entry is commented out as deferred), and no other place function calls `_input_routes_for_root(*, *, 3)`. The branch is shipped for a future `place_x_to_three`, but as the diff stands today it's a fairly subtle path that a refactor could break silently.

Suggest either (a) deleting this branch and re-adding when (3, m) lands, or (b) covering it with a unit test against the routes/tiles it produces, so the structure is exercised even though no shape function uses it yet.

scripts/cp_sat_placer.py
```diff
@@ -88,28 +758,30 @@ def main() -> int:
...
+        (2, 2): lambda: place_single_splitter(2, 2),
+        (1, 4): lambda: place_x_to_four(1),
+        (2, 4): lambda: place_x_to_four(2),
+        # (3, 4) deferred: per-lane cap correctly rejects it. Needs a
+        # structural mitigation (lane-balancer splitter) — phase 3.
+        (1, 8): lambda: place_x_to_eight(1),
+        (2, 8): lambda: place_x_to_eight(2),
+        (1, 16): lambda: place_x_to_sixteen(1),
+        (2, 16): lambda: place_x_to_sixteen(2),
+    }
```
storkme, line 865: Worth syncing the PR description with reality here — the description says scope is 6 shapes ((1,1)/(1,2)/(2,1)/(2,2)/(1,4)/(1,8)), but this dispatch table also includes (2,4)/(2,8)/(1,16)/(2,16), and `tests/cp_sat_round_trip.rs` exercises all 10. Not a blocker; just makes it harder for the next reviewer to know what's actually covered.

crates/core/src/balancer/synth.rs
```diff
@@ -160,6 +162,168 @@ fn synth_n_to_m(n: u32, m: u32) -> BalancerGraph {
...
+///     in-port 0, bottom via in-port 1).
+///
+/// Every splitter has exactly two in-arcs and two out-arcs — no
+/// sideloading, no underdegree. Compared to the tree-with-sideloads form
+/// of [`synth`], this trades a higher splitter count (e.g. 6 vs 3 for
+/// `(4, 4)`) for a cleaner placement target.
+///
+/// Errors with [`SynthError::BenesNonPowerOfTwo`] if `n` is not a
+/// power of two; for `n = 1` returns the trivial passthrough.
+pub fn synth_benes(n: u32) -> Result<BalancerGraph, SynthError> {
```
storkme, line 191: `synth_benes` is added with full docs and tests but has no caller anywhere in this PR — the new round-trip suite uses `synth(n, m)` for every shape, and no `place_*` function in `cp_sat_placer.py` consumes a Beneš graph. Per `rfp-cp-sat-placement.md`'s decision log, "(2, 2) and (4, 4) Beneš deferred to phase 4."

Reasonable to keep — it's self-contained and the construction is non-trivial enough to be worth landing alongside the proof-by-tests — but a one-line `// no production caller yet — phase 4` at the top would save the next reader from hunting for the integration point.

docs/rfp-lane-safe-synth.md
```diff
@@ -0,0 +1,358 @@
...
+2. **Feed back into the original tree**: cascade.S1.out_1 →
+   merger.in_0 alongside the input belt. Then merger.in_0 has 2 arcs
+   (input + cascade leftover), reintroducing multi-arc. NOT clean.
+3. **Saturate the cascade**: make the cascade a `(K, K)` Beneš
+   sub-network where all K outputs feed downstream consumers. Doesn't
+   fit our use case (we want K → 1, not K → K).
+
+Option 1 is the right answer. Rate balance: the loop arc carries the
+same rate as the cascade input, so the cascade is in steady state.
+The verifier sees each arc with consistent rates.
```
storkme, line 183: This is the load-bearing claim of the whole RFP — that a self-loop at a 2-port splitter (one output feeding back to one of its own inputs) is verifier-stable. It's sketched in two sentences with no test. Kill criterion #1 catches the failure mode after the cascade construction is built; it'd be cheaper to verify upfront by hand-constructing the smallest cascade graph (e.g. a 1-splitter cascade for `(2, 1)`) and running it through `verify_balancer` *before* committing to phase 1.

Concretely: build a `BalancerGraph` with one splitter, two input arcs feeding port 0, an output arc on out-port 0, and a self-loop arc from out-port 1 back to in-port 1. Does `verify_balancer` stay full-rank, or does it hit `Singular`? If full-rank, the RFP rests on a checked claim instead of an asserted one. Five minutes of work; saves us from discovering the issue mid phase-3 of a multi-day plan.

docs/rfp-lane-safe-synth.md
```diff
@@ -0,0 +1,358 @@
...
+   baseline shipped in the lane-aware-routing PR.
+
+## Phasing
+
+1. **Phase 1: cascade construction in synth.** Add
+   `synth_lane_safe(n, m)` and `enrich_lane_safe`. Verify on the
+   existing fluid verifier — every shape in `1..=10` produces a
+   graph with no multi-arc ports and the same per-output rate as
+   the unenriched synth. ~2 days. No placer changes.
+
```
storkme, line 305: Phase 2 here ("drop the sideload table, per-lane vars, per-lane caps") should land *with* phase 1, not after. The longer the obsolete lane-aware infrastructure sits, the more downstream code accretes around shapes that consume it (rate-aware encoding, dual-lane inputs).

Suggested re-phasing: phase 1 = cascade construction in synth + simplification of `_route_belts` (revert sideload + per-lane caps in the same change). What's currently phase 2 collapses into "place coprime shapes." That keeps the placer in one consistent shape during the cutover instead of a hybrid where two routing models try to coexist.

storkme (commented):
Phase 1 round-trip plumbing looks solid — the synth → CpSat → topology_of_template → from_splitter_graph → verify_balancer chain is the right shape, and `into_owned_template`'s exhaustive-match-then-error on names is exactly the pattern this conversion needs. A few real things to flag, mostly around the next-steps section and the deferred infrastructure that ships with this PR.

**On the next-steps RFPs (`docs/rfp-lane-safe-synth.md` is the load-bearing one):** The two RFPs in this PR are deliberately contradictory — `rfp-lane-aware-routing.md`'s final decision-log entry concedes that `rfp-lane-safe-synth.md` is the path forward and that "phase 2 of that RFP rolls back the sideload table and per-lane caps as obsolete." Yet this PR ships ~700 lines of lane-aware machinery (sideload table, per-lane vars, dual-lane inputs, rate-aware encoding), none of which is exercised by any shape currently in the dispatch table. The PR description says "the rate-aware encoding stays useful even after lane-safe synth lands" — that's plausible, but the sideload table and dual-lane input modeling specifically have no consumer in this PR and are explicitly slated for removal. If lane-safe-synth is the picked path, the cleaner sequence is to either (a) commit lane-safe-synth phase 1 + remove the obsolete lane-aware bits in the same PR sequence, or (b) hold the lane-aware infrastructure on a branch until lane-safe-synth is validated. Shipping both and planning to delete one later is the "exploration overrunning its evidence" shape `CLAUDE.md` calls out as the dominant rework mode.

**Self-loop terminator construction is unverified.** `rfp-lane-safe-synth.md` worked-example for `(1, 5)` picks "Option 1: cascade.S1.out_1 feeds cascade.S1.in_0 (its own input)" with the justification "Rate balance verified by the all-fluid model — both arcs at the same port carry equivalent rates." This is the load-bearing claim of the whole RFP, and it's sketched without a check. Self-loops at a 2-port splitter create a degenerate cycle in `verify_balancer`'s linear system; whether it stays full-rank with the loop arc depends on how the conservation equations get parameterized. This is exactly what kill criterion #1 (cascade self-loop verifier disagreement) is for, but it's also the kind of thing that's better tested by hand on a 1-splitter cascade (e.g. `(2, 1)` with a synthetic self-loop) *before* committing to the RFP — five minutes of work to either confirm the math or kill the RFP. Right now phase 1 of lane-safe-synth would be the first place this assumption is exercised.

**Phasing.** The lane-safe-synth phase ordering is right (cascade in synth → simplify placer → coprime → bench), but phase 2's "drop sideload table, per-lane vars, per-lane caps" should land *with* phase 1, not after. The longer the obsolete infrastructure sits, the more downstream code accretes around it.

**Couple of structural notes on the diff itself:**
- The PR description claims this folds in `crates/balancer-gen`'s spike binary — but `crates/balancer-gen/scripts/place.py` (the more-general CP-SAT placer with UG / multi-direction splitter / edge-aware routing) is untouched. Now there are two parallel CP-SAT placement scripts on disk, with no docstring or README explaining which is canonical. Either delete the spike or note explicitly which is which.
- The dispatch table in `cp_sat_placer.py` covers 10 shapes (`(1,16)` and `(2,16)` included), and the round-trip suite tests all 10, but the PR description says "6 shapes covered." Worth syncing.
- `synth_benes` and `_input_routes_for_root`'s `n=3` branch are both dead code (no caller in this PR; the description acknowledges (n,n) Beneš and (3,m) shapes are deferred). Reasonable to keep `synth_benes` since it's a self-contained pure construction with tests, but it should at least be documented as "phase 4 prep, no consumer yet" so the next reader doesn't assume it's wired up.

storkme:
Heads-up for sequencing: phase 4 (compose_parallel/compose_series) and phase 3.4 (bake_missing_shapes loop) work is currently anchored on `crates/balancer-gen/scripts/place.py` — the spike placer. Since this PR introduces `scripts/cp_sat_placer.py` as the canonical Rust-side placer, plan is: land this PR first, then migrate the bake pipeline onto `cp_sat_placer.py` in the same workstream as phase 3.4 cleanup. Until that migration, the spike script stays. Calling it out so reviewers know the duplication is temporary, not architectural.

PR #276, opened 2026-05-02
Phase 4 increments 1+2: per-arc rate plumbing and lane-balancer helper

storkme:
## Intent

Build out the CP-SAT placer to cover the first wave of `(1, m)` coprime shapes — `(1, 5)`, `(1, 6)`, `(1, 7)` — and lay scaffolding for the harder 4-level shapes `(1, 9)` / `(1, 10)`. Started as Phase 4 increments 1+2 from `rfp-lane-aware-routing.md` (rate plumbing + lane-balancer helper) but grew to include the full UG-belt extension and a synth-driven generalised placer once `(1, 5)` proved it needed both.

## Scope

**In:**
- **Wire format**: `CpSatRequest` gains optional `arc_throughputs: Vec<f64>` from `verify_balancer`. Backward compatible.
- **UG-belt support inside `_route_belts`** (~340 LOC). Three entity kinds per (tile, direction): `surf`, `ug_in`, `ug_out`. Pair vars couple entries to exits. Same-direction tunnel non-overlap respects Factorio's auto-pairing rule. Three-case flow conservation gated by `only_enforce_if(kind)`.
- **`_add_lane_balancer_south()`** helper for placement-only splitters that rebalance lanes after a perpendicular turn.
- **`place_one_to_five()`** — hand-tuned (1, 5) placer using lane balancer + UG.
- **`place_one_to_m_from_synth()`** — generalised synth-driven placer. Derives geometry from the BalancerGraph + arc throughputs:
  - 3-level branch covers `m ∈ {6, 7}`.
  - 4-level branch wires `m ∈ {9, 10}` with wide-spread L1→L2 staggered routing — model is built but doesn't solve in 15min/seed (see overnight bake).
- **Source-direction constraint fix in `_route_belts`**: forbid `d == OPPOSITE(splitter_dir)` at splitter sources. Belts heading back into the splitter face create self-loops in topology recovery; previously masked by short timeouts.
- **Round-trip tests**: `(1, 5)`, `(1, 6)`, `(1, 7)` active; `(1, 9)` added with `#[ignore]`. Default timeout bumped 60s → 180s with `round_trip_with_timeout(n, m, secs)` helper for per-shape tuning. Suite goes 10 → 12 active + 1 ignored.
- **Solver knobs**: `FUCKTORIO_CP_SAT_SEED` env override (used by bake harness), `FUCKTORIO_CP_SAT_UG_MAX` for layout debugging, optional `workers` param for parallel CP-SAT (4-level branch defaults to 8).
- **Overnight bake harness** (`scripts/bake_cp_sat_runner.py` + `bake_cp_sat_report.py` + `bake_overnight.sh`): async parallel seed-sweep with priority queue, journal w/ fsync, resume support. Adapted from PR #283's runner pattern.
- **Docs**: `rfp-lane-aware-routing.md` decision log entries through Phase 4 a.3, `handoff-cp-sat-ug.md` (next-session pickup), `bake-overnight-results.md` (32 seeds × 2 shapes = 64 probes, 0/64 solved → recommendation: layout work, not more compute).

**Out:**
- `(1, 9)`, `(1, 10)` — wired in dispatch but the 4-level CP-SAT model doesn't solve (overnight bake confirmed median wall = full timeout for every seed). Layout work needed; recommendation in the bake report is a compositional `(1, 10) = (1, 5) ⊕ (1, 2)` placer.
- `(1, 11)`, `(1, 12)` and the `(n>1, m)` family — separate plan.
- Phase 5 (bench + library regen) — not started.

## Verification

- `FUCKTORIO_RUN_CP_SAT=1 cargo test --manifest-path crates/core/Cargo.toml --test cp_sat_round_trip` — 12 active tests pass, 1 ignored.
- `cargo test --manifest-path crates/core/Cargo.toml` — full suite green.
- `cargo clippy --manifest-path crates/core/Cargo.toml --all-targets` — no new warnings.
- Wire-format serialization test updated to verify `arc_throughputs` round-trips.
- Dyadic shapes (`(1, 1)`–`(2, 16)`) regressions: `(1, 16)`/`(2, 16)` go from ~16s → ~60s under the new UG model; the optimiser correctly leaves UG vars at 0 for shapes that don't need it, so feasibility is preserved.
- Overnight bake: 64 probes (32 seeds × 2 shapes) at 5min each. 0/64 solved. Median wall = max wall = full timeout — not seed-sensitive.

## Deviations from intent

- **UG support brought in-scope.** Original RFP marked it Out for this PR. Once `(1, 5)` hit the layout wall it became the cleanest unblock path; deferred-decision in `rfp-lane-aware-routing.md`.
- **Phase 4 a.3 partially shipped.** `(1, 5)`/`(1, 6)`/`(1, 7)` complete; `(1, 9)`/`(1, 10)` blocked on layout work, not on placer code. Code path is wired in dispatch but no round-trip test until layout is tightened.

## Models / contracts touched

- `CpSatRequest` — added optional `arc_throughputs: Vec<f64>` (`skip_serializing_if`, backward compatible).
- `_route_belts` — model rewrite: three entity kinds, pair vars, source-dir constraint, three-case flow conservation. Return signature changed to `dict[tile -> (direction, kind)]`.
- `rfp-lane-aware-routing.md` — Phase 4 progress notes through a.3.
- `handoff-cp-sat-ug.md` — new pickup doc.

## Self-review feedback

Items 1, 2, 3, 4, 5 from the self-review on this PR have been addressed:

1. ✅ **PR description rewrite** — this body, reflecting actual scope.
2. ✅ **`(1, 5)` flake-under-load comment** — empirically tested (8/8 full-suite runs at `--test-threads=4` green); the d=OPPOSITE source-direction constraint added in `8d83725` makes every feasible solution topology-correct, so the route-emission ordering is just a search heuristic, not a correctness fix. Subsumed `(1, 5)` into the generalised placer; deleted hand-tuned `place_one_to_five` (~250 LOC); softened the comment to not over-claim.
3. ✅ **`verify_balancer` error swallow** — now propagates as `PlacementError::Engine` instead of silent fallback to discrete unit rates.
4. ✅ **Per-shape test timeouts** — dyadic shapes back to 60 s default; `(1, 16)` / `(2, 16)` at 120 s; `(1, 5)` at 300 s. Worst-case sequential CI ~33 min → ~18 min.
5. ✅ **Ignored `(1, 9)` test removed** — overnight bake confirmed it doesn't solve under current layout. Tracked under issue #136.

Deferred follow-ups (non-blocking):
- `_find_arc_rate` is O(arcs) per call — could be a dict for clarity. Marginal.
- Dyadic-regression timing data not captured here.


storkme (commented):
## Overview

The PR description undersells the actual scope. It claims to ship Phase 4 increments 1+2 (rate plumbing + the `_add_lane_balancer_south` helper) and explicitly defers `(1, 5)` and excludes UG-belt support. The diff actually ships:

1. ✅ Per-arc rate plumbing through the `CpSatRequest` wire format (~18 LOC Rust)
2. ✅ `_add_lane_balancer_south` helper (~35 LOC Python)
3. ❗ Full UG-belt support inside `_route_belts` (~340 LOC Python — model rewrite with three entity kinds, pair vars, same-direction non-overlap, three-case flow conservation)
4. ❗ Hand-tuned `place_one_to_five` (~280 LOC), generalised `place_one_to_m_from_synth` (~400 LOC)
5. ❗ New round-trip tests for `(1, 5) (1, 6) (1, 7)`, plus `#[ignore]`'d `(1, 9)`
6. ❗ Default test timeout raised 60s → 180s

**Body needs updating before merge.** A reviewer reading the description will be looking for one thing and find a much bigger PR. (The commit log is honest about phases 3.2A–3.2D, 3.3, 4 — the PR body needs to match.)

## Code quality

### `_route_belts` UG model — well-engineered

The CP-SAT extension is principled. Specifically strong:

- `scripts/cp_sat_placer.py:248–266` — entity-kind decomposition: `belt[(t,d)] == surf + ug_in + ug_out` with `sum(belt) ≤ 1` cleanly enforces tile uniqueness across kinds and directions.
- `scripts/cp_sat_placer.py:401–419` — same-direction tunnel non-overlap is explicit and correct: enumerate tiles each pair *covers* (k+1 inclusive) and bound `sum(pairs covering this tile) ≤ 1`. Matches Factorio's auto-pairing semantics.
- `scripts/cp_sat_placer.py:540–583` — three-case flow conservation, gated by `only_enforce_if(kind)`, is the right encoding for "surface and UG-entry accept surface inflow; UG-exit only receives the teleport." Avoids messy big-M.
- `scripts/cp_sat_placer.py:587–605` — outflow surface successor check correctly disables itself with `ug_in_r.Not()` so the route doesn't need a surface neighbor when descending.
- `scripts/cp_sat_placer.py:355–378` — the `pairs_to_t` walk (entry at `t - k·δ` exits at `t`) is correct; the indexing matches the tunnel direction convention.
- `scripts/cp_sat_placer.py:496–500` — forbidding UG roles at source/sink is necessary and explicit.
- Optimizer minimises total entity count, so non-UG-needing shapes recover surface-only — good guarantee that adding UG support doesn't degrade dyadic placements.

### `_pick_rate_scale` — nice touch

`scripts/cp_sat_placer.py:1417` uses `Fraction.limit_denominator(1000)` to recover exact denominators from the verifier's floats, then LCMs them. This is the right way to handle the `RATE_SCALE = 10` comment's overpromise (`1..=10 × 1..=10` includes `(3, 7)` whose 1/7 doesn't divide 10). The module-level constant is a default; the synth-driven placer auto-scales correctly.

### Comments are dense but useful

The docstrings on `_route_belts`, `place_one_to_five`, and the `_SYNTH_CTX` block are long but well-targeted — they explain *why* (Factorio physics, layout-design rationale) rather than *what*. Aligns with project convention.

## Specific issues

### 1. Stale PR description (above) — biggest reviewer-facing problem

### 2. Best-effort `verify_balancer` swallows errors silently

`crates/core/src/balancer/placement/cp_sat.rs:117-119`:
```rust
let arc_throughputs = crate::balancer::verify::verify_balancer(req.graph)
    .ok()
    .map(|outcome| outcome.arc_throughputs);
```
Comment says "shouldn't happen — synth produces verified graphs." If it does, the placer silently uses discrete unit rates, which produces *wrong* placements for non-dyadic shapes (rate-aware caps don't apply). At minimum log the error; better, propagate as `PlacementError` for non-dyadic requests.

### 3. The `(1, 5)` dispatch comment hints at a real flake

`scripts/cp_sat_placer.py:1893–1898`:
> the generalised one subsumes it in isolation but flakes intermittently in full-suite runs (CP-SAT picks OutputDegree-violating alternates when system load extends solve time)

That's a load-dependent correctness issue with the generalized placer, not just a perf thing. "OutputDegree-violating alternates" suggests a missing structural constraint that's only being filtered post-hoc by lucky timeouts. Worth either (a) adding the missing structural constraint that's being relied on implicitly, or (b) opening a follow-up issue and crisply stating the symptom in the dispatch table comment so the reader knows *where the bug lives*, not just that it exists.

### 4. Default timeout 60s → 180s

`crates/core/tests/cp_sat_round_trip.rs:51` and `:91`. With 11+ round-trip tests at 180s each, worst-case CI/local cycle is ~33 min. Two consequences:
- A regression on a previously-fast shape now takes 3× longer to surface.
- Local dev loop pain.

Suggest keeping default at 60s for shapes known to solve fast (`(1,1)`–`(2, 8)`) and only bumping for the slow ones (`(1, 16)`, `(2, 16)`, `(1, 5)`, …) with `round_trip_with_timeout(...)`. The harness already supports it.

### 5. `#[ignore]`'d `(1, 9)` test risks bitrot

`crates/core/tests/cp_sat_round_trip.rs:159–164`. Ignored tests are easy to forget; this one's documented well in the comment but still has no automatic prod. Either delete it for now (track in an issue) or add a periodic CI job that runs `--ignored` weekly.

### 6. Model size at scale — no profiling data

The model gains `(t × d × n_routes × n_lanes)` for `ug_in_r` + `ug_out_r` and roughly `(t × 4 × 3 × n_routes × n_lanes)` for `pair`. The PR notes `(1, 16) / (2, 16)` regress 16s → 60s. No data on how the simpler dyadic shapes regressed (the description says "stay green" but doesn't quantify time deltas). Worth adding a one-line benchmark table to the PR body.

### 7. `_find_arc_rate` is O(arcs) per call

`scripts/cp_sat_placer.py:651`. Fine in absolute terms (small graphs) but could be a dict from `(canonical(src), canonical(dst))` for clarity and to avoid the linear scan in inner loops. Minor.

## Performance

- UG search space increase is the main cost; the optimizer correctly leaves UG vars at 0 for dyadic shapes, so feasibility is preserved but solve-time grows.
- Test default of 180s is a CI signal-loss cost as noted.
- `verify_balancer` runs on every placement request; for dyadic shapes this is wasted work since they don't use `arc_throughputs`. Could skip on `n × m` being a power-of-two pair, but small enough to ignore.

## Test coverage

- 4 new round-trip tests (`(1, 5) (1, 6) (1, 7) (1, 9)`); good coverage for the new placers.
- Wire-format serialisation test updated for `arc_throughputs`.
- No direct test of the `(1, 5)` flake-under-load described in the dispatch comment — if that's a real issue it should have a regression test gating it.
- No test of `_route_belts` UG support in isolation. The integration tests are exhaustive but a focused unit test would help debug regressions.

## Security / risk

- Module-level state (`_SYNTH_CTX`, `_SOLVER_PARAMS`) is fine for the per-process subprocess model used here. If the placer ever becomes a long-lived service, this needs to be per-request.
- Wire-format addition is backward-compatible (`Option<Vec<f64>>`, `skip_serializing_if`). Older Rust + newer Python and vice versa both work.
- No external input changes; CP-SAT solver parameters still pinned.

## Cross-PR context with #283

This PR and #283 (lane-correctness validator) compose: #276 places templates that satisfy lane caps at *placement* time, #283 audits the resulting library at *runtime* and proves the placements are correct under steady-state Factorio physics. The same lane-physics model (head-on splitters, 50/50 lane distribution, sideload rules) shows up in both — #276 encodes it as constraints, #283 simulates it as a transfer function. After both land, the project has a coherent lane-physics model end-to-end (placement constraint ↔ runtime audit). Strong position to build coprime-shape coverage from.

## Recommendation

**Update the PR body** to reflect actual scope (UG support + `(1, 5..7)` placers + ignored `(1, 9)`), then this is mergeable. The constraint model is sound and the design rationale is well-documented. The two items I'd push on before merge:

1. PR description rewrite (5 min).
2. Either fix or sharply scope the `(1, 5)` flake-under-load comment in `main()` — that comment is admitting a known correctness bug that depends on system load, which is a real risk to land silently.

Items 3–7 above are good follow-ups, not blockers.

PR #272, opened 2026-05-01
Phase 3.2A.1/B/C: flow-conservation belt routing with UGs and direction freedom

storkme:
## Intent

Phase 3.2A.1 + 3.2B + 3.2C of [`docs/rfp-balancer-place-routing.md`](../blob/main/docs/rfp-balancer-place-routing.md): extend the spike from "place splitters with no-overlap" to a full **flow-conservation belt routing encoding** in CP-SAT, with **underground-belt support** and **per-splitter direction freedom**. Library round-trip (template → topology → CP-SAT routing → entity assembly → classify) closes for simple shapes; coprime back-loop shapes are diagnosed and deferred to phase 3.2A.2.

Continues the throughput-priority work from #269. Branch was reset to `main` after #269's squash merge; these are the new commits.

## Scope

- **In:**
  - **3.2A.1** — flow-conservation routing. `arc[(c, d, e)]` bool variables; conservation per (cell, edge); no-overlap on non-splitter cells; source/dest cells determined per edge from caller-provided splitter positions and slot assignments.
  - **3.2B** — UG belt support. `ug_arc[(c, d, L, e)]` virtual arcs (input cell → output cell); pairing constraint forbidding interfering UG outputs in the same input-output span on the same axis; entity emission for input/output halves.
  - **3.2C** — per-splitter direction freedom. Splitter direction caller-provided (extracted from library entities); splitter tile and slot positions computed from direction; outflow at splitter sources constrained to facing direction, inflow at splitter dests constrained to come from the back.
  - Round-trip tests for 8 library shapes; 3 pass cleanly, 5 fail with diagnosed root cause.
  - RFP decision log updated with status table and pinned next-session work.

- **Out (deliberately deferred):**
  - **3.2A.2** — make slot assignment a CP-SAT variable, not a Rust greedy. This is the single root cause of the 5 failing shapes — the greedy diverges from library slot patterns on back-loop shapes. Estimated 200-300 LOC of Python; unblocks all remaining sub-phases.
  - **3.2D** — bounding-box minimisation (gated on full shape coverage).
  - **3.3** — measurement against Factorio-SAT baseline (gated on 3.2A.2).
  - **3.4** — bake into `balancer_library_extra.rs` (gated on 3.2A.2 — only worth doing once we have shapes the library lacks).

## Verification

- [x] `cargo test --manifest-path crates/core/Cargo.toml --lib` — 489 tests pass (no new tests; phase-3 work is in the binary crate, gated by `cargo run`).
[...]

PR #283, opened 2026-05-02
belt_flow: lane-correctness validator improvements (cycle-breaker fix + iterative walker)

storkme:
## Intent

Lane-correctness validator (`belt_flow.rs`) iterative-walker fix is the **headline contribution** — drops the library lane audit from **344 errors / 29 templates → 14 errors / 1 template** (the remaining 14 are downstream of a real `belt sideloads into UG input` warning on the python-derived `(7, 2)`; re-bake tracked separately as #284).

The branch also carries a substantial chunk of earlier balancer-gen work that hadn't merged yet — Mode C/D CP-SAT routing, compose_series/compose_parallel, Phase 4 composition combinators, lane gate in `bake_missing_shapes`, RFPs and decision logs. ~38 commits, ~8000 LOC. The lane-validator fix is the part that's substantively new in this PR; the rest is bundled because it's all on the same branch and thematically aligned with the throughput-priority workstream.

Continues the work that #272 (now closed) was carrying — split off the lane-validator scope so it can be reviewed.

### Lane validator (the headline)

1. **Cycle-breaker fix** (`f6a9aa8`, `belt_flow.rs:2304-2315`) — replace "propagate non-zero half to both" with "always-average both halves". The old rule produced 2× the correct rate at stuck splitter secondaries (in-degree unification). Drops **252 of 344 audit errors**.
2. **Iterative-to-convergence Jacobi pass** (`2fae382`, `belt_flow.rs:2335-2440`) — fixed-point iteration after the topo-sort handles balancer feedback loops the single-pass walker can't resolve. ≤200 iters with `EPS=1e-5`; cycle gain damps by 0.5/pass. Drops **92 → 18**.
3. **Splitter lane-mixing** (`60d2dc5`, `belt_flow.rs:2387-2400`) — real Factorio splitters mix lanes; input `[L=15, R=0]` becomes output `[L=7.5, R=7.5]` per half. Phase 2 of the iterative pass distributes the pair's total feeder contribution across all 4 output lanes uniformly. Drops **18 → 14**.
4. **`compute_lane_rates_impl` UG-output `+= seed`** (caught during dev) — surface inserter injection on a UG-output tile must accumulate alongside the inherited behind-rate, not replace it.

### Reproducibility / tooling

5. **CP-SAT `random_seed` pinned** (`f6a9aa8`, `place.py:67`) — `DEFAULT_SEED = 42` (overridable via `FUCKTORIO_CP_SAT_SEED` env var or per-request `random_seed` field). Eliminates 35× run-to-run variance on hard Clos compose problems.
6. **Seed sweep harness** (`f25cd55`, `scripts/sweep_balancer_seeds.sh`) — sweep CP-SAT seeds on (4, 9) Clos compose, sorted summary. Currently running.
7. **`bake_missing_shapes` rebake plumbing** (`a85a116`) — `FUCKTORIO_REBAKE_SHAPES='(m,n);...'` overrides the "already in library" skip. Lets us regenerate grandfathered-in templates through the current lane gate.
8. **Three focused unit tests** (`fd0c48f`, `belt_flow.rs:tests`) — cycle-breaker math, (3, 3) iterative-walker convergence, UG-output injection. Per review feedback, separate from the audit count.

### CI / config

9. **Cherry-pick `8eb6ace`** (`778d572`) — `#[ignore]` the `stress_advanced_circuit_partitioned_7s_from_plates` test that was already too slow on main; CI was already bypassing it there.

### Bundled balancer-gen work (already on the branch when split from #272)

The following commits were already on `claude/throughput-priority-merges-YcXbi` from the throughput-priority workstream — they're landing here for the first time:

- **Mode C / Mode D CP-SAT routing** in `place.py` — `solve_pure_routing`, `solve_synth_place`, `solve_synth_place_dirs`. Spatial pruning for circuit encoding. ~1,800 LOC.
- **Phase 4 composition combinators** in `balancer-gen/src/main.rs` — `compose_parallel`, `compose_series`, `clos_interleave` integration. (4, 9) Clos compose works end-to-end via decomposition.
- **Phase 2 lane gate in `bake_missing_shapes`** — retries on `underground-belt Severity::Error`, bumps min_jh up to 3 attempts. (Inherited from `4ba6439`.)
- **L=1 UG fix** + **6 L=1-broken templates re-baked** via the circuit encoding (`1a9a5ac`, `92990e3`, `0dad418`).
- **RFP / decision-log docs** for jh-search, place-routing, spatial-pruning, bake-lane-validation.

## Deferred

- **`(7, 2)` re-bake** (#284) — recipe + plumbing landed; the actual library swap is held back because the re-baked `7×11` template (vs old `7×7`) triggered apparent timeouts in two `tier5`/stress tests on local. CPU contention from orphaned `place.py` processes contaminated the measurement; needs a clean re-run before deciding (a) shrink the template, (b) bump the timeouts, (c) leave grandfathered.
- **`DEFAULT_SEED = 42` is a placeholder** — sweep harness exists (`scripts/sweep_balancer_seeds.sh`) and is running; the empirical winner will be folded in as a one-line follow-up. Current state is "we made it reproducible but not fast."
- **Phase 3.4** (bake compositions into library, issue #136) — gated on compute availability.

## Audit progression

| Stage | Errors | Templates affected |
|-------|-------:|-------------------:|
| Before any work | 344 | 29 |
| After cycle-breaker fix (`f6a9aa8`) | 92 | 29 |
| After iterative walker (`2fae382`) | 18 | 2 |
| After splitter lane-mixing (`60d2dc5`) | 14 | 1 (`(7, 2)`) |

## Verification

- [x] `cargo test --manifest-path crates/core/Cargo.toml` — **582 tests pass**, 0 failures (29 ignored, includes `stress_advanced_circuit_partitioned_7s_from_plates` from `8eb6ace`).
- [x] `cargo clippy --lib` clean.
- [x] Three focused unit tests for the math (cycle-breaker, (3, 3) convergence, UG-out injection).
- [x] Library lane audit at branch head: 14 errors / 1 template / 75 total.

## Cross-PR context

This PR pairs with #276 (CP-SAT placement engine, rate-aware lane caps). The two are working the same problem from opposite ends: #276 encodes lane physics as CP-SAT constraints at *placement time*; #283 simulates lane physics as a transfer function at *runtime audit time*. Together they make "the bake produces a lane-correct library" a falsifiable claim.

## Notes on review feedback

- ✅ Updated body to enumerate full scope.
- ✅ Added 3 focused unit tests for the math (`fd0c48f`).
- ✅ Filed #284 for the `(7, 2)` re-bake follow-up.
- 🔁 Seed value: 42 is a placeholder; sweep is running.
- 🔁 `feeder_contribution` / `do_propagate` lane-logic dedup — non-blocker, could be a follow-up if anyone touches that path.
- 🔁 Audit test gating (`BALANCER_AUDIT_FAIL` should become default once 14 baseline is firm) — follow-up, gates on (7, 2) re-bake landing.


storkme (commented):
## Cross-PR context with #276

These two PRs are working the same problem from opposite ends of the SAT-driven balancer pipeline:

| | **#276** (CP-SAT placement) | **#283** (lane-correctness validator) |
|---|---|---|
| Layer | Generates balancer geometry | Audits placed/baked geometry |
| What it adds | Rate-aware lane caps in CP-SAT (multiple low-rate routes per lane) | Iterative fixed-point lane-rate walker that models splitter feedback + lane mixing |
| Key insight | Per-lane cap is `0.5` (head-on input fills both lanes); coprime shapes need rate-aware caps | Real splitters mix lanes (`[L=15, R=0] → [L=7.5, R=7.5]`); naive Kahn topo-sort settles on first-touch and misses feedback equilibria |
| Wire format | Adds `arc_throughputs: Vec<f64>` to placement request | n/a (Rust-only) |
| CP-SAT touch point | Builds the SAT model | Pins solver `random_seed=42` for reproducibility |

They compose: #276 places templates that satisfy lane caps at *placement* time, #283 audits the resulting library at *runtime* and proves the placements are correct under steady-state Factorio physics. Together they make "the bake produces a lane-correct library" a falsifiable claim. The same lane-physics model (head-on splitters, 50/50 lane distribution, sideload rules) shows up in both — #276 encodes it as constraints, #283 simulates it as a transfer function.

The audit drop from **344 → 14 errors** is the headline number, and it's plausibly that big because the previous validator was systematically wrong about feedback loops, not because the templates were wrong about routing.

## Overview

The PR does six things, ordered by leverage:

1. **Cycle-breaker fix** (`belt_flow.rs:2304–2315`) — replaces "propagate non-zero half to both" with "average both halves." Drops 252 of 344 audit errors (73% of the surface area).
2. **Iterative-to-convergence Jacobi pass** (`belt_flow.rs:2335–2440`) — splitter pairs settle to fixed point over ≤200 iterations with `EPS=1e-5`. Drops further 92 → 18.
3. **Splitter lane-mixing in phase 2 of the iterative pass** (`belt_flow.rs:2387–2400`) — sums all 4 input lanes, distributes uniformly across all 4 output lanes. Drops 18 → 14.
4. **CP-SAT seed pin** + env override (`balancer-gen/scripts/place.py`) — eliminates 35× run-to-run variance on `(4, 9)` Clos compose.
5. **`template_validate` module + audit test** (new files) — phase 1 of `rfp-balancer-bake-lane-validation.md`.
6. **`bake_missing_shapes` rebake plumbing** (`FUCKTORIO_REBAKE_SHAPES` env var).

Plus ~1,800 LOC in `balancer-gen/src/main.rs` for Mode C / compose_clos / Phase 4 composition infrastructure that the PR description doesn't mention.

## Code quality

### Cycle-breaker fix — correct and small

`belt_flow.rs:2304–2315`: the new "always-average" rule is the right physics. A 1→2 splitter with input `[5, 0]` and sibling `[0, 0]` *does* produce `[2.5, 0]` on each output — the old "propagate to both" rule was producing `[5, 0]` per half (2× the truth). The diff comment explains the bug clearly:

> The old "propagate non-zero to both" rule was producing 2× the correct rate at stuck secondaries — 252 of the 344 audit errors.

That's a direct correctness fix, no new state, and the audit drop is the test.

### Iterative pass — well-engineered Jacobi

`belt_flow.rs:2335–2440`. Three phases per iteration (non-splitter / splitter pairs / UG outputs), `EPS=1e-5`, `MAX_ITER=200`. The damping argument is solid: splitters halve cycle gain per pass, so feedback error decays geometrically; 14 iterations is plenty for a 15/s seed to drop below `1e-3`. 200 is a comfortable cap.

The `seed_rates: FxHashMap = lane_rates.clone()` snapshot before the topo-sort is the key insight — the iterative pass needs the immutable injection base because feeders accumulate on top of seed values, and a naive recompute would lose them. The comment at `belt_flow.rs:2092–2097` is good.

### `feeder_contribution` — duplication risk

`belt_flow.rs:2466–2526` reimplements the four-case lane logic from `do_propagate` (`belt_flow.rs:2528+`) but in pull direction. The cases are correctly mirrored (same-dir / behind / sideload / 90° turn) but **two functions encoding the same Factorio belt-mixing rule will drift**. Future bug: someone fixes a sideload edge case in `do_propagate` and forgets `feeder_contribution`, and the validator silently disagrees with itself.

Suggest factoring the four-case decision into a single helper that both `do_propagate` (push) and `feeder_contribution` (pull) call, with the function body just doing the lane-vector mapping. Not blocking, but I'd want it tracked.

### Splitter lane-mixing — physically correct

`belt_flow.rs:2387–2400`: `total = a_fc + b_fc`; `per_lane = total / 4`; both halves get `[per_lane, per_lane]`. That's what real Factorio splitters do (the input lane discipline is destroyed by the splitter and rebuilt on output). Distinct from the cycle-breaker fix but conceptually adjacent.

The two together explain why the 92 → 18 → 14 deltas land where they do: cycle-breaker fixes the symmetric-feedback case, lane-mixing fixes the asymmetric-input case. Both rules are simultaneously necessary.

### UG-output `+=` not `REPLACE` — caught by regression, but no targeted test

`belt_flow.rs:2402–2425`. The fix is right (the comment is good), but the PR body notes this was "caught by tier4 regression, fixed pre-merge." There's no focused unit test in this PR for the inserter-injecting-on-UG-out case. The next time someone touches this code path, only the e2e runtime catches the regression — which is slow feedback. Worth a small unit test.

### `template_validate` synthesis — defensive

`bus/template_validate.rs:121–127`: only `input_tiles` get tagged `carries: TEST_ITEM`. The comment explains why (avoiding double-injection at empty splitter halves). This is the kind of defensive comment that's worth its weight — without it the next reader would absolutely break this in a "tidy-up." Keep it.

### Audit test prints, doesn't gate

`tests/balancer_lane_audit.rs:148–155`: the test always passes; gating is via `BALANCER_AUDIT_FAIL=1` env. Reasonable for phase 1 (discovery), but it means a regression that adds 5 errors back to the library *will not break CI*. Once the 14 baseline is firm, make `BALANCER_AUDIT_FAIL` the default and require explicit override to suppress.

## Specific issues / risks

### 1. PR body undersells scope

PR body claims this is "the lane-validator scope split out for review." Diff includes ~1,800 LOC of `balancer-gen` Phase 3.2A/B/C/D/3.3/4 composition infrastructure (Mode C routing, `compose_clos`, `bake_compose_1_9`, lane-balancer-south helpers). Update the body to enumerate what's actually being landed; the commit log is honest and the body should match.

### 2. `DEFAULT_SEED = 42` is a placeholder

`scripts/place.py`: comment says "placeholder; updated after empirical sweep." The sweep harness (`scripts/sweep_balancer_seeds.sh`) is included but the sweep hasn't been run (or the result hasn't been folded into the constant). Two options: run the sweep and pin the empirically-fast value before merge, or land `42` and treat the value swap as a follow-up — but call this out explicitly in the PR body so the next reviewer doesn't think they're getting an optimised seed.

The current state is "we made it reproducible but not fast." That's progress, but worth being explicit.

### 3. `(7, 2)` re-bake deferred — known timeout regression

PR body:

> The new `(7, 2)` is `7×11` (vs old `7×7`) which expands downstream layouts and triggered timeouts in two `tier5`/stress tests on local; needs investigation.

Recipe + plumbing are landing; the re-bake itself is held back. Reasonable, but the deferred work is **a known-failing case being staged**. Open a tracking issue with the two timeout sites pinned, or this gets forgotten.

### 4. No targeted unit test for the cycle-breaker math

The fix is validated only by the audit count drop (344 → 92). A unit test of the form "given a 1→2 splitter with feeder rates `[5, 0]` and `[0, 0]`, expect outputs `[2.5, 0]` per half" would be:
- A regression guard against future changes (someone "tidies up" the averaging math).
- A documented contract for what the function is *supposed* to compute, which is currently only described in comments.

5–10 lines, high value.

### 5. Iterative-pass perf at scale

`MAX_ITER=200` × number of splitter pairs in the largest templates × per-iteration cost. The audit walks every template at startup; with 29 templates and the largest containing dozens of splitters, the walker fires ~29 × 200 × O(splitters) Jacobi ops. Probably fine in absolute terms (small graphs), but `compute_lane_rates` is also called by every e2e and bus-layout test, so this lands on the hot path. A perf note in the PR (before/after timing on a tier4 e2e) would close that loop.

### 6. `feeder_contribution` / `do_propagate` duplication (above)

Long-term maintenance hazard. File a follow-up issue.

### 7. `tier4` regression: how was it caught and how do we keep catching it?

PR body: "Bonus catch during dev: UG-output rate accumulation needed `+=` not `REPLACE` to preserve inserter injections on the UG-out's surface — caught by tier4 regression." Which tier4 test? Was a regression-specific assertion added, or did a generic e2e fail and the fix made it pass again? If the latter, the test will silently stop catching this if the e2e topology changes. Worth pinning with a focused test as in (4).

## Performance

- Audit overhead: probably small but not measured in the PR.
- Iterative pass: bounded by `MAX_ITER=200`, geometric decay, fine.
- CP-SAT seed pin removes a 35× variance regime — wins more than it costs.

## Test coverage

- `template_validate.rs` ships with three focused tests: `(4,4) Benes`, `(1,2)`, and an L=1 UG-pair rejection (the lane gate).
- `balancer_lane_audit.rs` is a discovery test, not a gate (per above).
- No targeted tests for the cycle-breaker math, lane-mixing math, or UG-output `+=` regression.

Suggest minimum: 3 unit tests for the math (cycle-breaker, lane-mix, UG-out injection). Total ~30 LOC.

## Security / risk

- Env-var-driven config (`FUCKTORIO_CP_SAT_SEED`, `FUCKTORIO_REBAKE_SHAPES`, `BALANCER_AUDIT_FAIL`, `FUCKTORIO_DEBUG_2_2`, …) — all dev-time, no production exposure.
- No external input parsing changes.
- No new dependencies.

## Recommendation

The lane-validator fix is **substantively correct** — the math is right, the audit drop is real, and the iterative pass is the right structural addition. Cross-cuts well with #276's lane-cap encoding (both PRs are now treating splitter physics consistently in two different layers).

Before merge:

1. **Update PR body** to enumerate actual scope (lane fix + balancer-gen Mode C + compose infra + rebake plumbing).
2. **Add 3 focused unit tests** for the math (cycle-breaker, lane-mix, UG-out injection). ~30 LOC. Closes the "next contributor breaks this" gap.
3. **Decide on the seed**: either run the sweep and pin the winner, or commit to "42 is placeholder, swap is follow-up" in the body.
4. **File the `(7, 2)` re-bake follow-up issue** with the two timeout sites named.

Items 5–7 (`feeder_contribution` dedup, audit-test gating, perf timing) are good follow-ups, not blockers.

Notable cross-PR positive: between this PR and #276, the project now has a coherent lane-physics model end-to-end (placement constraint ↔ runtime audit). That's a genuinely strong position to build coprime-shape coverage from.

storkme:
Thanks for the thorough review — the four blockers are addressed:

1. **PR body updated** to enumerate actual scope (lane validator + balancer-gen Mode C/D + compose combinators + rebake plumbing + cherry-pick). The "1,800 LOC of compose infra not mentioned" was real — that's all earlier-on-this-branch work from the throughput-priority workstream that's also landing.

2. **Three focused unit tests added** in `fd0c48f`:
   - `splitter_one_feeder_outputs_balanced_halves` — 1→2 split with full-belt input + empty sibling → both halves [3.75, 3.75]/lane (lane-mixed average).
   - `iterative_walker_balances_3_3_template` — every internal tile of (3, 3) settles at exactly [7.5, 7.5]/lane post-iteration. The headline regression case.
   - `ug_output_preserves_inserter_injection` — UG-out rate = inherited behind-rate + inserter injection on the surface (not REPLACE). The bug caught by tier4 e2e during dev now has a unit-level guard.

3. **Seed = 42 placeholder** explicitly called out in the deferred section. `scripts/sweep_balancer_seeds.sh` is running now (10 seeds, ~2-3h); winner folds in as a one-line follow-up to `place.py:67`.

4. **Issue #284** filed for `(7, 2)` re-bake, with the two timeout sites pinned and reproduction notes.

Non-blocker items (`feeder_contribution`/`do_propagate` dedup, audit-test gating, perf timing) noted in the PR body as follow-ups.

PR #306, opened 2026-07-11
RFP: Fulgora scrap-based layouts

storkme:
## Intent

Design doc (RFP) for scrap-based Fulgora production chains — the follow-up the 2026-07-11 solver spike scoped. No production code in this PR; the RFP itself is the deliverable, and the PR exists for design review.

## Scope

- **Voiding policy**: layout-level voider rows synthesized from honest solver surplus, sidestepping the eps_surplus laundering family the spike discovered (the LP is never asked to *want* destruction).
- **Solid surplus export**: step-7 merger extension (independently valuable — un-strands uranium chains).
- **Recycler rows + item sorting**: the centerpiece; mixed recycler output vs the one-item-per-lane bus model. Architecture decision (filter-inserter sushi vs filtered-splitter cascade vs pre-solved block) explicitly gated on a Phase 0 physicals verification.
- Kill criteria per house template, including a sorter footprint bound benchmarked against the community blueprint corpus.

## Verification run

Docs-only change; full suite green on the branch base (545 unit + 39 e2e). Spike evidence backing the design is committed on main (`30f1c2e`) with its findings in `docs/rfp-solver-net-flow.md`'s decision log.

## Deviations

None — new document following `docs/rfp-template.md`.

### DIFF
--- a/docs/rfp-fulgora-scrap.md
+++ b/docs/rfp-fulgora-scrap.md
@@ -1,0 +1,219 @@
+# RFP: Fulgora scrap-based layouts
+
+## Summary
+
+Support production chains that START from scrap and work downward —
+Fulgora's inverted economics. The solver side is already proven: the
+2026-07-11 spike (see `docs/rfp-solver-net-flow.md` decision log)
+landed 310 real recycling recipes (excluded by default), opt-in
+`NetflowOptions { allow_recycling, allow_voiding }`, and demonstrated
+the LP's scrap arithmetic is exact (holmium golden to 4 decimal
+places) with value-recovery recycling activating for free. This RFP
+covers what remains: the **voiding policy** (the spike falsified the
+obvious mechanism), **solid surplus routing**, the **recycler row +
+item sorting** problem (the layout centerpiece — recycler output is
+inherently mixed), and end-to-end Fulgora fixtures.
+
+The scope cut that shapes everything below: **voiding is a layout
+policy, not a solver objective.** The spike proved that pricing
+surplus high enough to make the LP void also makes it profitable to
+craft entities purely to recycle them (a laundering family the cycle
+guard refuses loudly, killing the solve). Rather than fight the
+objective function, the solver keeps reporting surplus honestly at the
+frozen cost table, and the LAYOUT decides what to do with surplus
+streams: export them to the perimeter, or synthesize voider rows that
+destroy them — the same way it already synthesizes balancers and
+output mergers without solver involvement.
+
+## Motivation
+
+Fulgora is the one place the game inverts production: a single input
+(scrap) fans out into ~12 probabilistic outputs, and the challenge is
+consuming/sorting/destroying the excess while extracting the binding
+resource (holmium). Today:
+
+- The solver handles it exactly (spike-verified: `holmium-plate@1/s`
+  → 40.0 scrap/s, hand-derived and matched; recycling columns recover
+  value from the byproduct soup at default pricing).
+- The layout cannot build ANY of it: no recycler row template, no
+  item-sorting story for mixed recycler output, no solid surplus
+  routing (only fluids reach the perimeter today), no voiders.
+- The UI can't express "scrap as raw input" chains.
+
+Concrete target the RFP must make real: `electromagnetic-science-pack`
+at a modest rate from `{scrap, water}`, 0 validation errors, with
+every one of scrap's ~12 output streams either consumed, exported, or
+voided — nothing stranded.
+
+## Design
+
+### D1. Voiding policy: curated, layout-level
+
+The solver's `allow_voiding` machinery (validated in isolation by the
+spike) stays available but is NOT the shipping mechanism. Instead:
+
+- The solver runs Fulgora chains with `allow_recycling: true` and
+  default pricing. Surplus lands in `surplus_outputs`, honestly.
+- `LayoutOptions` gains `surplus_policy: SurplusPolicy` with variants
+  `Export` (today's behavior: perimeter routing) and `Void`.
+- Under `Void`, the layout synthesizes **voider rows** for solid
+  surplus: recycler banks that consume a surplus stream and recycle it
+  down to nothing. The voider cascade is deterministic arithmetic, not
+  an LP decision: an item chain-recycles through its recycling recipes
+  until it reaches self-voiders (X → 0.25X), and a self-feeding
+  recycler bank destroys any input completely. Machine counts derive
+  from gross throughput (each pass returns 25%, so a stream of `r`/s
+  needs recyclers sized for `r / 0.75` gross plus the cascade's
+  intermediate volumes — all computable from the recipe data).
+- The voider row template is the **net-negative variant of
+  `self_loop_row`**: loop corridor recirculating the 25% return, net
+  INPUT from the surplus stream, no export splitter. The spike's
+  break-even test already pins the LP-side arithmetic; the template is
+  new but sits on proven geometry.
+
+Why curated / why this dodges the laundering family: voider rows are
+synthesized ONLY for items in `surplus_outputs` — streams the LP
+already decided are genuinely excess at honest pricing. No recipe
+selection happens at voiding time, so "craft a combinator to recycle
+it" can never arise: crafting a combinator was never in the plan.
+
+### D2. Solid surplus export (needed regardless of policy)
+
+`Export` must work for solids: extend the step-7 output merger to
+`surplus_outputs` entries (solid), reusing the multi-item merge-cursor
++ UG-hop machinery landed for multi-item outputs. Each solid surplus
+stream gets a merge column tiling east of the target's. This is the
+smallest piece of the RFP and is independently valuable (uranium
+chains: the U-238 surplus stream currently errors as stranded).
+
+### D3. Recycler rows and the sorting problem (the centerpiece)
+
+Scrap-recycling's output is inherently MIXED — one machine emits ~12
+item types. The entire bus model assumes one item per lane, and
+`check_belt_item_isolation` errors on mixing. This needs a genuinely
+new sub-system, and its design is gated on **Phase 0 physicals
+verification** (via the factorio-expert pass + draftsman data):
+
+- Recycler footprint (believed non-square) and whether it EJECTS
+  directly onto belts (mining-drill-style) or is inserter-extractable.
+  Both break current template assumptions (square `machine_size`,
+  inserter extraction) in different ways.
+- Filter entities available to sort: filter inserters (blueprint
+  `filter` field) and/or filtered splitters (`filter` on splitter).
+  Neither is modeled today (`PlacedEntity` has no filter field;
+  validators have no filter semantics).
+
+Candidate sorting architectures to evaluate in Phase 0 (with
+`blueprint-analyze` runs over the community corpus for Fulgora sorter
+designs as reference):
+
+- **(a) Filter-inserter sorting**: recyclers drop to a short "sushi"
+  collection belt per row; a bank of filter inserters lifts each item
+  type onto its own bus lane. Pros: inserter mechanics already
+  modeled except the filter field; per-item rates are known from the
+  solver so inserter counts are computable. Cons: the sushi belt
+  segment violates item-isolation as modeled — needs a validator
+  concept of an intentional mixed segment (analogous to the
+  `:selfloop:` exemption, e.g. a `:sushi:` segment class with its own
+  saturation math).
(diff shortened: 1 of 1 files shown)


docs/rfp-fulgora-scrap.md
```diff
@@ -0,0 +1,219 @@
...
+  until it reaches self-voiders (X → 0.25X), and a self-feeding
+  recycler bank destroys any input completely. Machine counts derive
+  from gross throughput (each pass returns 25%, so a stream of `r`/s
+  needs recyclers sized for `r / 0.75` gross plus the cascade's
+  intermediate volumes — all computable from the recipe data).
+- The voider row template is the **net-negative variant of
+  `self_loop_row`**: loop corridor recirculating the 25% return, net
+  INPUT from the surplus stream, no export splitter. The spike's
+  break-even test already pins the LP-side arithmetic; the template is
+  new but sits on proven geometry.
```
storkme: `self_loop_row` is built around a net-POSITIVE `major_item`: a far corridor recirculates the loop-back majority (fed from the machine's OWN output via a priority splitter's loop branch) AND the same splitter's export branch becomes the row's declared bus output (templates.rs:3053-3057). A voider inverts both: the corridor is fed by a bus-TAP consuming 100% of the surplus stream, recirculates only ~25%, and has no export splitter. Vocabulary overlaps (loop corridor, inserter row) but the plumbing is different (bus→corridor instead of splitter→corridor).

"Sits on proven geometry" oversells the reuse. Also: for the 1-item shape, the near_item / near2 belt / dual inserter rows are entirely absent from a voider — dropping them isn't a code deletion in `self_loop_row`, it's a different template that shares vocabulary. If Phase 2 finds the inversion isn't drop-in, that framing has already set the wrong LOC expectations. Suggest hedging to "new template inspired by `self_loop_row`, delivered as a separate design pass in Phase 2".

storkme (commented):
The scope cut is right — "voiding is a layout policy, not a solver objective" correctly reads the spike's finding, and D3's Phase 0 gate is the right shape for the recycler unknowns.

Two structural things worth resolving before Phase 1 starts:

- D2's "un-strand uranium" claim can't work through the merger alone. `placer.rs` emits a single east-flowing belt per row (only `solid_outputs.first()` gets a belt at lines 508, 516, 519, 778), and `RowSpan` carries one `output_belt_y` — the merger has nothing to read for U-238. That machinery is the netflow RFP's own open Phase 2 item ("partitioner audit for multi-solid-output specs"), still unlanded.
- Phase 2's "start immediately" contradicts its own dependency on Phase 0. Voider rows ARE recycler banks (D1), and Phase 0 is what documents recycler footprint and belt-vs-inserter ejection (D3). Phase 2 can't design its template on speculation about those.

Two of the kill criteria also don't bite as-written: KC1 is unfalsifiable ("cannot be modeled at all") and KC2 is a pivot criterion, not a kill. Details inline.

docs/rfp-fulgora-scrap.md
```diff
@@ -0,0 +1,219 @@
...
+it" can never arise: crafting a combinator was never in the plan.
+
+### D2. Solid surplus export (needed regardless of policy)
+
+`Export` must work for solids: extend the step-7 output merger to
+`surplus_outputs` entries (solid), reusing the multi-item merge-cursor
++ UG-hop machinery landed for multi-item outputs. Each solid surplus
+stream gets a merge column tiling east of the target's. This is the
+smallest piece of the RFP and is independently valuable (uranium
+chains: the U-238 surplus stream currently errors as stranded).
```
storkme: The uranium fixture claim doesn't hold with today's row templates. `placer.rs` only emits an east-flowing belt for `solid_outputs.first()` (lines 508, 516, 519, 778), and `RowSpan` carries a single `output_belt_y` for the merger to read (output_merger.rs:64). Uranium-processing produces U-235 and U-238 from the same machine; the LP will report U-238 in `surplus_outputs`, but the row spec has no belt at the machine port for the merger to extend.

The netflow RFP's decision log flags this exact hole under Phase 2 ("partitioner audit for multi-solid-output specs") — it hasn't landed. Also, "reusing the multi-item merge-cursor" is a category error: that machinery merges N producer rows each producing ONE item, tiled east; it does not invent a second east-flowing belt on a row that already has one.

Two ways forward: (a) extend D2 to include multi-solid-output row templates (which makes it substantially bigger than "the smallest piece of the RFP"), or (b) replace the uranium example with a case where the surplus item comes from a DIFFERENT producer row than the target (the merger's existing shape).

docs/rfp-fulgora-scrap.md
```diff
@@ -0,0 +1,219 @@
...
+
+- **(a) Filter-inserter sorting**: recyclers drop to a short "sushi"
+  collection belt per row; a bank of filter inserters lifts each item
+  type onto its own bus lane. Pros: inserter mechanics already
+  modeled except the filter field; per-item rates are known from the
+  solver so inserter counts are computable. Cons: the sushi belt
+  segment violates item-isolation as modeled — needs a validator
+  concept of an intentional mixed segment (analogous to the
+  `:selfloop:` exemption, e.g. a `:sushi:` segment class with its own
+  saturation math).
```
storkme: The `:selfloop:` exemption lives in `check_belt_loops` (belt_structural.rs:4), not `check_belt_item_isolation` (line 430). Item-isolation errors on any two adjacent belts where `carries` differs — the mixed-belt problem here. Bolting on a segment-tagged exemption raises three boundary questions the RFP doesn't touch:

1. What does the check do when a `:sushi:` belt is adjacent to an ordinary per-item bus belt? The sushi → filter-inserter → per-item bus transition is required by architecture (a) and MUST be validator-safe.
2. How is per-item saturation math expressed when `carries` on a `:sushi:` tile is untagged / heterogeneous?
3. What's the containment mechanism? KC5 asserts containment is required but doesn't say WHERE the exemption lives — without that, KC5 isn't evaluable.

Naming the mechanism now — even at "item-isolation looks up segment kind by segment_id prefix; sushi ↔ non-sushi boundaries fall to a separate check verifying filter-inserter presence" precision — would make KC5 something you can actually trip.

docs/rfp-fulgora-scrap.md
```diff
@@ -0,0 +1,219 @@
...
+  already lists Fulgora targets.
+- The `recycling-or-hand-crafting` category quirk (scrap-recycling's
+  actual category) is already handled by the spike's mapping.
+
+## Kill criteria
+
+1. **Physicals**: if Phase 0 finds the recycler cannot be modeled in
+   the current row abstractions at all (e.g. direct ejection with no
+   viable collection geometry under our belt model), stop and rescope
+   before any template work — the sorting architecture depends on it.
```
storkme: "cannot be modeled AT ALL" isn't a bar this project trips — some geometry always works if we're willing to add code. This is effectively "we'll stop if it's physically impossible," which is the RFP template's bad-example shape.

The actual load-bearing bet in D3's design is that recycler physicals fit inside the existing row-template category (single_input / dual_input / oil_refinery / self_loop) without needing a new `RowKind`. Tighter phrasing: "if fitting the recycler forces a new `RowKind` variant AND either (a) on-belt ejection forces a machine that no current inserter-extractable template can wrap, or (b) collection geometry needs the bus abstraction to carry multi-item lanes at the template level (not just the sushi segment), stop and rescope." That's falsifiable in Phase 0 without a moral judgment call.

docs/rfp-fulgora-scrap.md
```diff
@@ -0,0 +1,219 @@
...
+2. **Sorter viability**: if neither sorting architecture can express a
+   12-way sort within ~2× the footprint of community reference
+   designs (measured via `blueprint-analyze` on the corpus), the bus
+   model is the wrong substrate for scrap sorting — stop and consider
+   a dedicated "sorter block" pre-solved template (balancer-library
+   style) instead of generative geometry.
```
storkme: "Stop and consider a different implementation approach" is a pivot criterion, not a kill. If the pre-solved "sorter block" is a viable Plan B, it belongs in the Design section as an explicit alternative — with its OWN kill trigger — not as the kill fallback for D3(a)/(b). Otherwise the RFP hasn't actually committed to when it'd abandon scrap-based layouts vs. just switch tactics inside them.

Concrete question worth answering: what evidence would make you actually stop shipping ANY Fulgora support in this codebase? "Both approaches fail AND the pre-solved block also fails" would be a kill; the current text stops one step short of that.

Also, "~2× the footprint of community reference designs" needs a corpus and a measurement protocol nailed down BEFORE Phase 0 runs, or it'll be tuneable post-hoc (the netflow RFP's frozen-cost-table pattern is the precedent).

docs/rfp-fulgora-scrap.md
```diff
@@ -0,0 +1,219 @@
...
+  `SurplusPolicy::Void`; voider-purity fixture.
+- **Phase 3 — recycler rows + sorter**: per the Phase 0 decision;
+  scrap-recycling row template + sorting into per-item bus lanes;
+  `holmium-plate` fixture goes green.
+- **Phase 4 — UI/API**: scrap flow exposure, surplus-policy toggle,
+  `electromagnetic-science-pack` fixture; CLAUDE.md ladder gains a
+  Fulgora tier.
+
+Phases 1 and 2 are independent of the Phase 0 decision and can start
+immediately; Phase 3 is the long pole.
```
storkme: Phase 2 builds voider rows out of recycler banks (D1). Whether the recycler ejects onto belts directly or is inserter-extractable — and its footprint — are Phase 0 outputs (D3 flags them as unknown). Phase 2 can't design a template on speculation about that: the machine-port inserter geometry AND the machine-count / row-height math both change.

Phase 0 actually has TWO artifacts: (i) recycler physicals, (ii) sorter architecture decision. Phase 2 depends on (i) but not (ii). Suggest splitting the Phase 0 output artifacts explicitly and saying Phase 2 waits on (i), Phase 3 waits on both. As-written the reader takes away "Phase 2 unblocked" when it isn't.

storkme (commented):
Revision cleanly resolves all six earlier findings; each landed with real substance rather than acknowledgment. Approving in intent (self-approve blocked by GitHub — recording as a review comment).

- **D1 voider template**: hedged to "inspired by `self_loop_row`", explicitly calls out the bus-tap-fed inversion (corridor fed from surplus, no export splitter) and defers to a Phase 2 design pass. Reuse claim now scoped to vocabulary.
- **D2 split**: D2a covers streams with an existing east belt (single-solid-output rows, and post-Phase-3 sorter lanes); D2b names the actual gap — `placer.rs`'s single `output_belt_y` per `RowSpan` and `solid_outputs.first()` limiting the belt — inherits the netflow RFP's open multi-solid-output item, and correctly moves the uranium fixture behind D2b. Verified against `placer.rs:74` and `:516`.
- **Sushi exemption (D3(a))**: mechanism is now implementation-precise — `:sushi:` segment-id prefix in `check_belt_item_isolation`, sushi↔sushi skipped, sushi↔ordinary owned by a new companion check that requires filter inserters plus per-item saturation math (Σ per-item rates vs belt capacity). `:selfloop:` precedent correctly re-attributed to `check_belt_loops`. KC5 is now evaluable against this.
- **KC1**: rephrased to a concrete `RowKind`-variant + ejection/collection-geometry test; falsifiable in Phase 0 from the physicals artifact alone.
- **KC2**: architecture (c) pre-solved sorter block promoted to first-class alternative alongside (a)/(b), Phase 0 evaluates all three. Pivot ((a)/(b) → (c)) and actual kill (all three fail → stop Fulgora layout support, keep solver-side) are separated. Benchmark protocol frozen in D3 (bbox area / entity count of ≥10-item sorter at ≥1 full yellow, corpus committed before measurement).
- **Phase 0 artifact split**: Phase 0 output cleanly split into (i) physicals / (ii) sorter decision. Phase 1 and 2b start immediately, Phase 2 gated on (i), Phase 3 on both. D1's voider-rows-are-recycler-banks dependency is also stated inline.

No new internal contradictions: KC5's "switch to (b) or the pre-solved block" is consistent with (c) being first-class, KC1(b)'s "beyond the tagged sushi segment" reads correctly as an upper-bound on exemption breadth, and the Verification/Fixture list matches the revised phasing.

Note (informational, not blocking): D2a ships before Phase 3, and the RFP doesn't cite a concrete non-uranium single-solid-output surplus case it un-strands today — its main value materializes when the sorter lands. Called out honestly by the "and (once Phase 3 lands)" clause; worth confirming an actual D2a-only fixture exists before Phase 1 ships, but not a design issue for this RFP.

Ship it.

PR #360, opened 2026-07-22
docs(rfc-048): end-state definition — no structural exemptions

storkme:
One decision-log entry recording the agreed end-state for cell composition (user steer, 2026-07-22): no chain class structurally excluded — oil complexes as pipe-ported mega-cells (fluid loops stay intra-cell), high rates via pre-verified corridor/distribution cells, the two permanent contract disciplines named, bus engine as fallback-during-growth. Sections omitted: single doc entry.

### DIFF
--- a/docs/rfc-048-cell-composition.md
+++ b/docs/rfc-048-cell-composition.md
@@ -285,6 +285,23 @@ entries as the Phase 1 gate.
 
 ## Decision log
 
+- *2026-07-22 — End-state definition set (user steer, post-#359): the
+  method is not "done" with structural exemptions. Revised target: NO
+  chain class structurally excluded — the existing tier ladder
+  (through the refinery tiers) reachable by composition, phased:
+  solids with AC-from-plates as the Phase-2 gate; the oil complex as
+  an opaque pipe-ported MEGA-CELL (fluid loops are cell INTERIORS —
+  the boundary contract is simple even when the inside is stateful;
+  gated on the harness's fluid calibration, which is why the Phase-1
+  fluid entry is an enabler, not a side quest); high rates via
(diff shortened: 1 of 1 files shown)


storkme (commented):
Session-side review (bot green but silent — verified from the run log it executed the review with posting enabled and had nothing to post; consistent with the triviality gate on a 17-line docs-only diff, not the broken pattern).

**Consistency checks all pass:**
- "one S=4 express corridor = 180/s" matches F3's sim-measured 179–186/s.
- "kill 5 applied to corridors" is coherent — kill criterion 5 is the premise-falsification criterion bounding corridor machinery, and extending it to distribution cells as a permanent discipline is a faithful generalization.
- "Phase-1 fluid entry as enabler" lines up with the PR #359 fold (fluid consumer as the second sim-verified catalog entry).
- Decision-log ordering (newest first) and the provenance marker ("user steer, post-#359") follow convention; the documentation-of-autonomous-decisions requirement is satisfied.

**Two minor tensions, both optional to address (PR body says the single-entry scope was deliberate):**
1. "What this is NOT" still says the bus engine "remains the fallback for whatever cells can't express" (open-ended), while this entry revises the end-state to fallback-*during-catalog-growth* with no chain class structurally excluded. A one-line pointer from the Status line or the NOT-section would keep body readers from getting the stale framing.
2. The Phasing section assigns fluids to Phase 2's integration RFC; the entry inserts a solids AC-from-plates Phase-2 gate and defers oil behind the Phase-1 fluid calibration. Not contradictory, but the phase map now lives partly in the log.

Ship it as-is; both nits are cheap follow-ups whenever the doc next gets touched.

PR #365, opened 2026-07-22
RFC-048 Phase 1: cell composition delivers — composed EC@15 runs at plan in headless Factorio

storkme:
## Intent

RFC-048 Phase 1: catalog + extraction + composition harness (test-only, kill-4-compliant: cells are engine-generated), with the existence proof measured, all five kill criteria evaluated with real data, and a documented GO for the Phase-2 integration RFC.

## Headline

**A cell-composed factory runs in headless Factorio at exactly its planned rate** — 15.0 EC/s produced, 15/15 machines working, converged — on the config the bus engine refuses outright (#336's EC@15-from-plates). The composed layout validates at 0 errors (now a permanent gate test), and the sim adjudicated the 6 inherited validator warnings as attribution false-positives (machines claimed under-fed measured at full rate).

## Scope

- `crates/core/tests/cell_composition.rs`: segment-partition cell extraction (`row:*` + poles + machines), contiguity-derived lane-aware ports with terminal coordinates, polyline corridor stamping (straight/B11/UG-weave/splitter-merge), calibrated-orientation composer (north feeds / south drains), boundary-record population for `export_with_manifest`, permanent gates + measurement probes.
- RFC-048 doc: full Phase-1 decision-log trail (three sim-diagnosed harness gotchas → #363 data + composition rules; #364 fluid-path attribution), kill 1–5 evaluation (all PASS), gate dispositions, GO for Phase 2.
- Issues filed en route: #363 (feed-rig geometry: two live datasets + consumer-side rules), #364 (fluid path delivers nothing for ANY layout — controlled attribution with the engine's own plastic layout as the failing control; full diagnostic bundle + recommended first move attached).

## Verification actually run

- Suite 886/0/46 (single clean invocation), goldens 8/8 zero diffs; the two new permanent gates green.
- Headless sim runs (2.0.76): EC@15 composed — converged, produced 15.0/s (report preserved in the decision log; scenario names recorded). Plastic: composition 0/0; sim blocked by the harness fluid path, proven harness-side by the engine-control run.
- Axis measurement: EC cell regenerated at AM2 from one changed string — variants are parameters; the #360 end-state is a plan.

## Deviations from agreed approach

- Gate (a)'s fluid half closes as "composition-proven, sim blocked cross-RFC (#364)" — the fluid entry's stated purpose (calibration forcing function) is delivered as a datum rather than a passing measurement. All other gates fully met.
- The composer encodes two sim-kit workarounds (#363) that should migrate harness-side.


storkme (commented):
Deep session-side review. Context: **no CI ran on this PR at all** (head SHA has 0 check runs — neighboring PRs at 19:32/19:38 triggered fine, so investigate), **no bot review**, and the PR is **CONFLICTING** with main (#360's nit-fold + #361's review-fold both landed after the branch point). I couldn't re-run the suite (no toolchain in my env), so instead I verified claims against the committed artifacts in the dev worktree (`.claude/worktrees/build-quality`): decoded `rfc048-ec15.bp` / `rfc048-plastic.bp` tile-by-tile. Two doc numbers don't survive contact with the artifacts.

## Findings from artifact verification (the important ones)

**1. Kill 3's numbers are stale — the sim-verified layout is 110×22=2420 tiles / 461 entities, not 89×22=1958 / 422.** Timeline: 89×22 was recorded at `502bcfe` (19:56); the 4-tile feed-pitch fix `4767ed7` (20:01) widened every pair; the exported artifact used for the successful sim (mtime 20:05, two minutes before the sim-success commit) decodes to x∈[0,109], y∈[0,21] → 110×22 = **2420 tiles, 461 entities = 2.48× the 975 extrapolation**, not "2.0× at the boundary". Kill 3 still passes on the compensating-win clause (engine refusal + measured-at-plan) — but it passes *over* the boundary, not at it, and the close-out should say so with the real numbers. The gate prints dims, so this is a one-command confirmation.

**2. The "2→1 fast-splitter merge" is built as a sideload; the splitter takes one input.** First pair's splitter sits at tiles (21,9)-(21,10) facing east. Top half fed by an east-facing belt at (20,9) ✓. South half's west neighbor at (20,10) is a **north-facing** belt — the b-run corners north at x=sx-1 and sideloads into the a-run's tail tile one tile west of the splitter. The splitter's second input and second output are both unconnected; the merge happens by sideload. Root cause: the corner geometry (`cell_composition.rs:756-758`) only approaches the splitter's south half eastward if `o2.y == o1.y+1` — but the outs are at y=9 and y=16 (two internal rows, 2+1 split), 7 apart. Functionally fine (the sim proves 15/s flow; rate math has headroom), but: (a) the milestone entry and kill 5's machinery list describe a splitter merge that was never built; (b) the merge is exactly the sideload form F4's invariant exists to catch — invisible lane-count degradation, here in a corridor rather than at a port. Fix: `assert_eq!(o2.y, o1.y + 1)` to pin the assumption loudly, or route the b-run to approach eastward (corner at `sx-2`), or drop the decorative splitter and own the sideload in the doc.

**3. The plastic petroleum column works today by coincidence, and nothing can catch it drifting.** `cell_composition.rs:597-609` hardcodes the column at x=6; the artifact shows it connects because the cell's max-x petroleum pipe happens to sit at cell-local x=2 (cx=4 → absolute 6). Nothing asserts this. Per the RFC's own kill-2 text, corridor pipe runs are outside the fluid checks' model ("a discontinuous pipe validated clean"), and the sim — the only oracle that could catch it — is blocked by #364. So the gate cannot fail on the exact failure mode the doc calls load-bearing. Fix: derive the column x from `cx + pipe_terminal.x` and assert the column terminates adjacent to a petroleum-segment tile.

## Process / docs findings

**4. Merge conflict + dark CI.** Rebase onto main; the resolution must preserve both #361's gate text and this close-out, and (see 5) reconcile the gate letters. Before merge, get CI to actually run — a 1k-line PR with zero check runs and a conflict should not land dark regardless of local-suite claims.

**5. Gate letters (a)/(b)/(c) are never enumerated anywhere.** The close-out says "Gate (a) — sim verification… Gate (b): delivered and measured. Gate (c): variant-is-a-parameter" — neither the branch text nor #361's amended gate paragraph defines these labels, and (b) is undecodable from the doc ("delivered" what?). Enumerate them in the rebase.

**6. The never-finishing tripwire isn't serviced.** #361 (main, fbe8c13) requires measuring fixture-ladder catalog coverage at each phase close-out. This close-out evaluates kills 1–5 but never states coverage. One sentence fixes it — today: 2 ladder chains (EC, plastic), everything else bus-owned, coverage clock starts now. Branch predates #361 by ~10 min, so this is a merge-resolution item, but the tripwire only works if the datum exists at every close-out.

**7. Docs taxonomy:** `docs/status.md` has no RFC-048 entry (status-ledger duty — RFC-049 got a close-out entry; a Phase-1 close-out + GO for Phase 2 is exactly ledger material), and the `rfcs.md` registry row still says "Spike (Phase 0 recon open; PR #341)" — stale on main already, staler after this.

**8. PR body says "report preserved in the decision log"** — the log records the headline numbers (15.0 produced / 14.4 delivered / 15/15 working) but the report itself isn't in the repo (I searched the dev worktree; the .bp/.manifest artifacts are there, no report). Paste the sim summary block into the log or attach it to the PR.

**9. Minor code nits:** `cell_composed_ec15_zero_errors` tolerates any *number* of `inserter-item-throughput` warnings — pin ≤6 so silent growth trips. `export_composed_plastic_for_sim` is a permanent gate named like an artifact producer and writes `target/tmp` files on every suite run — rename or split.

## What I verified that checks out

- Boundary records in the manifest match the calibrated orientation (south-facing feeds/drains, west→east ordering) ✓
- Plastic composition geometry: coal column corners east into the port terminal-1; out corners south ✓
- The config-axis measurement (gate c) delivers exactly what the whole-RFC review asked, and the verdict's framing is right: *"a variant is a parameter, not a project — the real axis cost is sim-verifying variants actually used."* That correctly relocates config-axis cost from authoring to verification, which is the honest version of "the end-state is a plan."
- Deviation honesty: gate (a)'s fluid half closed as datum-not-measurement, with #364's controlled attribution (the engine's own plastic as failing control) — textbook.
- The warnings discipline is implemented, not just prose: the gate pins the sim-disproven warning *category* rather than blanket-tolerating.

**Verdict:** approve-in-principle after rebase. The substantive asks are small: two one-line asserts/derives (#2, #3), corrected kill-3 numbers (#1), and doc reconciliation (#5–8). The phase-1 story itself — composed factory runs at plan where the engine refuses — stands up to tile-level scrutiny.

PR #375, opened 2026-07-22
RFC-051: cell composition integration (production path) — design for review

storkme:
## Intent

Author the Phase-2 integration RFC that RFC-048 Phase 1's GO verdict commissioned (see #365's close-out): convert cell composition from a test-only harness into a flag-gated production layout path, gated on **advanced-circuit-from-plates composed at 0 errors + sim-verified at plan**.

## Scope

Doc-only: `docs/rfc-051-cell-composition-integration.md` + registry row (RFC-051 claimed after a fresh origin + open-PR collision check).

Design highlights for review attention:
- **No static cell catalog** — cells generate at layout time (Phase 1's variant-is-a-parameter verdict); the checked-in artifact is a sim-verified *registry* (config hash → verdict), since CI can't run the sim.
- **AC-from-plates as the gate is deliberately one notch harder than Phase 1**: fan-out 2 (cable feeds EC *and* AC) forces the splitter fan-out corridor.
- **Sim-kit scaffolding stays out of production geometry** (Phase 1's 2.48× area carried ~24% rig overhead); a `calibrated_for_sim` mode preserves the verification path.
- **No-value kill**: composition must either handle a bus-engine refusal or win the existing scorer on warnings somewhere — no score biasing to make it win.
- Flag default Off; zero golden re-blesses guaranteed by construction until the flip, which is its own logged decision.

## Verification actually run

Doc-only PR — no code paths touched; suite untouched. Registry collision check: `git show origin/main:docs/rfcs.md | tail` says next = 051; no open PR or `rfc-05*` branch claims it.

## Dependencies / sequencing

- Phase A (the lift) starts after #365 (the harness this lifts) merges.
- The gate's sim leg uses #373's pipe-to-ground export fix for any fluid-adjacent verification (composed plastic re-verification closes RFC-048 gate (a) fully once both land).

### DIFF
--- a/docs/rfc-051-cell-composition-integration.md
+++ b/docs/rfc-051-cell-composition-integration.md
@@ -58,7 +58,13 @@ splits into production modules and thin test consumers:
   attachment emits ordinary `boundary_inputs`/`boundary_outputs` records
   at the layout edge, same contract as bus layouts. A
   `calibrated_for_sim` mode keeps the rig-compatible geometry for
-  verification runs.
+  verification runs, under a stated and **asserted invariant** (review
+  fold, 2026-07-22): a production layout and its calibrated twin are
+  **bit-identical in every non-boundary entity** — cell interiors,
+  corridor topology, port attachment — differing only in the boundary
+  scaffolding outside the ports. A parity test diffs the two entity
+  sets minus boundary segments; verification provenance holds only as
+  long as that diff is empty, so it fails the build, not a hope.
 - Solver→cell rounding: per-chain ratio quantum from solver machine
   counts (e.g. 3 cable : 2 EC = 5 EC/s), K copies rounded up, headroom
   explicit in the candidate's metadata.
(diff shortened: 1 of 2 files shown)


storkme (commented):
Session-side review (bot still in progress at time of writing). Read against the decomposition-search source and the #365 evidence trail.

**Claims verified against the code:**
- The integration seam is real and correctly described: `DecompositionCandidate::produce(&SolverResult, &LayoutOptions) -> Result<LayoutResult, String>` (`decomposition_search.rs:65`) is exactly the right shape for `CellComposedCandidate`; candidates are validated via `validate()` with error classes feeding acceptance (:512). "No new decision machinery" is accurate for the *slot*.
- Registry arithmetic correct (next was 051; collision check documented).
- The RFC visibly metabolized the #365 review: 2.48× area with ~24% rig overhead cited, sim-kit scaffolding excluded from production geometry, tripwire inherited as kill 5. Good folding.

**Findings:**

1. **The sim-registry hash has a hole: no engine-geometry component.** Cells generate at layout time from the live engine; the registry keys (recipe, tier, modules, quality, rate quantum) don't include anything that changes when row templates / placer / inserter logic change. Same config hash, different geometry, stale verdict — "pre-verified cell" silently decays into an unverified one, which is precisely the drift class the whole method exists to kill. Add an engine-geometry version to the key (a hash of the generated cell entities would be self-maintaining) or a stated re-verification policy on layout-engine changes.

2. **`calibrated_for_sim` divergence needs a stated invariant.** Phase 1's own lesson — the sim catches what the validator can't — is strongest when the measured artifact *is* the shipped artifact. If production ships boundary records while the sim measures rig-scaffolded geometry, the RFC should state exactly what is bit-identical between the two (cell interiors, corridor topology, port attachment) and make it an asserted invariant (production layout and its calibrated twin share all non-boundary entities), not a hope. Otherwise verification provenance erodes one scaffold at a time.

3. **"Wins the existing scorer on warnings" (kill 3b) vs the scorer as it exists.** `CandidateScore` is density/overproduction/entity-count with error-class acceptance; I don't see a warning-kinds term in the score itself. If kill 3(b) is to be meaningful, either the scorer needs a warnings term — which *is* new decision machinery, shared by all candidates, and should be logged as such — or 3(b) should be reworded to acceptance-class terms. Small point, but kill criteria should be decidable as written.

4. **Lift should not carry the two known fragilities verbatim.** #365 review: the splitter merge assumes adjacent out-ports (today it silently builds a sideload — tile-level proof in that review) and the plastic pipe column x is a validator-invisible hardcode. Phase A's generalization is the natural place to fix both; worth naming in the RFC so the parity gate doesn't enshrine them (parity is "reproduce Phase-1 results", and one of those results is a wrong-geometry corridor).

5. **Nit:** the rfcs.md edit touches the registry without backfilling RFC-048's row ("Spike (Phase 0 recon open; PR #341)" — two phases stale), despite the doc header's standing instruction to backfill on touch.

**Verdict:** the design is the right shape — flag-gated, scorer-arbitrated, kill criteria with teeth (the no-value kill especially: it forbids the integration from rationalizing itself). Ship after the registry-hash hole and the calibrated-twin invariant are addressed in text; both are one-paragraph fixes.

PR #376, opened 2026-07-22
RFC-049 Phase 2: tech-state parity (#370) + sim-calibrated level-aware input rating

storkme:
## Intent

Close RFC-049's kill 2 the way it asked to be closed: with measured data. The validator's belt→machine input rating was deliberately flat at every inserter-capacity level ("no linear extrapolation on belt-pickup sides without measured data"); the sim harness now provides the measurement, and the measurement says the extrapolation is safe.

**Stacked on #362** (needs its kit fix + forensic dumps for trustworthy calibration runs) — merge #362 first.

## Scope

1. **Tech-state parity (#370, sim-harness)**: the scenario sets the force's inserter-capacity bonuses to the manifest's `inserter_capacity` level instead of inheriting `research_all`'s bonus 7. Probe-verified force-field model: non-bulk hand = 1 + `inserter_stack_size_bonus`; {bulk, stack} hands = {1, 5} + `bulk_inserter_capacity_bonus` — reproduces all three I8b tables exactly (L0 → hands 1/2/6, L7 → 4/12/16). Realized bonuses dumped into every report as the verification channel. Baselines key tech_state on the level (blessed files need one re-bless when this lands).
2. **Calibration matrix**: 15 cells (5 types × L{0,2,7}), flooded express feed → quad-speed-3 AM3 stone-wall sink (37.5/s ceiling), kit self-audit clean on every cell. Full table in the RFC-049 decision log.
3. **`common::machine_feed_rate`** = `flat × hand(L)/hand(0)` — conservative in every measured cell (+4–30% margins; the belt-drop swings×hand decomposition measured ~12% *optimistic* for machine feeds and is not used on the input side). Bit-identical to the flat constant at L0.
4. **Both input-side call sites** in `check_inserter_throughput`/`check_inserter_item_throughput` switched to it.
5. **Not changed**: the sizing ladder stays L0-flat — L>0-aware placement (denser layouts requiring research) is a user-facing trade deferred to its own decision.

This also adjudicates #352's four warnings: correct at the L0 semantics they were computed under; the clean-kit sim PASS was measured pre-parity at research-inflated hands and does not contradict them.

## Verification

- `cargo test` full core suite: **787 unit + 60 e2e, 0 failures** — the L0 bit-identity means zero behavior change for every existing fixture.
- Harness tests 39 green; parity probes at L0/L7 confirm exact force-bonus values.
- Calibration reports in session scratchpad (`calib/report-*.json`); numbers transcribed to the RFC-049 decision log.

## Deviations

None from the discussed plan. Validator-semantics change: per house rules this wants the CI bot **plus** a local adversarial review before merge.

### DIFF
--- a/crates/core/src/common.rs
+++ b/crates/core/src/common.rs
@@ -571,14 +571,28 @@ pub fn inserter_throughput(name: &str, quality: QualityTier) -> f64 {
 ///
 /// This closes RFC-049 kill 2 ("no linear extrapolation on belt-pickup
 /// sides without measured data") WITH the measured data: the 2026-07-22
-/// sim calibration matrix (5 types × L{0,2,7}, flooded express feed into
-/// a 37.5/s sink; RFC-049 decision log) measured hand-ratio scaling as
-/// conservative for every type — non-bulk measured/model 1.04–1.19,
-/// stack 12.6/20.8/34.3 vs this rule's 12/16/32. The swings×hand
-/// decomposition used on belt-DROP sides is ~12% optimistic for machine
-/// feeds and is deliberately not used here. At L0 the ratio is 1 for
-/// every type — bit-identical to the flat constant.
+/// sim calibration matrix (all 8 levels for stack/bulk, L{0,2,7} for
+/// non-bulk whose hand values {1,2,4} those cells fully cover; flooded
+/// express feed into a 37.5/s sink; RFC-049 decision log).
+///
+/// Non-bulk and bulk use hand-ratio scaling — conservative in every
+/// measured cell (non-bulk margins 1.04–1.19×, bulk 1.93–2.27×). The
+/// stack inserter does NOT: its measured machine-feed curve is
+/// non-monotone in hand size (dips at hands 7 and 14 — swing-cycle
+/// quantization), and hand-ratio over-credits at L1 (14.0 vs 12.38
+/// measured) and L6 (28.0 vs 23.88) — caught by the #376 adversarial
+/// review, confirmed by completing the matrix. Stack credits therefore
+/// come from a measured floor table: monotone, 3–8% under every
+/// measured cell, and exactly the flat 12.0 at L0 (bit-identity
+/// preserved). The belt-DROP swings×hand decomposition is ~12%
+/// optimistic for machine feeds and is deliberately not used here.
 pub fn machine_feed_rate(name: &str, quality: QualityTier, level: u8) -> f64 {
+    if name == "stack-inserter" {
+        // Measured intakes: 12.62, 12.38, 20.75, 20.75, 23.12, 24.00,
+        // 23.88, 34.25 (Normal quality). Floors below every cell.
+        const STACK_FEED: [f64; 8] = [12.0, 12.0, 19.2, 19.2, 21.6, 22.8, 22.8, 32.0];
+        return STACK_FEED[usize::from(level.min(7))] * quality.multiplier();
+    }
     inserter_throughput(name, quality) * inserter_hand(name, level) / inserter_hand(name, 0)
 }
 
@@ -875,12 +889,23 @@ mod tests {
         }
         assert!((machine_feed_rate("inserter", Normal, 7) - 0.84 * 4.0).abs() < 1e-9);
         assert!((machine_feed_rate("fast-inserter", Normal, 7) - 2.31 * 4.0).abs() < 1e-9);
-        assert!((machine_feed_rate("stack-inserter", Normal, 7) - 12.0 * 16.0 / 6.0).abs() < 1e-9);
-        // Credits stay below the sim-measured intake floors
-        // (calibration matrix 2026-07-22: stack measured 20.75 at L2,
-        // 34.25 at L7) — the check must never over-credit.
-        assert!(machine_feed_rate("stack-inserter", Normal, 2) <= 20.75);
-        assert!(machine_feed_rate("stack-inserter", Normal, 7) <= 34.25);
+        // Credits stay below the sim-measured intake floors at EVERY
+        // level (complete 8-level matrix, 2026-07-22 — the #376 review
+        // caught hand-ratio over-crediting stack at the unmeasured
+        // L1/L6 cells; measurements confirmed: the stack curve is
+        // non-monotone in hand). The check must never over-credit.
+        const STACK_MEASURED: [f64; 8] = [12.62, 12.38, 20.75, 20.75, 23.12, 24.00, 23.88, 34.25];
+        const BULK_MEASURED: [f64; 8] = [5.38, 8.00, 10.88, 13.00, 15.50, 20.75, 23.12, 27.75];
+        for l in 0..8u8 {
+            let s = machine_feed_rate("stack-inserter", Normal, l);
+            assert!(s <= STACK_MEASURED[l as usize], "stack L{l}: credit {s} > measured");
+            let b = machine_feed_rate("bulk-inserter", Normal, l);
+            assert!(b <= BULK_MEASURED[l as usize], "bulk L{l}: credit {b} > measured");
+            if l > 0 {
+                // Monotone: more research never credits less.
+                assert!(machine_feed_rate("stack-inserter", Normal, l) >= machine_feed_rate("stack-inserter", Normal, l - 1));
+            }
+        }
         assert!(machine_feed_rate("fast-inserter", Normal, 2) <= 5.38);
     }
 
--- a/crates/core/src/common.rs
+++ b/crates/core/src/common.rs
@@ -565,6 +565,23 @@ pub fn inserter_throughput(name: &str, quality: QualityTier) -> f64 {
     base * quality.multiplier()
 }
 
+/// Throughput of an inserter feeding a **machine** (belt→machine input
+/// side) at inserter-capacity research `level`: the flat I8 constant
+/// scaled by the hand-size ratio `hand(L)/hand(0)`.
+///
+/// This closes RFC-049 kill 2 ("no linear extrapolation on belt-pickup
+/// sides without measured data") WITH the measured data: the 2026-07-22
+/// sim calibration matrix (5 types × L{0,2,7}, flooded express feed into
+/// a 37.5/s sink; RFC-049 decision log) measured hand-ratio scaling as
+/// conservative for every type — non-bulk measured/model 1.04–1.19,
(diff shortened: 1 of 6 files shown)


storkme (commented):
**Note: GitHub won't let the PR author's own token submit REQUEST_CHANGES, so this is posted as COMMENT — treat the first inline finding as blocker-severity, not a style note.**

Verified the mechanical claims rather than just reading the diff. L0 bit-identity is real (`inserter_hand(name, 0)` cancels for every type by construction; the pinned test plus a full local `cargo test --manifest-path crates/core/Cargo.toml` both come back at 787 unit + 60 e2e, matching the PR body exactly). The two call sites in `inserters.rs` only feed `input_avail` — output side is correctly re-shadowed by `belt_drop_throughput` a few lines later, so no double-credit. Quality isn't double-applied (`inserter_hand` doesn't take quality at all). The `#[cfg(test)]`-gated `inserter_throughput` import is legitimate hygiene, not dead weight.

One finding below is serious enough to hold this: the calibration matrix only sim-measured 3 of the 8 research levels per inserter type, and the code applies the same formula to all 8 with no fallback for the unmeasured ones — which is what RFC-049's own kill criterion 2 says not to do. The other three are test/claims gaps in the sim-harness half that weaken confidence in the 'verified' framing without being wrong per se.

crates/core/src/common.rs
```diff
@@ -565,6 +565,23 @@ pub fn inserter_throughput(name: &str, quality: QualityTier) -> f64 {
...
+/// This closes RFC-049 kill 2 ("no linear extrapolation on belt-pickup
+/// sides without measured data") WITH the measured data: the 2026-07-22
+/// sim calibration matrix (5 types × L{0,2,7}, flooded express feed into
+/// a 37.5/s sink; RFC-049 decision log) measured hand-ratio scaling as
+/// conservative for every type — non-bulk measured/model 1.04–1.19,
+/// stack 12.6/20.8/34.3 vs this rule's 12/16/32. The swings×hand
+/// decomposition used on belt-DROP sides is ~12% optimistic for machine
+/// feeds and is deliberately not used here. At L0 the ratio is 1 for
+/// every type — bit-identical to the flat constant.
+pub fn machine_feed_rate(name: &str, quality: QualityTier, level: u8) -> f64 {
```
storkme, line 589: RFC-049 kill criterion 2 (`docs/rfc-049-inserter-capacity-research.md` line 146-154) is explicit: "every input-side constant at level > 0 cites a source reproduced across ≥2 independent raw extractions, or a human-verified in-game measurement ... If that bar can't be met for a level, that level's input-side scaling ships as L0 (conservative floor) with the gap documented — never derived."

The calibration matrix here measured 5 types × L{0,2,7} — 3 of 8 levels per type. L1 and L3–L6 have neither a measurement nor an L0 fallback; they're covered by this same `hand(L)/hand(0)` formula, i.e. derived from the measured points, which is exactly what the criterion rules out for a level that hasn't itself cleared the bar.

This isn't just a process nitpick: the three measured margins for stack-inserter aren't monotonic (measured/model per the decision log: 1.05× at L0, 1.30× at L2, 1.07× at L7). A curve that dips at both measured endpoints and bulges in the middle gives no basis for assuming L1/L3–L6 stay ≥1.0× — one of them could plausibly fall below the model line, which would mean this check silently under-warns (fails open) for a genuinely inserter-bound machine at that level. That's the exact failure mode this rating exists to catch. Given kill 2's own fallback text, should L1/L3-L6 ship at the L0 floor until they're individually measured, rather than interpolated from the 3 sampled points?

crates/sim-harness/src/scenario.rs
```diff
@@ -424,6 +425,21 @@ script.on_init(function()
...
+  -- out-provisions that assumption, masking genuine L-level shortfalls
+  -- (and making inserter-throughput warnings untestable — #352). Set
+  -- the force bonuses directly to the level's values — the two fields
+  -- reproduce all three I8b hand tables exactly (probe-verified
+  -- 2026-07-22): non-bulk hand = 1 + stack_size_bonus, bulk hand =
+  -- 1 + bulk_bonus, stack-inserter hand = 5 + bulk_bonus. Direct
+  -- assignment beats un-researching capacity techs, which left
+  -- non-bulk one step high (an unidentified tech grants +1). The
+  -- realized bonuses are dumped into the result for verification.
+  local NB_BONUS = {0, 0, 1, 1, 1, 1, 1, 3}
```
storkme, line 439: This is the load-bearing part of #370 — get `NB_BONUS`/`BULK_BONUS` or the `INSERTER_CAPACITY + 1` indexing wrong and every calibration number downstream (including `machine_feed_rate`'s conservative-credit claim in `common.rs`) gets measured against the wrong tech state, silently. Nothing in this file's test module exercises it: no test asserts `build_control_lua` emits the right `INSERTER_CAPACITY` for a given `manifest.inserter_capacity`, or that these two arrays agree with `inserter_hand`'s `NON_BULK`/`BULK`/`STACK` tables in `common.rs` (non-bulk hand − 1, bulk hand − 1, stack hand − 5). They're also now a third hand-maintained copy of that schedule (the two Rust tables plus this Lua literal) with nothing to catch drift if the schedule is ever revised — kill 4's in-game anchor is explicitly still open, so a revision is plausible. Worth a test pinning the generated Lua for a couple of capacity levels.

crates/sim-harness/src/scenario.rs
```diff
@@ -638,7 +654,11 @@ local function finalize(s, converged)
     proxies_fulfilled = storage.proxies_fulfilled,
     samples = storage.samples, checkpoints = storage.checkpoints,
     machine_census = census, converged = storage.converged, final_tick = game.tick,
-    fluid_errors = storage.fluid_errors, kit_errors = storage.kit_errors}, false)
+    fluid_errors = storage.fluid_errors, kit_errors = storage.kit_errors,
+    -- Realized capacity bonuses after tech-state parity (#370) — the
```
storkme, line 670: The comment frames this as "the verification channel that the tech rollback actually took effect," but tracing where `result` goes: `Report::compute` in `report.rs` builds the persisted `Report` from this same JSON and never reads `inserter_stack_size_bonus`/`bulk_inserter_capacity_bonus` — they reach neither `print_human` nor the `report` object `baseline_from_report` consumes. They only survive at all when `--out` is passed (buried under `raw_result` in `main.rs`), and nothing compares them against the value expected for the manifest's `inserter_capacity`. So today this is a value someone happened to eyeball once during the 2026-07-22 probe, not an ongoing check — a future regression (a mod update changing force-bonus semantics, a refactor of `finalize`) wouldn't be caught by anything that runs automatically. If a durable verification channel is the intent, it needs an actual comparison somewhere (e.g. `report.rs` asserting against the level's expected bonus), not just a field that's present in the raw JSON for whoever remembers to look.

crates/sim-harness/src/baseline.rs
```diff
@@ -67,11 +67,15 @@ pub fn baseline_from_report(report: &serde_json::Value) -> Result<Baseline, Stri
     if produced.is_empty() {
         return Err("report has no measured rates to bless".into());
     }
+    // Tech-state parity (#370): inserter-capacity research is rolled
+    // back to the fixture's level, so the level is part of the world
+    // the rates were measured in — key the baseline on it.
+    let capacity = r.get("inserter_capacity").and_then(|v| v.as_u64()).unwrap_or(0);
```
storkme, line 77: Bless/check symmetry itself is fine — both paths run through `baseline_from_report`, so they can't derive the key two different ways. But there's no test exercising the exact scenario #370 exists to prevent: bless at one `inserter_capacity` level, check a report from a different level, confirm it's flagged as tech-state drift instead of silently compared. Also worth pinning: `unwrap_or(0)` means a report that's missing the field entirely (e.g. an older report format, or a bug upstream that drops it) silently keys as L0 rather than erroring — a test asserting the field survives manifest → `report.rs` → here end-to-end would catch both.

crates/core/src/common.rs
storkme: Conceded — and you were more right than the objection even claimed. Completing the matrix (all 8 levels for stack and bulk, 10 further cells, kit-audit clean) shows stack's measured machine-feed curve is **non-monotone in hand size**: 12.62, 12.38, 20.75, 20.75, 23.12, 24.00, 23.88, 34.25 across L0–L7 — dips at hands 7 and 14 (swing-cycle quantization). Hand-ratio over-credited exactly the two cells you flagged as unproven: L1 (14.0 vs 12.38 measured) and L6 (28.0 vs 23.88), with L5 at zero margin. Fixed: stack now uses a measured floor table [12.0, 12.0, 19.2, 19.2, 21.6, 22.8, 22.8, 32.0] — monotone, 3–8% under every measured cell, flat-12.0 bit-identity at L0. Bulk measured safe under hand-ratio at all 8 levels (1.93–2.27×); non-bulk hand values {1,2,4} were fully covered by the original cells (the formula keys on hand value, not level). Test now pins credit ≤ measured at every level for both types plus monotonicity. Full table in the RFC-049 decision log amendment.

crates/sim-harness/src/scenario.rs
storkme: Landed: `parity_bonus_tables_and_level_are_emitted` pins both Lua tables byte-for-byte plus the INSERTER_CAPACITY emission and the self-audit string. On the third-copy concern: the harness deliberately doesn't depend on spaghettio_core (manifest-JSON-schema-only boundary), so a pinned-constants test is the available drift guard — the test comment directs any I8b table change to update both copies and re-run the calibration matrix.

crates/sim-harness/src/scenario.rs
storkme: Right — a verification channel nobody checks isn't one. Fixed structurally: the scenario now reads the force fields back immediately after assignment and appends to `kit_errors` on mismatch (same pattern as the chest-overlap audit), which forces the verdict to NO DATA — a failed parity assignment can no longer produce numbers. The realized bonuses are also now first-class `Report` fields printed on every run (`layout: ... (realized bonuses: nb=N bulk=M)`), not just raw_result residue.

crates/sim-harness/src/baseline.rs
storkme: Test added: bless at L0, check a L7 report — asserts exactly one loud 'tech state' drift. The `unwrap_or(0)` default is now commented as deliberate: a report missing the field (pre-parity or foreign) keys as L0 and therefore mismatches any deliberately-leveled baseline loudly in check_against, which is the failure mode we want for stale artifacts.

PR #389, opened 2026-07-23
feat(analysis): blueprint strategy classification (classify) + 2.0 quality-only-filter parse fix

storkme:
## Intent

Strategy-feature extraction for parsed blueprints — the Rust port of the community-corpus mining work (strategy-gap map, multi-strategy exploration). `classify()` answers "HOW is this blueprint built" (archetype, chain level, direct insertion, network structure, periodicity) as a sibling to `analyze()`'s "what does it make", reusing the shared engine machinery instead of a parallel implementation. Features attach to `BlueprintAnalysis.features`, so `blueprint-analyze --json` emits them for free.

## Scope

- `crates/core/src/classify.rs` (new, 742 lines): census, direct-insertion count (engine drop-side convention via the shared import un-flip, `blueprint.rs`'s #348 flip), belt/pipe/pole network segmentation, sideloads, UG pairs, sushi inference, periodicity/tiling score, power self-sufficiency (validator ground-truth `supply_area_distance`/`pole_wire_reach`), archetype + chain-level decision lists. 17 unit tests incl. a blueprint round-trip through the real parser.
- `analysis.rs`: `features` field, computed in `analyze()`.
- `mining-cli`: strategy + power summary lines in human output.
- `blueprint_parser.rs`: Factorio 2.0 **quality-only filter conditions** (`{"index":1,"quality":"legendary","comparator":"="}` — no item `name`) no longer break the parse; dropped on collapse + regression test. Found via the corpus: broke 16 of 172 top-favorites prints, including every large modern book.
- `docs/file-reference.md`: classify row.

No production layout paths touched; analysis/classify are read-only over parsed layouts.

## Verification actually run

- `cargo test --manifest-path crates/core/Cargo.toml` — full suite green: 800 lib (incl. 17 new classify tests + 1 parser regression test), 60/60 e2e, all integration targets, 0 failures. Run with `SPAGHETTIO_ZONE_CACHE_PATH` pinned to the committed cache; the pin file was restored afterwards (local runs grow it via read-through-write; the two e2e "failures" seen before pinning were the known unpinned-cache flake).
- `cargo clippy --workspace` clean.
- Corpus cross-validation: all 172 surviving 2.0-era Factorio-Prints top-favorites (6,087 book members) swept through the built analyzer; strategy totals match the independent Node reference implementation within heuristic tolerance (DI 16,623 vs 16,379; tileable 1,632 vs 1,593; mixed-belt 753 vs 739). 3 residual errors are planner-only books (zero blueprints — correct behavior). Spot-checked the early-game green-circuit block: `production-block / intermediates, DI:8, tileable p3, self-powered` ✓.

## Deviations from agreed approach

- Built binaries linked via a hermetic /tmp toolchain (bootlin gcc + bookworm glibc-2.36 sysroot from Debian packages) — the review box has rustup but no `cc`/libc-dev; nothing system-wide was installed or changed.
- The corpus itself, the fetch/sweep/deep-dive `.mjs` scripts, and derived JSONL stay local per the rights decision (corpus dir is gitignored).

## Models / contracts touched

`docs/factorio-mechanics.md` — new P4a entry (big-electric-pole: 2×2 footprint, 4×4 supply area, wiki-verified quality scaling), added with the 801b4120 supply-model correction; the doc update shipped in the same commit per the contract-change rule. `common.rs` additions (`supply_area_distance` arms, `entity_size` big-pole) verified additive for all production callers (see review thread). No other contracts touched.


storkme (commented):
Read the whole `classify.rs`, `blueprint_parser.rs` change, the mining-cli hookup, and the surrounding `common.rs` / `validate/power.rs` context. The change is genuinely additive (nothing under `bus/`, `solver`, `netflow`, `validate/`), the parser fix is well-scoped with a good regression test, and the direct-insertion convention lines up correctly with `blueprint.rs`'s import un-flip — the round-trip test proves it end-to-end.

**Confirmed findings** (heuristic-imprecision and one census-bug the corpus sweep would silently mask):

1. Census `logistic_chests` never matches a Factorio 2.0 blueprint — 2.0 renamed the entities (`active-provider-chest`, `passive-provider-chest`, `requester-chest`, `buffer-chest`, `storage-chest`), and `recipes.json` already uses the new names. The prefix filter always reports 0 for the target corpus.
2. Power self-sufficiency has three real divergences from `validate::power::check_power_coverage` despite the PR body's "validator ground-truth" claim (small/big poles counted as supply sources with a wrong 3.5 supply distance inherited from `common::supply_area_distance`; inserters excluded from the subject set; burner biochambers included). Details inline.
3. Sideload heuristic double-counts corners as sideloads — a plain east-then-south L-turn scores identically to a genuine perpendicular T-sideload.
4. Test `direct_insertion_reversed_inserter_is_not_counted` asserts `direct_insertion == 1` (i.e., IS counted) — the internal comment even acknowledges it. Rename it.
5. Inserter reach is hardcoded in two spots (`e.name == "long-handed-inserter"`) while `common::inserter_reach` already exists — one more silent-drift risk when new inserter tiers appear.

None of these are blockers — the module is heuristic and read-only over parsed layouts, and the PR body correctly frames it that way — but #1 is a straight census bug and #2 makes the power number quantitatively different from the validator it cites, which is worth fixing before the strategy-gap map is built on top of these numbers.

**Verification I ran** (PR head `a3f4b694`, `SPAGHETTIO_ZONE_CACHE_PATH` pointed at the committed `crates/core/data/sat-zones.bin`):
- Lib: **800 passed, 0 failed, 3 ignored** — matches the PR body exactly. New classify + parser tests all green.
- E2E: **58 passed, 2 failed, 22 ignored** — `partition_strategy_scoreboard` and `tier4_advanced_circuit_from_ore_am2`. This diverges from the PR body's "60/60". The failures are in tests unrelated to any code this PR touches (partition scoreboard + AC-from-ore), so I don't think the PR caused them, but the "60/60 with pinning" claim didn't reproduce here — worth double-checking whether the pin file (`sat-zones.bin` vs `sat-zones-ci.bin`) is the one you were using.

**Unverifiable** (out of scope for this checkout):
- The 172-book / 6,087-member corpus cross-validation numbers (DI 16,623 vs 16,379; tileable 1,632 vs 1,593; mixed-belt 753 vs 739). The corpus and Node reference impl are local/gitignored, so I can't rerun them. Taking on trust; noting for the record.
- The "3 residual errors are planner-only books" claim, same reason.

crates/core/src/classify.rs
```diff
@@ -0,0 +1,742 @@
...
+        splitters: ents.iter().filter(|e| common::is_splitter(e.name.as_str())).count(),
+        inserters: ents.iter().filter(|e| common::is_inserter(e.name.as_str())).count(),
+        pipes: ents.iter().filter(|e| is_pipe_entity(e.name.as_str())).count(),
+        poles: ents.iter().filter(|e| is_pole_entity(e.name.as_str())).count(),
+        beacons: ents.iter().filter(|e| e.name == "beacon").count(),
+        rails: ents.iter().filter(|e| is_rail(e.name.as_str())).count(),
+        train_stops: ents.iter().filter(|e| e.name == "train-stop").count(),
+        combinators: ents.iter().filter(|e| is_combinator(e.name.as_str())).count(),
+        roboports: ents.iter().filter(|e| e.name == "roboport").count(),
+        logistic_chests: ents.iter().filter(|e| e.name.starts_with("logistic-chest")).count(),
```
storkme: This filter always returns 0 for Factorio 2.0 blueprints. 2.0 renamed the logistic-chest family: they are now `active-provider-chest`, `passive-provider-chest`, `requester-chest`, `buffer-chest`, `storage-chest` (see `crates/core/data/recipes.json` for the canonical names). Nothing in a 2.0 blueprint starts with `logistic-chest-` — only legacy 1.x exports would match. Since the target corpus is 2.0-era Factorio-Prints, the `logistic_chests` census field is effectively dead for the actual use case. A suffix match on the modern names, or a small const set, matches what the analyzer will actually see.

crates/core/src/classify.rs
```diff
@@ -0,0 +1,742 @@
...
+    let poles: Vec<&PlacedEntity> = ents.iter().filter(|e| is_pole_entity(e.name.as_str())).collect();
+    let machines: Vec<&PlacedEntity> = ents.iter().filter(|e| is_machine(e)).collect();
+    if !machines.is_empty() {
+        let covered = machines
+            .iter()
+            .filter(|m| {
+                let (mx, my) = center(m);
+                poles.iter().any(|p| {
+                    let (px, py) = center(p);
+                    let sd = common::supply_area_distance(p.name.as_str(), QualityTier::Normal);
```
storkme: The PR body says this reuses "validator ground-truth `supply_area_distance`/`pole_wire_reach`", but the semantics diverge from `validate::power::check_power_coverage` in three ways:

1. **Pole set is broader here.** The validator only counts `medium-electric-pole` and `substation`. Here, `is_pole_entity` returns `true` for any name where `pole_wire_reach` is `Some`, which includes `small-electric-pole` and `big-electric-pole`.
2. **`supply_area_distance` returns 3.5 for anything except substation** (`_ => 3.5` in `common.rs`). That is right for medium poles but wrong for the two poles you just added to the set — small pole is 2.5 (5×5 area), and big pole has no supply area at all in vanilla 2.0 (it is a wire-only pole). So this over-counts coverage in blueprints wired through small or big poles.
3. **Subject set is different from the validator.** Validator uses `needs_electricity(e.name)` (electric machines including inserters, excluding the burner biochamber). Here, `is_machine(e)` skips inserters entirely and includes biochambers. `machines_powered_fraction` is therefore quantitatively different from what the validator would emit.

Also `QualityTier::Normal` is hardcoded — legendary poles in the parsed blueprint lose their +5 supply bonus. Not a blocker (`e.quality` is available), but worth using it if you want the "validator ground-truth" phrasing to hold up.

crates/core/src/classify.rs
```diff
@@ -0,0 +1,742 @@
...
+    for (&(tx, ty), &e) in &belt_at {
+        if !common::is_surface_belt(e.name.as_str()) {
+            continue;
+        }
+        let (dx, dy) = common::dir_to_vec(e.direction);
+        if let Some(&fwd) = belt_at.get(&(tx + dx, ty + dy)) {
+            if common::is_surface_belt(fwd.name.as_str()) {
+                let (fdx, fdy) = common::dir_to_vec(fwd.direction);
+                if (fdx, fdy) != (dx, dy) {
+                    f.sideloads += 1;
```
storkme, line 361: This counts any belt whose forward neighbor differs in direction as a sideload — but a plain L-corner (east belt at (0,0), south belt at (1,0)) matches the same pattern. In Factorio the corner is normal turn-flow, not a sideload; a true sideload requires the parent belt to continue in its own direction past the join. As-written, `sideloads` inflates by the corner count, which will skew `sideload_per_100_belts` on any blueprint that turns belts (i.e. most of them). Either rename the metric to something like "direction-changes" or gate it by "parent belt has a same-direction forward continuation past the junction".

crates/core/src/classify.rs
```diff
@@ -0,0 +1,742 @@
...
+            machine("assembling-machine-3", 0, 0, "copper-cable"),
+            machine("assembling-machine-3", 4, 0, "electronic-circuit"),
+            ent("fast-inserter", 3, 1, EntityDirection::East),
+        ]);
+        assert_eq!(f.direct_insertion, 1, "east-facing inserter between machines should be DI");
+        assert_eq!(f.archetype, "production-block");
+    }
+
+    #[test]
+    fn direct_insertion_reversed_inserter_is_not_counted() {
```
storkme: Test name says `is_not_counted` but the assertion is `direct_insertion == 1` (i.e., IS counted), and the inline comment acknowledges the wiring is still counted as DI. Rename to something like `_still_counted` / `_is_direction_agnostic`; as-is a future reader hunting through test names will misread what the function does.

crates/core/src/classify.rs
```diff
@@ -0,0 +1,742 @@
...
+        "270+/s".to_string()
+    };
+
+    // ---- direct insertion (engine convention: direction = drop side) ----
+    for e in ents {
+        if !common::is_inserter(e.name.as_str()) {
+            continue;
+        }
+        let (dx, dy) = common::dir_to_vec(e.direction);
+        let reach = if e.name == "long-handed-inserter" { 2 } else { 1 };
```
storkme: Reach is hardcoded here (and again at line 376). `common::inserter_reach(name)` already exists as the single-source table — using it means a future new inserter tier with reach ≠ 1 does not silently mis-count DI/sushi. Cheap swap.

crates/core/src/classify.rs
storkme: Fixed in b153c80c: `logistic_chests` now matches the 2.0 family explicitly (`active-provider-chest`, `passive-provider-chest`, `requester-chest`, `buffer-chest`, `storage-chest`) via a new `is_logistic_chest`. Legacy 1.x spellings deliberately not matched — pinned by `logistic_chests_use_2_0_names`.

crates/core/src/classify.rs
storkme: Fixed in b153c80c, all three divergences plus quality:

1. **Subject set** now mirrors the validator exactly: `needs_electricity` (electric machines + inserters, excluding the burner biochamber). Pinned by `inserters_are_coverage_subjects_biochambers_are_not` (uncovered inserter halves the fraction; biochamber-only → `None`).
2. **Supply distances**: `common::supply_area_distance` gained explicit arms — small pole 2.5 (5×5), big pole 0.0 at every tier (wire-only; quality is not added, since a wire-only pole has no supply area to scale and the +1/level rule was kill-1-verified only for medium/substation). Sources with 0 supply are skipped. I deliberately kept the broader-than-validator source *set* (small poles are real supply sources in community blueprints — excluding them would under-count the corpus), but now with correct per-tier geometry. `small_pole_supplies_its_own_smaller_area` / `big_pole_is_wire_only_and_supplies_nothing` pin the boundary.
3. **Quality**: both supply distance and wire reach now use each pole's parsed `quality` (`unwrap_or_default`), same as the validator.

The PR body's "validator ground-truth" phrasing now holds: same subjects, same per-pole distances, same quality rule — the only extension is small/big poles, which the validator never sees because the engine never places them.

crates/core/src/classify.rs
storkme: Fixed in b153c80c: a direction change at the forward tile now only counts as a sideload when the belt there has a same-direction **upstream** continuation behind the join — i.e. it sits on a through line. (Upstream, not forward: a turned L-corner belt also continues in its new direction *past* the turn, so a forward gate can't separate the two — the discriminator is that the parent line existed before the junction.) `sideload_detected` gained the upstream belt it was missing; new `l_corner_is_not_a_sideload` pins the old 3-belt fixture at 0.

crates/core/src/classify.rs
storkme: Renamed to `direct_insertion_reversed_inserter_still_counted` in b153c80c.

crates/core/src/classify.rs
storkme: Swapped both sites (DI at :286, sushi at :376) to `common::inserter_reach` in b153c80c.

storkme (commented):
Delta review of `b153c80c`. All five original findings resolved; the `common.rs` change is verified additive; the new pins are on the right invariants.

**Resolution vs. original findings:**
1. `logistic_chests` — **fixed**. `is_logistic_chest` matches the exact 2.0 set, legacy names explicitly excluded, pinned by `logistic_chests_use_2_0_names` (includes a rejection for the 1.x name).
2. Power self-sufficiency — **fixed**. Subjects are now `needs_electricity` (matches validator); per-pole supply distances now correct via new arms in `common::supply_area_distance`; each pole's parsed `quality` is now threaded through both supply and wire-reach calls. The three pins (`inserters_are_coverage_subjects_biochambers_are_not`, `small_pole_supplies_its_own_smaller_area`, `big_pole_is_wire_only_and_supplies_nothing`) lock the exact boundaries. One modeling caveat left in-thread about the big-pole = 0 choice.
3. Sideload heuristic — **fixed**. Upstream-continuation gate discriminates T-sideloads from L-corners; L-corner test pins the negative case, T-sideload test was updated to include the parent line it needed. One small false-negative class left in-thread (perpendicular feed at the very head of a line).
4. Test rename — **fixed** (`_still_counted`).
5. Inserter reach → `common::inserter_reach` — **fixed** at both sites.

**New risk (common.rs): verified additive.** Grepped every non-test caller of `supply_area_distance`. `validate::power::check_power_coverage` at `power.rs:78` receives `&e.name` but the entities are filtered to `medium-electric-pole || substation` two lines above at `power.rs:72`; every other production call site (`bus/layout.rs`, `validate/power.rs`) passes a string literal that is one of those two. The `unknown-pole` default (3.5) is preserved and its regression test still passes. The two new arms fire only for names no production caller ever passes, so behavior for existing callers is byte-identical. The PR's "no production layout paths touched" claim holds.

**Minor drift note (not blocking):** `BlueprintFeatures.machines_powered_fraction` is now populated from `consumers` (includes inserters, excludes biochambers) rather than the machine set the field name implies. The mining-cli formatting (`main.rs:275-278`) still prints "machines covered", which now means "electric consumers covered". A rename would ripple through the public field and CLI output; leaving as-is is defensible, but the name has quietly drifted from what it measures.

**Tests I ran** (PR head `b153c80c`, fresh worktree, `SPAGHETTIO_ZONE_CACHE_PATH=crates/core/data/sat-zones.bin`):
- Lib: **805 passed, 0 failed, 3 ignored** — up from 800 by exactly the 5 new pins (`l_corner_is_not_a_sideload`, `logistic_chests_use_2_0_names`, `small_pole_supplies_its_own_smaller_area`, `big_pole_is_wire_only_and_supplies_nothing`, `inserters_are_coverage_subjects_biochambers_are_not`); the rename left count unchanged.
- E2E not re-run — `common.rs` change is provably no-op for existing callers, and the previous run's two failures (`partition_strategy_scoreboard`, `tier4_advanced_circuit_from_ore_am2`) are in tests this PR does not touch.

Happy with this. No further blockers.

crates/core/src/classify.rs
storkme: One follow-up on the big-pole = 0 choice, for the record. Factorio 2.0's vanilla big-electric-pole actually does have a supply area (`supply_area_distance = 2.0` in prototypes, → 4×4 supply beyond its 2×2 footprint). The commit's "wire-only" characterisation is a conservative modeling choice, not the game's literal rule. That's a fine call for corpus classification (big poles overwhelmingly do carry long-range wire between grids, not power adjacent machines), and it errs toward under-counting rather than over-counting, so it won't inflate self-powered rates. Just flagging that the docstring wording ("wire-only pole has no supply area to scale") could mislead a future reader who cross-checks against the game. The subject-set / small-pole / quality fixes here are all textbook — no complaint on those.

crates/core/src/classify.rs
storkme: Gate reasoning checks out — the upstream-continuation discriminator is the right one. Small false-negative class left: a perpendicular feed into the very first tile of a through-line (no belt behind the join) won't count, since there's no upstream continuation to check. In practice this is rare enough not to matter for corpus classification, and it's a clean tradeoff against the L-corner false-positive class you just closed. Worth noting in a comment somewhere if a future reader wonders why one specific T-shape scores 0.

crates/core/src/classify.rs
storkme: Checked against the official wiki before touching anything — you're right, and it's slightly worse than flagged: the big pole is not just supplied (4×4, `supply_area_distance` 2.0), it's also **2×2**, which meant `entity_size` had it wrong too (center off by half a tile). Quality scaling on the wiki (4×4 → 6/8/10/14×14) matches the +1/level rule exactly, so no special casing needed there.

Fixed in 801b4120 by putting the game's literal values in `common::supply_area_distance` (2.0 + level) rather than keeping 0 with a doc caveat — ground truth shouldn't carry a consumer-specific conservative model. `entity_size` gained the (2,2) arm (inert for production paths, exact for imports), the `sd <= 0` source skip in classify is gone (dead), the docstring now cites the verified values, and `big_pole_supplies_its_tiny_area` pins both sides of the 2.0 boundary. Also added rule P4a to `docs/factorio-mechanics.md` (and corrected P4's "only the substation is 2x2").

crates/core/src/classify.rs
storkme: Noted in the gate comment in 801b4120 — head-of-line perpendicular feed scores 0, recorded as the accepted tradeoff against the L-corner false-positive class.

storkme (commented):
Delta-2 review of `b153c80c..801b4120`.

**Big-pole commit (`801b4120`) — clean and correct.** Verified the numbers against Chebyshev-distance-on-tile-centers geometry (a 2×2 pole at center `(1,1)` with sd=2.0 covers tile-centers `t+0.5 ∈ [-1.5, 2.5]` → `t ∈ {-1,0,1,2}` = 4 tiles per axis = 4×4 area; scaling by +1/level gives 4/6/8/10/14 across `QualityTier::level` 0/1/2/3/5, matching the doc). The `entity_size` change from `(1,1)` to `(2,2)` for `big-electric-pole` is a strict correctness fix: I audited every non-test caller of `entity_size` and none receive big-pole names on engine-emitted layouts (validator filters pre-exclude, engine never places big poles). The only paths that get the new value are those operating on IMPORTED community blueprints (belt-structural overlap check, `power_wires::pole_center`, exporter fallback), and there the new value matches what Factorio actually places. `docs/factorio-mechanics.md` P4a numbers all check out — 4×4 base, 6/8/10/14×14 across tiers, wire 32→42. No complaint.

**Bot finding (`tiles_of()` double-rotates E/W splitters) — confirmed real.** Wrote a scratch harness in a fresh worktree of `801b4120`. Fixture: east-facing splitter at `(2,0)` with a probe belt at `(2,2)`. Correct game footprint `(2,0)+(2,1)` merges probe into one network via the vertical tile; the buggy horizontal footprint `(2,0)+(3,0)` leaves `(2,1)` empty and orphans the probe.

```
test east_splitter_second_tile_is_at_y_plus_one_not_x_plus_one ... FAILED
  left: 2   (observed — probe orphaned)
  right: 1  (correct — single network)
test north_splitter_baseline_still_bridges ... ok  (N/S orientation unaffected)
```

The existing `balancer_archetype` test's east-facing splitter doesn't catch it because it only asserts `archetype` and `chain_level` — neither reads the DSU. The bug misses precisely the corpus-facing features (`belt_networks`, `largest_belt_network`, `sideloads`, `ug_pairs`, `mixed_belt_networks`) for any blueprint containing an E/W splitter, which is most non-trivial ones. Fix is what the bot suggested — skip the swap when `is_splitter(name)` is true, since `oriented_splitter_dims` already returns rotated dims. I'd add a regression test parallel to the two I wrote above so this doesn't come back.

Merging as-is means the corpus mixed-belt / belt-network numbers cited in the PR body are quantitatively off. Not a merge blocker if the intent is "land the scaffolding, iterate on accuracy" — but if the strategy-gap map depends on those specific counts, worth fixing before it's built on top.

**PR-template ack.** The bot is right that the description omits the "Models / contracts touched" section, and this PR now modifies `docs/factorio-mechanics.md` (the exact file the template names as an example of what belongs there). Worth a one-line addition to the body — the wiki-verified big-pole P4a is a genuine model correction with corpus-wide effect on our own analysis output, not a doc-noise change.

**Tests I ran** (PR head `801b4120`, fresh worktree, `SPAGHETTIO_ZONE_CACHE_PATH=crates/core/data/sat-zones.bin`):
- Lib: **805 passed, 0 failed, 3 ignored** — same count as `b153c80c`, since the big-pole commit modified one existing test rather than adding new ones.
- Scratch adversarial test for the double-rotation bug: 1 failed (E/W), 1 passed (N/S baseline) — evidence in the confirmation above.
- E2E not re-run — the `common.rs` `entity_size` and `supply_area_distance` changes are still additive for every production call site (validated in the delta-1 review + re-audited here for the `entity_size` big-pole arm).

crates/core/src/classify.rs
```diff
@@ -0,0 +1,864 @@
...
+        common::oriented_splitter_dims(e.name.as_str(), e.direction)
+            .unwrap_or((2, 1))
+    } else {
+        common::entity_size(e.name.as_str())
+    };
+    // rotate non-square footprints for E/W orientation
+    if matches!(e.direction, EntityDirection::East | EntityDirection::West) && w != h {
+        std::mem::swap(&mut w, &mut h);
+    }
+    let mut out = Vec::with_capacity((w * h) as usize);
```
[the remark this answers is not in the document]
storkme: Confirmed empirically. Wrote a scratch integration test in a fresh worktree of `801b4120` — east-facing splitter at `(2,0)` with a probe belt at `(2,2)`. Under the correct footprint `(2,0)+(2,1)` the probe merges into one network via the vertical tile; observed `belt_networks == 2` (probe orphaned), which is exactly what the phantom horizontal `(3,0)` tile produces. A north-facing splitter with the same shape passes as baseline, isolating the failure to E/W.

Blast radius on the corpus-facing features: `belt_networks`, `largest_belt_network`, `sideloads` (iterates `belt_at.keys()`), `ug_pairs` (uses `belt_at.get` on the output tile), and `mixed_belt_networks` (derived via `net_of` from the belt groups) all silently take the wrong second tile for every E/W splitter. Splitter COUNT is fine — census reads `is_splitter(name)` directly, not via `tiles_of`. The suggested one-line fix (skip the swap for splitters, since `oriented_splitter_dims` already returns rotated dims) is what I'd take; recommend pinning it with a parallel unit test so this can't come back.

Issue #385, opened 2026-07-22
belt_drop_rate over-credits researched hands onto belts — RFC-049 Phase 1's L>0 output credits falsified by sim


storkme:
Found by #381's verification gate (the PR itself is exonerated — its input-side change measures non-binding). Mirror micro-calibration (machine→belt drop onto yellow, parity harness):

| out-inserter | credited (swings×hand) | measured |
|---|---|---|
| fast @ L0 (hand 1) | 2.31 | 2.55 ✓ |
| fast @ L7 (hand 4) | **9.24** | **6.00 (−35%)** |
| stack @ L0 (hand 6) | 14.4 | ≥7.5 (sink-capped) |

In-row it compounds: shared-lane gap scarcity takes fast@L7 from 6.0 solo to ~2.5/s effective (ec10-L7 fixture: cable rows halve, whole factory −50%, zero warnings — a clean-but-failing case produced by our own wiki-derived constants). Monotone corroboration across full runs: out-hand 16 → 30/s cable, hand 6 → 22.4, hand 4 → 14.8.

**Scope**: `common::belt_drop_rate`'s swings×hand model (RFC-049 Phase 1, level-aware belt-drop sizing + validator credits) is optimistic at L>0. L0 flat constants are fine (2.31 vs 2.55 measured). Same shape as the input-side story kill 2 anticipated: the wiki tables don't transfer; only measurement does.

**Fix shape**: measured belt-drop table (Phase-2-style matrix, but row-context — shared-lane drops, realistic drain), then re-derive both the ladder's output sizing and the validator's belt-drop credits from it. #381 stays blocked until its verification fixture passes on the corrected constants.

Evidence: session scratchpad calib/report-outdrop-*.json, p3diff L0/L7 parity reports, per-segment divergence table on #381.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #402, opened 2026-07-23
#385 second half: row output lane-budget check (sim-calibrated, bridge-aware)

storkme:
## Intent
Completes #385: a row whose output exceeds what inserter drops onto its belt-out can physically realize (LANE_UTILIZATION × lane rate × S × lanes-loaded, bridge-aware) now warns at generation time — closing the clean-but-failing class the sim exposed on ec10 (5.00/10 with a silent validator).

## Scope
- `check_row_output_lane_budget` + dispatch: physical rows resolved by tile-adjacency clustering (segment tags are recipe-keyed, not row-keyed; `effective_rows` is unreliable and absent in the cell pipeline — both verified against real fixtures), machines attributed to rows via their output inserter's drop tile.
- 14 unit tests; 6 fixture expectation updates, each documented inline and sim-adjudicated (configs proven at-plan stay silent; configs the game fails now warn).

## Verification
- Acceptance matrix (empirical): fires on ec10 S1 L0 **and** L7 — on **copper-plate** (24 machines, 15/s vs 12.75 realizable even bridged), which matches the sim better than the issue's original cable guess (predicted floor 12.75 vs 11.28 measured); silent on gear10/automation/logistic/military (all sim-proven at plan).
- Full suite: 811 lib + 60 e2e + 10 netflow + 10 cell-composition green; clippy workspace-clean; WASM clean.
- One tolerated PLAUSIBLE: `cell_candidate_resolves_ec15_refusal` gains a warning on a geometry with no sim proof either way — reported, not tuned away.

## Models / contracts touched
New check name `row-output-lane-budget` (warning severity). No signatures, no serialized models.

Validator-semantics change: local adversarial review dispatched alongside the bot per house rules.

## Deviations

Two adversarial rounds reshaped the constants mid-review, deliberately documented rather than squashed: the original ×2 bridge multiplier (unmeasured) was challenged; the interim ×1.5 correction was itself falsified (derived from a consumption-limited cell); the shipped 1.733 comes from a dedicated ceiling cell (cable@13 uncommon, delivered at plan in-game). The four e2e expectation changes from the interim round reverted on their own under the measured constant. Spin-offs filed rather than scope-crept: #404 (lane-throughput error-severity false positives on bridged rows) and a first measured cell on #311 (merger over-commit).

### DIFF
--- a/crates/core/src/validate/inserters.rs
+++ b/crates/core/src/validate/inserters.rs
@@ -12,7 +12,7 @@ use crate::common::{
 };
 #[cfg(test)]
 use crate::common::inserter_throughput;
-use crate::models::{LayoutResult, PlacedEntity, SolverResult};
+use crate::models::{EntityDirection, LayoutResult, MachineSpec, PlacedEntity, SolverResult};
 
 /// RFC-046: throughput an inserter delivers when dropping onto a **belt**.
 /// At the layout's belt stack size S > 1, a stack inserter's belt drops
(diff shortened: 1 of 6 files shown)


crates/core/src/validate/inserters.rs
```diff
@@ -588,6 +588,342 @@ pub fn check_inserter_item_throughput(
...
 
+// ── check_row_output_lane_budget ─────────────────────────────────────────────
+
+/// #385's second half: a machine row's belt-out realizes at most
+/// `LANE_UTILIZATION × lane_capacity_stacked(tier, stacking)` **per lane
+/// actually loaded** — inserter drops fill only the far lane (I5), so a
+/// row whose belt-out has no midpoint sideload bridge realizes ONE lane
+/// regardless of how many inserters/machines feed it, no matter the
+/// inserter type, count, or research level (sim-measured 2026-07-23,
+/// `docs/sim-harness-forensics.md`: 7.4/s realized on yellow at S=1 vs
```
storkme: "Regardless of ... inserter type" is falsified by this same investigation's own calibration data. `report-row-stack-inserter-L0.json` / `-L7.json` (3 unmoduled copper-cable machines, 15/s combined demand, feeding one UNBRIDGED belt-out line — identical geometry to `row-fast-inserter-L0/L7`, only `ins_type` varies) measure **15.0/s and 15.2/s delivered**, at/above full both-lane nominal — not the ~7.4–8.0/s the fast-inserter runs show. Stack inserters escape the one-lane cap this check assumes universally.

This isn't hypothetical for this PR: in `stress_electronic_circuit_60s_red_from_ore` and `tier2_electronic_circuit_20s_from_ore`, the copper-cable rows this check newly flags are fed by `stack-inserter` output (confirmed by decoding each fixture's snapshot and tracing every output inserter's drop tile back to the belt-out cluster) — exactly the case the sweep shows doesn't need a bridge to hit full throughput. That's 6 of the 11 new warnings in the stress fixture and 2 of the 3 in the 20s fixture.

Separately: the citation doesn't hold up. `docs/sim-harness-forensics.md` is 104 lines; it has no `7.4`, no per-inserter-type table, nothing about rows or bridges. The only place `7.4` appears anywhere in the repo is in this PR's own new prose (here and the two decision-log entries) — the actual measurement lives in an uncommitted `/tmp` scratchpad JSON, not in anything a future reader can go check.

crates/core/src/validate/inserters.rs
```diff
@@ -588,6 +588,342 @@ pub fn check_inserter_item_throughput(
...
+                .filter(|e| crate::common::is_surface_belt(&e.name))
+                .collect();
+            if belts.is_empty() {
+                continue;
+            }
+
+            let tier = row_belt_tier(&belts);
+            let lanes_loaded = row_lanes_loaded(&belts);
+
+            let realizable = crate::common::LANE_UTILIZATION
```
storkme: Following from the doc comment above: `lanes_loaded` only reads belt-out geometry (bridge y-spread) — it never looks at which inserter type feeds the row. So a stack-inserter row and a fast-inserter row with the same bridge get an identical `realizable` ceiling, even though the only data available (the sweep referenced above) shows a ~2x gap between them on the *unbridged* case.

More importantly, I couldn't find a single bridged-row sim measurement anywhere (committed or in the scratchpad) for any inserter type — every calibration run available tests an unbridged single line. Every warning this check fires in the three fixtures I decoded (ec10, ec20, stress60) is on a row the check itself detects as bridged, so the "bridge doubles the ceiling to two lanes" half of the formula — the half every single fired warning actually depends on — is asserted from geometry, not measured. #385's own decision log calls out exactly this failure mode as the reason `belt_drop_rate` got rebuilt ("never checked against the belt's own physical throughput... an order-of-magnitude over-credit"); I'd want the same bar applied here — a bridged calibration run (any inserter type) — before trusting the ×2 multiplier, plus either a per-inserter-type distinction or evidence it doesn't matter once a bridge is present.

storkme (commented):
Verified against the sim ground truth: adjacency clustering holds on a real multi-row fixture (`stress_electronic_circuit_60s_red_from_ore` decodes to 6 separate, correctly-unmerged copper-cable physical rows, not one pooled bucket), the rate math is exactly right (ec10's copper-plate row fires at 15.00 vs 12.75 as claimed), severity is warning-only and doesn't touch any refusal/error path, the tolerated `cell_candidate_resolves_ec15_refusal` change is warning-only and pinned to exactly 1, and the suite/clippy counts match (811 lib / 60 e2e / 10 netflow / 10 cell-composition, clippy clean on this file).

The blocker is the check's central physical claim. Both inline comments below point at the same problem: the "no bridge ⇒ one lane, *regardless of inserter type*" model isn't just unproven for stack inserters, it's contradicted by data sitting in this same investigation's own scratchpad, and it's applied to real fixtures where that exact inserter type is the one feeding the flagged row. I'd want the check to at least account for inserter type (or a bridged-row calibration run confirming the doubling holds for stack inserters) before this ships — as written, I think a meaningful fraction of the new warnings (6 of 11 in the 60s stress fixture, 2 of 3 in the 20s fixture) are false positives on rows that actually deliver at plan in-game.

crates/core/src/validate/inserters.rs
storkme: Resolved by measurement (commit 20440289) — and the finding was even more right than argued: not only was the ×2 unmeasured, my interim ×1.5 correction was derived from a consumption-limited cell (ec10's plate row at 11.28 — a floor mistaken for a ceiling). A dedicated ceiling cell (engine-midpoint-bridged single row, cable@13 uncommon, yellow S1, parity world) delivers **13.00/s at plan in-game** → bridged factor = 13.0/7.5 ≈ 1.733 measured floor. On the stack-rows evidence: those 15.0/s cells were the pre-parity S=4-contaminated measurements — the true-S1 re-measures in the same scratchpad (report-row-stack-inserter-*-trueS1.json) show stack rows at 7.40, identical to fast, which is why the check stays type-blind under the declared-world doctrine. Citation fixed; the world-labeled table now lives in the RFC-047 decision log. Full suite green; the four e2e disputes resolved themselves under the measured constant.

crates/core/src/validate/inserters.rs
storkme: Resolved by measurement (commit 20440289) — and the finding was even more right than argued: not only was the ×2 unmeasured, my interim ×1.5 correction was derived from a consumption-limited cell (ec10's plate row at 11.28 — a floor mistaken for a ceiling). A dedicated ceiling cell (engine-midpoint-bridged single row, cable@13 uncommon, yellow S1, parity world) delivers **13.00/s at plan in-game** → bridged factor = 13.0/7.5 ≈ 1.733 measured floor. On the stack-rows evidence: those 15.0/s cells were the pre-parity S=4-contaminated measurements — the true-S1 re-measures in the same scratchpad (report-row-stack-inserter-*-trueS1.json) show stack rows at 7.40, identical to fast, which is why the check stays type-blind under the declared-world doctrine. Citation fixed; the world-labeled table now lives in the RFC-047 decision log. Full suite green; the four e2e disputes resolved themselves under the measured constant.

crates/core/src/validate/inserters.rs
```diff
@@ -588,6 +588,361 @@ pub fn check_inserter_item_throughput(
...
+            //   the 7.5 lane rate, any inserter type/level → 0.95/lane.
+            // - engine-midpoint-bridged single row (cable@13 uncommon,
+            //   yellow S1): delivers 13.00/s at plan in-game →
+            //   ≥ 1.733 lanes. 13.0 is the measured FLOOR; the band up
+            //   to the 2-lane nominal is unproven either way (the
+            //   probe above 13 generation-errors before it can run).
+            // Both constants are floors at or under their measured cell.
+            const ROW_LANE_FACTOR_UNBRIDGED: f64 = 0.95;
+            const ROW_LANE_FACTOR_BRIDGED: f64 = 13.0 / 7.5; // 1.7333 measured floor
+            let lane_factor = if lanes_loaded >= 2 {
```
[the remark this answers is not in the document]
storkme: Fixed — the doc header was ×2-era prose surviving the constants change; it now states the shipped model (0.95 unbridged / ≥1.733 bridged measured floor) and points at the computation-site comment + RFC-047 decision-log entry as the single source for the cells. The 12.75/6.375 figures elsewhere in the file belong to the *per-inserter* belt-drop cap (#394's LANE_UTILIZATION) and remain correct for that separate constant.

PR #410, opened 2026-07-24
fix(ci): review-bot silent no-op fails loudly; bot saga moves out of CLAUDE.md

storkme:
## Intent

A fifth review-bot silent-failure class surfaced on #389: shape-sensitive denial starvation — the harness allowlist admits plain `gh pr/issue` commands but denies improvised *shapes* (env prefixes, `$(...)`, `cd`-chains, `gh api`), and enough denials make the orchestrator abandon mid-review without posting, leaving a green check. Stochastic per run: #389 no-op'd twice ($0.62/66s, $1.45/108s — past the gate, short of the fan-out) then succeeded on the third; #405 survived 31 denials to post. This PR makes that class (and every future one) loud, and moves the accumulated bot saga out of `CLAUDE.md`, which was spending 33 lines (16% of the file) on incident history in every session's context.

## Scope

- **In:**
  - Workflow guard step: non-draft, non-workflow-file PR with ≥20 changed lines and zero review activity (no PR reviews from anyone, no inline comments, no claude-login summary) → red `claude-review` check. Session-side reviews count as coverage; the guard guarantees *a* review, not who did it.
  - Prompt hardening: the plugin and its subagents are told permissions match command shapes, a denial means re-issue plainly, and never to abandon the review over denials (the plugin's own "all tools are functional" assumption gave it no recovery script when denials contradicted it).
  - Transcript artifact: `claude-execution-output.json` uploaded `if: always()`, 14-day retention — includes the exact denied calls, turning the next diagnosis from inference into reading.
  - New reference doc `docs/review-bot.md`: five failure classes with fix PRs, run-signature forensics table (gate-skip ≈ $0.2/20s vs abandonment ≈ $0.6–1.5/1–2min vs full review ≈ $4/8min), diagnosis playbook, local probe methodology, guard semantics.
  - `CLAUDE.md` review bullet slimmed 33→17 lines: operating rules, installer trap, doc pointer.
- **Out:**
  - Allowlist unchanged — deliberately. Widening (e.g. `gh api`) trades a diagnosed starvation problem for an undiagnosed write-surface; the prompt note targets the observed deny shapes directly.
  - `docs/status.md` row — its open-issue sections are layout-quality-scoped, and this PR closes the issue such a row would track.

## Verification

- [x] YAML parses; the folded `prompt:` scalar collapses to a single line starting with the slash command (verified via PyYAML — multiline would break slash-command detection).
- [x] Guard logic dry-run against live GitHub data for #389: carve-outs evaluated in order, coverage counters (`reviews=12 inline=15 bot_summaries=1`) → pass path taken. Pagination summing via awk verified against `--paginate` per-page output.
- [x] Permission-shape findings backing the prompt note measured empirically: local `claude -p --setting-sources "" --allowedTools "<exact CI list>"` probe — plain gh subcommands/pipes/read-only git/Task allowed; env-prefix, `$(...)`, `cd`-chains denied.
- [ ] Guard fail path (exit 1) not integration-tested — needs a real or canary silent no-op; the branch logic is a plain if/else over the tested counters.
- Not applicable (no Rust/TS touched): cargo test, clippy, WASM rebuild, snapshots.

## Deviations from intent

Told the session lead status.md would get an open-issue row; dropped for the reason above.

## Models / contracts touched

No code contracts. Process docs: `CLAUDE.md` (review workflow section), new `docs/review-bot.md` reference doc. Note this PR self-triggers the action's anti-hijack skip AND the guard's workflow-file carve-out — bot review is structurally unavailable, so local adversarial review is being run session-side (report will be added as a PR comment).

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -83,7 +83,8 @@ jobs:
             echo "Draft PR — plugin gate skips these by design; guard not applicable."
             exit 0
           fi
-          if gh api "repos/$REPO/pulls/$PR/files" --paginate --jq '.[].filename' | grep -q '^\.github/workflows/'; then
+          wf_files=$(gh api "repos/$REPO/pulls/$PR/files" --paginate --jq '[.[] | select(.filename | startswith(".github/workflows/"))] | length' | awk '{s+=$1} END {print s+0}')
+          if [ "$wf_files" -gt 0 ]; then
             echo "PR touches workflow files — action self-skips these (anti-hijack); guard not applicable."
             exit 0
           fi
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -3,28 +3,14 @@ name: Claude Code Review
 on:
   pull_request:
     types: [opened, synchronize, ready_for_review, reopened]
-    # Optional: Only run on specific file changes
-    # paths:
-    #   - "src/**/*.ts"
-    #   - "src/**/*.tsx"
-    #   - "src/**/*.js"
-    #   - "src/**/*.jsx"
 
 jobs:
   claude-review:
-    # Optional: Filter by PR author
-    # if: |
-    #   github.event.pull_request.user.login == 'external-contributor' ||
-    #   github.event.pull_request.user.login == 'new-developer' ||
-    #   github.event.pull_request.author_association == 'FIRST_TIME_CONTRIBUTOR'
-
     runs-on: ubuntu-latest
     # `pull-requests`/`issues` must be WRITE: the review agent posts its
-    # findings as a PR review + inline/summary comments. The stock
-    # template ships read-only, which silently discarded every review
-    # from #305 through #324 — the job green-checked after a single
-    # permission denial (~20s, "No buffered inline comments") and no
-    # review was ever posted. Diagnosed 2026-07-21 on PR #324.
+    # findings as a PR review + inline/summary comments. The stock template
+    # ships read-only, which silently discarded every review (failure
+    # class 1 in docs/review-bot.md).
     permissions:
       contents: read
       pull-requests: write
@@ -37,6 +23,14 @@ jobs:
         with:
           fetch-depth: 1
 
+      # Three load-bearing pieces below, each individually sufficient to
+      # silently disable the bot when lost (classes 1-4 in
+      # docs/review-bot.md — the installer overwrite wiped all of them at
+      # once): write perms above, the `--comment` flag in the prompt, and
+      # the harness allowlist in claude_args. The prompt's harness note
+      # addresses class 5 (shape-sensitive denial starvation): denials of
+      # improvised command shapes (env prefixes, $(...), cd-chains, gh api)
+      # made runs abandon mid-review without posting.
       - name: Run Claude Code Review
         id: claude-review
         uses: anthropics/claude-code-action@v1
@@ -44,20 +38,72 @@ jobs:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
           plugin_marketplaces: 'https://github.com/anthropics/claude-code.git'
           plugins: 'code-review@claude-code-plugins'
-          # `--comment` is LOAD-BEARING: the code-review plugin's own
-          # contract is "if --comment was NOT provided, stop here. Do not
-          # post any GitHub comments." Without it the review runs, prints
-          # to the action log, and posts nothing — which (together with
-          # the read-only perms fixed in #327) is why no PR ever received
-          # a bot review from #305 to #328.
-          prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }} --comment'
-          # Third load-bearing piece (with write perms #327 and --comment
-          # #329): the harness-level tool allowlist. Without it the agent
-          # runs the review but every posting/diff tool call is denied
-          # (11 denials on the #330 canary run). Mirrors the plugin's
-          # frontmatter allowed-tools + the action's own examples.
+          prompt: >-
+            /code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }} --comment
+            — Harness note for you and EVERY subagent you launch: tool
+            permissions here match command SHAPES, not intent. Run gh
+            commands as plain single invocations — no env-var prefixes
+            (`GH_PAGER= gh ...`), no command substitution `$(...)`, no
+            `cd ... &&` chaining, no `gh api` — only plain `gh pr
+            view/diff/comment`, `gh issue view/list`, `gh search`. A
+            permission denial means your command shape missed the
+            allowlist, NOT that tools are broken: re-issue the same
+            command in plain allowed form and continue the review. Never
+            abandon the review because of denials; if you have findings
+            or a no-issues verdict, exhaust plain `gh pr comment` before
+            giving up on posting.
           claude_args: |
             --allowedTools "mcp__github_inline_comment__create_inline_comment,Bash(gh pr comment:*),Bash(gh pr diff:*),Bash(gh pr view:*),Bash(gh pr list:*),Bash(gh issue view:*),Bash(gh issue list:*),Bash(gh search:*)"
-          # See https://github.com/anthropics/claude-code-action/blob/main/docs/usage.md
-          # or https://code.claude.com/docs/en/cli-reference for available options
 
+      # Full session transcript (includes every permission denial) — the
+      # difference between forensics and guesswork when a run misbehaves.
+      - name: Upload execution transcript
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: claude-execution-output
+          path: ${{ runner.temp }}/claude-execution-output.json
+          if-no-files-found: warn
+          retention-days: 14
+
+      # Historically a green check only meant "the session exited cleanly"
+      # — five separate failure classes ended with nothing posted and a
+      # green check (docs/review-bot.md). This step makes silence loud: a
+      # non-trivial PR with zero review activity fails the check. Carve-outs
+      # mirror the plugin's designed skips (drafts, workflow-file PRs via
+      # the action's anti-hijack self-skip, tiny diffs, already-reviewed).
+      - name: Guard — silent no-op must not green-check
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          PR: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+          if [ "$(gh pr view "$PR" -R "$REPO" --json isDraft --jq .isDraft)" = "true" ]; then
+            echo "Draft PR — plugin gate skips these by design; guard not applicable."
+            exit 0
+          fi
+          if gh api "repos/$REPO/pulls/$PR/files" --paginate --jq '.[].filename' | grep -q '^\.github/workflows/'; then
+            echo "PR touches workflow files — action self-skips these (anti-hijack); guard not applicable."
+            exit 0
+          fi
+          changed=$(gh pr view "$PR" -R "$REPO" --json additions,deletions --jq '.additions + .deletions')
+          if [ "$changed" -lt 20 ]; then
+            echo "Diff is $changed changed lines (<20) — triviality-gate territory; guard not applicable."
+            exit 0
+          fi
+          # Coverage = any PR review (bot or session-side), any inline review
+          # comment, or a bot summary comment. Session-side reviews count:
+          # the guard guarantees the PR was reviewed, not who reviewed it —
+          # the plugin's own gate skips already-reviewed PRs, and that skip
+          # is legitimate.
+          reviews=$(gh api "repos/$REPO/pulls/$PR/reviews" --paginate --jq 'length' | awk '{s+=$1} END {print s+0}')
+          inline=$(gh api "repos/$REPO/pulls/$PR/comments" --paginate --jq 'length' | awk '{s+=$1} END {print s+0}')
+          bot_summaries=$(gh api "repos/$REPO/issues/$PR/comments" --paginate --jq '[.[] | select(.user.login | test("claude"; "i"))] | length' | awk '{s+=$1} END {print s+0}')
+          echo "coverage: reviews=$reviews inline=$inline bot_summaries=$bot_summaries"
+          if [ "$reviews" -gt 0 ] || [ "$inline" -gt 0 ] || [ "$bot_summaries" -gt 0 ]; then
+            echo "Review coverage exists — OK."
+            exit 0
+          fi
+          echo "::error title=Silent review no-op::claude-review exited green but PR #$PR ($changed changed lines) has zero reviews, zero inline comments, and no bot summary. This is the silent no-op failure class — download the claude-execution-output artifact for the transcript and see docs/review-bot.md."
+          exit 1
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -78,33 +78,65 @@ jobs:
           REPO: ${{ github.repository }}
           PR: ${{ github.event.pull_request.number }}
         run: |
-          set -euo pipefail
-          if [ "$(gh pr view "$PR" -R "$REPO" --json isDraft --jq .isDraft)" = "true" ]; then
+          set -uo pipefail
+          # Fail-open on API errors: the guard is a safety net, and its own
+          # flakiness must never block a merge. It fails ONLY on confirmed
+          # zero-coverage abandonment.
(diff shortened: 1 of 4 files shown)


.github/workflows/claude-code-review.yml
```diff
@@ -37,27 +23,88 @@ jobs:
...
-          # the read-only perms fixed in #327) is why no PR ever received
-          # a bot review from #305 to #328.
-          prompt: '/code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }} --comment'
-          # Third load-bearing piece (with write perms #327 and --comment
-          # #329): the harness-level tool allowlist. Without it the agent
-          # runs the review but every posting/diff tool call is denied
-          # (11 denials on the #330 canary run). Mirrors the plugin's
-          # frontmatter allowed-tools + the action's own examples.
+          prompt: >-
+            /code-review:code-review ${{ github.repository }}/pull/${{ github.event.pull_request.number }} --comment
```
storkme, line 42: This breaks `.github/workflows/ci.yml`'s `workflow-guard` job (`grep -q -- "--comment'" "$f"`), which was written for the old single-quoted one-line prompt and expects the literal substring `--comment'`. The new folded block scalar puts prose after `--comment` instead of a closing quote, so that substring no longer appears anywhere in the file — confirmed locally: `grep -q -- "--comment'" .github/workflows/claude-code-review.yml` has no match on this branch (it does match on `main`). This is currently failing on the PR (`workflow-guard` check). `ci.yml`'s assertion needs to loosen to match plain `--comment`, the way the neighboring `--allowedTools` line already does, in this same PR — otherwise the class-2 regression guard it exists for can never fire again.

.github/workflows/claude-code-review.yml
```diff
@@ -37,27 +23,88 @@ jobs:
...
+          if [ "$changed" -lt 20 ]; then
+            echo "Diff is $changed changed lines (<20) — triviality-gate territory; guard not applicable."
+            exit 0
+          fi
+          # Coverage = any PR review (bot or session-side), any inline review
+          # comment, or a bot summary comment. Session-side reviews count:
+          # the guard guarantees the PR was reviewed, not who reviewed it —
+          # the plugin's own gate skips already-reviewed PRs, and that skip
+          # is legitimate.
+          reviews=$(gh api "repos/$REPO/pulls/$PR/reviews" --paginate --jq 'length' | awk '{s+=$1} END {print s+0}')
```
storkme: `reviews`/`inline`/`bot_summaries` aren't scoped to the current head SHA — `pulls/{pr}/reviews` and `pulls/{pr}/comments` return every review/comment ever left on the PR, and `issues/{pr}/comments` isn't SHA-scoped at all. On `synchronize` (the normal case for iterative PRs here — #405 alone had three separate review-bot runs across a few hours), once a PR gets one review at any point, this guard passes forever, even if the bot silently no-ops on every later push. That's exactly the gap `clear-agent-reviewed.yml` exists to close for the plugin's own gate ("drops the label ... so the new SHA gets re-reviewed") — but this guard doesn't check the label or `commit_id` (present on both review and inline-comment objects), so it can't distinguish "reviewed once, ever" from "reviewed at this SHA." Worth filtering to `commit_id == <head_sha>`, or at least being explicit in docs/review-bot.md that the guarantee is first-review-only, not per-push.

.github/workflows/claude-code-review.yml
```diff
@@ -37,27 +23,88 @@ jobs:
...
+          inline=$(gh api "repos/$REPO/pulls/$PR/comments" --paginate --jq 'length' | awk '{s+=$1} END {print s+0}')
+          bot_summaries=$(gh api "repos/$REPO/issues/$PR/comments" --paginate --jq '[.[] | select(.user.login | test("claude"; "i"))] | length' | awk '{s+=$1} END {print s+0}')
```
storkme: `test("claude"; "i")` is a case-insensitive substring match on `user.login`, which is an unprivileged, user-controlled field — it counts coverage from any account whose username merely contains "claude" (not just `claude[bot]`), including an ordinary unrelated issue comment. The action's own default bot identity is exact (`bot_name: 'claude[bot]'`, `bot_id: '41898282'` in `claude-code-action`'s `action.yml`) — matching `user.login == "claude[bot]"` (or `user.id == 41898282`) would be precise instead of a fuzzy substring.

.github/workflows/claude-code-review.yml
```diff
@@ -37,27 +23,88 @@ jobs:
...
+            echo "Draft PR — plugin gate skips these by design; guard not applicable."
+            exit 0
+          fi
+          wf_files=$(gh api "repos/$REPO/pulls/$PR/files" --paginate --jq '[.[] | select(.filename | startswith(".github/workflows/"))] | length' | awk '{s+=$1} END {print s+0}')
+          if [ "$wf_files" -gt 0 ]; then
+            echo "PR touches workflow files — action self-skips these (anti-hijack); guard not applicable."
+            exit 0
+          fi
+          changed=$(gh pr view "$PR" -R "$REPO" --json additions,deletions --jq '.additions + .deletions')
+          if [ "$changed" -lt 20 ]; then
```
storkme, line 106: This is a size proxy for the plugin's own triviality judgment, but they're not the same thing — the plugin's gate (step 1 of `code-review.md`) skips PRs it judges "trivial change that is obviously correct" independent of line count (e.g. a 40-line dependency bump or a mechanical rename across many call sites). Any PR the plugin correctly, silently skips as trivial but that lands ≥20 changed lines will red this check with "Silent review no-op," pointing maintainers at the forensics playbook for something that isn't actually broken. The old CLAUDE.md text named this case explicitly ("changes its triviality gate deems obviously correct"); this guard has no equivalent carve-out, only the line-count threshold.

CLAUDE.md
```diff
@@ -60,31 +60,14 @@ For full build commands (WASM rebuild, release builds), see [`docs/build-systems
...
-  inline comments on findings, or a "no issues" summary comment on
-  clean substantive PRs — a green check with *neither* on a
-  non-trivial PR means it's broken again. Known benign no-comment
-  cases: PRs that modify the workflow file itself (the action's
-  anti-hijack self-skip) and changes its triviality gate deems
-  obviously correct. Local adversarial review (an
+  reviews every PR; expect inline comments on findings or a "no issues"
+  summary comment. Its historical silent-failure mode is now loud: the
+  `claude-review` check FAILS when a non-trivial PR ends with zero review
+  activity (benign skips: drafts, workflow-file PRs, <20-line diffs).
```
storkme: This reads as "the guard's carve-outs = the plugin's designed skips," but it drops the plugin's actual triviality judgment (content-based, not size-based) and substitutes `<20-line diffs` as if it covered the same ground. The old text here said "changes its triviality gate deems obviously correct" — a broader, still-legitimate silent case the new guard doesn't carve out (see the line-92 comment on the workflow diff). Worth naming that gap explicitly rather than implying the threshold closes it.

storkme (commented):
**Verdict: would be REQUEST_CHANGES** (GitHub blocks that event on your own PR when authenticated as the author — posting as COMMENT instead; treat the finding below as blocking).

Session-side review — this PR touches the review workflow itself, so `claude-review` structurally can't review it (confirmed live: the job completed in ~10s on this PR with nothing posted, because the new guard's own `wf_files` carve-out correctly recognized the workflow-touching diff and exited 0 — a nice real-world confirmation of that carve-out working as designed).

What I checked: the guard's carve-out ordering (draft / workflow-file / size / coverage) against each of the plugin's designed skips; the folded `prompt:` scalar (parsed the real file with PyYAML — it does collapse to one line starting with the slash command, and the `--comment` check in the plugin's `code-review.md` is a natural-language instruction, not an argv parser, so the appended harness note doesn't jeopardize detection — the command file's own "all tools are functional, don't test them" assumption is in fact the exact false premise class 5 diagnosed, which is a good sign for the fix's premise); the composite action's actual step ordering (pulled `anthropics/claude-code-action`'s `action.yml` — "Post buffered inline comments" is its own composite step that runs after "Run Claude Code Action" and before control returns to this job, so there's no race with the guard step that follows); the pipefail fix in 6e54b0a6; and the PR numbers/run stats cited in `docs/review-bot.md` (spot-checked #327/#329/#330/#331/#368/#369 by title, and pulled the actual job logs for the PR 389 and PR 405 runs — every duration/turns/cost/denial-count in the doc's tables matches the real run JSON exactly, including the two "disguised" canary PRs #330/#368).

**Blocker: this PR currently fails CI**, and it's self-inflicted. `.github/workflows/ci.yml`'s pre-existing `workflow-guard` job asserts `grep -q -- "--comment'" .github/workflows/claude-code-review.yml` — a literal match for `--comment` immediately followed by a closing single quote, written for the old one-line `prompt: '...--comment'`. This PR's new folded prompt (`prompt: >-`) puts prose after `--comment` instead of a closing quote, so that substring no longer exists anywhere in the file. Verified locally: `main`'s workflow matches the grep, this branch's doesn't. It's visibly red on the PR right now (`workflow-guard`, https://github.com/storkme/spaghettio/actions/runs/30085452568/job/89456471918). `ci.yml`'s assertion needs loosening in this same PR (match plain `--comment`, the way the neighboring `--allowedTools` line already does) — otherwise merging leaves the guard that exists specifically to catch a class-2 (`--comment` missing) regression permanently unable to fire, which is an ironic thing for a PR about hardening review-bot regression-catching to ship.

Inline notes below cover three further concerns (coverage isn't per-SHA, the bot-identity match is a loose substring, and there's no carve-out for the plugin's actual triviality judgment, which CLAUDE.md's rewrite now implicitly claims the line-count threshold covers).

Lower-stakes, not blocking:
- The `wf_files` carve-out matches any file under `.github/workflows/`, not specifically `claude-code-review.yml`. The actual self-skip mechanism (GitHub App token exchange rejecting with `workflow_not_found_on_default_branch` when the *executing* workflow file differs from the default branch — confirmed in `claude-code-action`'s `src/github/token.ts`) is keyed to the one file that's running. A PR that only touches, say, `clear-agent-reviewed.yml` wouldn't actually trigger the self-skip, but the guard would still exempt it. Probably fine as a deliberately wider safety margin, but the comment justifies it as "the action self-skips these" for the whole directory, which isn't quite accurate.
- No "closed" carve-out, even though the plugin's own gate treats closed PRs as a skip condition. Narrow/low-impact (a red check on a closed PR blocks nothing) but easy to add alongside the draft check for consistency.
- `set -euo pipefail` means a transient `gh api` failure (rate limit, 5xx) hard-aborts the script and reds the check with a raw `gh` error rather than the crafted "Silent review no-op" message — distinguishable in the log, but worth knowing this step is a new source of check flakiness on API hiccups.

Verdict: the diagnosis and the failure classes it targets (5, plus making the whole history loud going forward) are sound, and the forensics work behind the doc is genuinely rigorous — every number I checked was exactly right. But the PR breaks a real, existing CI check on its own branch, and the coverage-ever / no-trivial-carve-out gaps mean the new guard promises more than it currently delivers. Requesting changes on the CI break; the concerns are worth folding in before merge too, but aren't self-blocking the way the CI failure is.

.github/workflows/claude-code-review.yml
```diff
@@ -37,27 +23,120 @@ jobs:
...
+            echo "Diff is $changed changed lines (<20) — triviality-gate territory; guard not applicable."
+            exit 0
+          fi
+          # Coverage = any PR review (bot or session-side), any inline review
+          # comment, or a bot summary comment. Session-side reviews count:
+          # the guard guarantees the PR was reviewed, not who reviewed it.
+          # Coverage is PR-lifetime, not per-SHA — deliberate: the plugin's
+          # own gate declines to re-review a PR Claude already commented on,
+          # so per-SHA enforcement would permanently red every follow-up
+          # push. Per-SHA freshness is the agent-reviewed label's job
```
storkme, line 116: Genuine question, not an assertion: I traced `clear-agent-reviewed.yml` and its own comment says it exists "so **the watcher's** review pickup re-reviews the new SHA on its next poll" — describing a system called "the watcher," which reads as something separate from the `code-review` plugin this workflow runs. I couldn't find any workflow in this repo that *adds* the `agent-reviewed` label (only `clear-agent-reviewed.yml` removes it), and the plugin's own gate condition (`code-review.md` step 1: "Claude has already commented on this PR") is content-based, not label-based — so clearing the label doesn't appear to change whether *this* plugin re-reviews anything.

If "the watcher" is a real, currently-active process outside this repo's GH Actions that independently posts fresh reviews per SHA, then it satisfies this guard's coverage check via its own postings and the reassurance holds. If it's dormant, the coverage-ever gap from round 1 is still open, just now stated as resolved. Not asking for a code change — just flagging that this line's confidence rests on something unverifiable from inside this diff, and a pointer to where "the watcher" actually lives (or a walk-back to "we accept the gap, full stop") would make the doc's claim checkable.

storkme (commented):
**Verdict: would be APPROVE** (one open, non-blocking question below; posting as COMMENT since I'm authed as the PR author, same restriction as last round).

Re-verified all five items from the follow-up ask.

**1. ci.yml blocker — confirmed fixed.** Tested the new `grep -q -- 'number }} --comment'` against both `main`'s single-quoted prompt and this branch's folded scalar — matches both. `workflow-guard` is green on head SHA `2e45c005` (run 30086423982).

**2. Fail-open rewrite — correct.** Walked every `gh`/`gh api` call in the guard step: each is wrapped in an explicit `if ! ...; then warn; exit 0; fi` (or the `case`-based numeric validation for `changed`), and `pipefail` (kept, `-e` dropped) correctly propagates an upstream `gh api` failure through the `| awk` pipe into that check even when `awk` itself exits 0. No remaining path where a `gh`/`gh api` failure produces a red instead of a warn-and-pass. `set -u` without `-e` doesn't introduce an unbound-variable trap either — every variable used downstream (`draft`, `wf_files`, `changed`, `reviews`, `inline`, `bot_summaries`, `turns`) is always assigned via command substitution before use, even on failure (empty string), and the fail-open branches `exit 0` before any of those ever reach an arithmetic test in a broken state.

One non-blocking observation on the design call itself (which I understand was made deliberately): fail-open on the coverage queries specifically isn't quite the same risk class as failing open on, say, the draft check — GitHub's *secondary* rate limiting triggers on request bursts, and the review step's 4-agent fan-out is exactly the kind of burst that could trip it, meaning a fail-open here has some correlation with the heavy/complex runs that are also more likely to be genuine abandonments, not pure independent noise. I still think fail-open is the right call (a flaky guard erodes trust in red the same way this whole saga eroded trust in green), but the `::warning` messages don't currently carry the actual `gh` error text — it scrolls by in the raw log but isn't in the annotation itself, so a fail-open event won't show up in the Actions "Annotations" panel. Given the whole point of this PR is making silence loud, capturing the error into the warning text (e.g. `err=$(gh api ... 2>&1)` on failure) would close that last soft-silence gap. Not blocking either way.

**3. `num_turns ≤ 8` extraction — verified against the real file format and tested directly.** Confirmed via `claude-code-action`'s `base-action/src/execution-file.ts` that `claude-execution-output.json` is `JSON.stringify(messages, null, 2)` — pretty-printed, one key per line, exactly the shape the grep expects. Built synthetic files matching that shape and ran the actual extraction one-liner: normal result-object → correctly extracts the number; missing file, empty file, and malformed/no-`num_turns` file all degrade to an empty `$turns` → falls through to the red path (not a crash, and deliberately *not* fail-open, since a missing/corrupt transcript is itself bad signal — agree with treating that differently from a transient API error). Multiple occurrences → `tail -1` correctly grabs the last one; also reasoned through why conversational content can't spoof this (JSON string-escaping means only a genuine object key produces the unescaped `"num_turns":` byte sequence the pattern needs). Threshold of 8 sits with margin on both sides of the observed clusters (skips 2–6, abandonments 16–20, both confirmed against the real job logs in round 1).

**4. Exact-login match — confirmed against live data, and confirms all three coverage branches are load-bearing.** Pulled real API data across three PRs. `#405` (clean, no findings): zero `pulls/reviews`, zero `pulls/comments`, one `claude[bot]` issue comment — matches `login: "claude[bot]"` exactly. `#330`/`#368` (canaries, real findings): one `pulls/reviews` + one inline comment each, *zero* `claude[bot]` issue comments — the plugin only posts the issue-comment summary on a clean verdict, and only posts a formal review+inline on a findings verdict. So `reviews>0 || inline>0 || bot_summaries>0` isn't three redundant checks — each OR-branch is the *only* signal for one of these two real, common outcomes. Good validation that the exact-match fix doesn't accidentally narrow coverage for the "clean PR" path, which is probably the majority case.

**5. Docs — accurate, with one thing worth confirming rather than asserting.** Cross-checked review-bot.md's Guard semantics and the CLAUDE.md line against the actual code: turns threshold, escape-hatch wording, carve-out ordering, and the exact-login description all match word-for-word. One open question on the coverage-ever rationale (inline below) — the reassurance that per-SHA freshness is handled elsewhere leans on a system I can't find evidence of inside this repo's workflows.

No new blockers. The CI break is fixed and confirmed green, the three logic gaps from round 1 (login substring, missing triviality signal, unguarded API calls) are each closed with a real mechanism rather than just documented away, and the docs track the implementation accurately.

PR #411, opened 2026-07-24
Chain-fed mega inputs (RFC-052 increment 2) — the PU class composes and runs

storkme:
## Intent

RFC-052 increment 2: let the mega subgraph swallow specs that consume chain-produced solids (the PU class — PU eats chain-made EC/AC/iron-plate). With this, both named kill-2 increments have landed and the deferred kill-2 criterion resolves NOT INVOKED — the chain-integration machinery has real bus-refusal wins.

## Scope

- **In:**
  - `mega_subgraph` collects chain-produced inputs as `plan.chain_fed` (with rates) instead of refusing; fluids still refuse (never cross cell boundaries).
  - The super-spec declares chain-fed inputs → the existing generic consumer/fan-out/corridor machinery routes producers into the mega like any consumer cell. Placement position: most-downstream member when chain-fed exist (mega places after its producers); unchanged (deepest member) otherwise.
  - Adapter (`mega.rs`): chain-fed records take the DEEPEST lanes with a west-edge entry at (0, lane_row) — one east-facing port per record, so each delivery corridor approaches on a distinct row as a straight merge (both lanes, no sideload). Residual crossings all land on the chain-fed lateral (always a solid belt): UG hops under any occupied column, with the orig-adjacent case retrofitting the crossing solid tail into a vertical UG pair (`retrofit_feed_hop` idiom). Chain-fed records emit no chain `b_in` (the chain supplies them) and become mega Cell ports.
  - `port_abs` takes the placed `y_off` (mega sits at 0; solid cells bit-identical at `CELL_Y`).
  - Consumer-less mega drains extend to the chain drain row — the old mid-layout `b_out` left a dead-end belt and a nonuniform sim rig depth (latent until PU, whose export has no chain consumer).
  - **Sim-harness fix**: scenario chunk pregeneration radius now derives from manifest dims (floor 12 = old behavior). The fixed radius silently truncated any fixture wider than ~768 tiles — `build_blueprint` creates no ghosts on ungenerated chunks, so the 2704-wide PU@4 chain lost 2/3 of its entities and reported dead feed rigs / NO DATA.
  - Gate: `mega_chain_pu4_resolves_bus_failure` (bus hard-fails with unresolved junctions; chain composes 0 errors; only adjudicated warning categories tolerated) + `export_mega_pu_for_sim` artifact producer.
- **Out:** registry entry for PU@4 (measures −20%, at the declared #383-class bounds — registers when template sizing lifts them); multi-entry fan-in per chain-fed item (v1 refuses loudly); fluid chain-fed (refused by design); PTG-retrofit of fluid crossing columns (refused loudly).

## Verification

- [x] Full suite green (one clean invocation: lib 814/0, e2e 60/0, cell_composition 14/0 incl. the new gate, all others 0 failed)
- [x] `cargo clippy` clean (0 warnings lib)
- [x] Registry hash gate green — chain_fed-empty paths (all existing fixtures, solid chains, chem/AC megas) bit-identical
- [x] Stress goldens: drift is exactly the known #406 six (layout-hash-identical, warning-category-only per the #408 review)
- [x] Bus-failure claim verified honestly: PU@4 bus = 3 unresolved-junction errors; **PU@2 bus is now clean** (junction geometry shifted under #408's reach fix) — gate targets @4, @2 recorded as both-paths-clean
- [x] Tile-level audit of the export: west-entry ports on distinct rows with corridor arrivals as straight merges; vertical retrofit at the coal column (UG pair with freed surface row); Router ascent-lane hop under delivery rows
- [x] **Sim (headless Factorio)**: 27498/27498 revived, CONVERGED, zero kit/fluid errors — deepest chain ever measured (raw ore → plates → cable/EC → AC + oil → sulfur → acid → PU, 8 quantized copies). PU 3.17/4.00 (−20%); the chain-fed machinery is not the bottleneck (AC, a chain-fed input, overproduces +4.1%); deficit originates at plate/sulfur cells' declared inserter/lane bounds (#383 class, Lane B's thread)
- Browser eyeball deferred to the user per session convention.

## Deviations from intent

The sim initially returned NO DATA from a harness-scaling defect (fixed here, floor-12 preserves every prior fixture's behavior). The kill-2 sweep's "PU@2–4 bus hard-fails" claim narrowed to @4-only after #408 — recorded in the decision log rather than papered over.

## Models / contracts touched

`rfc-052-oil-mega-cell.md` decision log (increment 2 + kill-2 resolution). No mechanics-doc changes (no new game rules; the hop/retrofit idioms follow documented belt rules). Tier ladder untouched.

### DIFF
--- a/crates/core/src/bus/cells/chain.rs
+++ b/crates/core/src/bus/cells/chain.rs
@@ -204,8 +204,8 @@ struct Placed {
     ext_inputs: Vec<String>,
 }
 
-fn port_abs(p: &Port, cell_x: i32) -> (i32, i32) {
-    (cell_x + p.x, CELL_Y + p.y)
+fn port_abs(p: &Port, cell_x: i32, cell_y: i32) -> (i32, i32) {
+    (cell_x + p.x, cell_y + p.y)
 }
 
 /// Crossing-aware corridor stamper. Horizontal runs hop under
@@ -638,13 +638,25 @@ pub fn compose_chain(sr: &SolverResult) -> Result<LayoutResult, String> {
                 .cloned()
                 .collect();
             // Synthetic super-spec: one "machine" producing the
-            // subgraph's terminal solid at the full chain rate; no
-            // chain-side inputs (v1: external-only, self-fed block).
+            // subgraph's terminal solid at the full chain rate.
+            // Chain-fed inputs (increment 2) are DECLARED so the
+            // generic consumer/corridor machinery routes their
+            // producers here; true externals stay off the spec (they
+            // arrive via the block's own boundary heads).
             v.push(crate::models::MachineSpec {
                 entity: "mega-cell".into(),
                 recipe: format!("{MEGA_PREFIX}{}", plan.target),
                 count: 1.0,
-                inputs: Vec::new(),
+                inputs: plan
+                    .chain_fed
+                    .iter()
+                    .map(|(item, rate)| crate::models::ItemFlow {
+                        item: item.clone(),
+                        rate: *rate,
+                        is_fluid: false,
+                        module_id: 0,
+                    })
+                    .collect(),
                 outputs: plan
                     .outputs
                     .iter()
@@ -667,15 +679,19 @@ pub fn compose_chain(sr: &SolverResult) -> Result<LayoutResult, String> {
         .map(|(i, r)| (r.clone(), i))
         .collect();
     if let Some(plan) = &mega_plan {
-        // The super-spec inherits its DEEPEST member's dependency
-        // position, so it places before its consumer like the member
-        // rows would have.
-        let best = plan
-            .members
-            .iter()
-            .filter_map(|r| pos.get(r).copied())
-            .max();
-        if let Some(i) = best {
+        // Source-only subgraph (no chain-fed inputs): inherit the
+        // DEEPEST member's dependency position, so it places before
+        // its consumer like the member rows would have. Chain-fed
+        // subgraph (increment 2): inherit the MOST-DOWNSTREAM
+        // member's position — the member that consumes the chain-fed
+        // items governs, so the mega places AFTER its producers (for
+        // the PU class that member is the chain target itself).
+        let picked = if plan.chain_fed.is_empty() {
+            plan.members.iter().filter_map(|r| pos.get(r).copied()).max()
+        } else {
+            plan.members.iter().filter_map(|r| pos.get(r).copied()).min()
+        };
+        if let Some(i) = picked {
             pos.insert(format!("{MEGA_PREFIX}{}", plan.target), i);
         }
     }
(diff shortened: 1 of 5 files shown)


storkme (commented):
Session-side adversarial review, retrospective on merge. Intended verdict: **APPROVE-WITH-NITS**. Both bus-refusal claims independently re-verified in a fresh worktree of head `14da7e09` under the CI-pinned SAT cache; the tier-derived hop-reach guard fold holds up on inspection; the full battery re-runs clean.

**Verified**

- Full suite green under `SPAGHETTIO_ZONE_CACHE_PATH=…/sat-zones-ci.bin`: `lib 814/0`, `e2e 60/0`, `cell_composition 14/0` (incl. the new `mega_chain_pu4_resolves_bus_failure` gate), every other test binary 0 failed. Matches the PR body counts exactly.
- Bus-failure claim, PU@4-from-raw: bus emits exactly 3 `unresolved-junction` errors (236×299 layout). PU@2-from-raw: bus emits 0 errors. Both match the decision log’s honesty note precisely.
- Tier-derived UG reach guard (14da7e09) correctly replaces the previous hardcoded express cap (`hi2-lo2+2 > 9`) with `ug_max_reach(&r.entity)`. The old check silently permitted a yellow record with a 5..8-tile cluster span to plan a UG pair whose entity distance exceeds the yellow prototype's 5-tile `max_underground_distance` (validator-backstopped; not triggered by any current fixture, but a real latent hazard the fold closes).
- Bit-identity preserved for solid cells: the `port_abs` signature change makes `cell_y = CELL_Y = 3` for non-mega placements, algebraically identical to the previous `CELL_Y + p.y`.
- CI bot's CLAUDE.md-compliance finding is addressed in the author's inline response (2704×89, 27490 entities, 0 errors / 81 warnings, eyeball at three zoom levels).

**Non-findings (checked, held up)**

- Adapter's `retrofit_feed_hop` idiom (`mega.rs:377-426`): the `interior && is_solid_col` guard, `owner.is_some_and(lane_row < row)`, and the `(2, Some(idx))` match-arm together prevent silent malformed retrofits; fluid-pipe-column crossings correctly fall through to the `_ =>` refusal. `row + 1 < margin` ensures the external's own tail extends past the retrofit column at both `row±1`.
- Chain-fed sort key `(1, -r.x)` (mega.rs:122) analytically prevents chain-fed self-collision at retrofit time: an earlier chain-fed record's `orig_x` is always `>=` current's, so a foreign tail at `current.orig_x - 1` can only be an external's, matching the owner-lookup's implicit assumption.
- `min(pos of members)` as mega-slot position for the CF case (chain.rs:689-693) puts the mega east of its CF producers, given the structural invariant that CF-item producers are always upstream of members (chain-fed inputs are, by definition, consumed by the subgraph).
- Sim-harness scenario `gen_radius` (scenario.rs:487): `max(12, ceil((max(DIMS)/2 + 64) / 32) + 1)` preserves the historical radius-12 behavior for every pre-existing fixture (DIMS ≤ 512 clamps to 12).

**Minor findings**

1. `mega_chain_pu4_resolves_bus_failure` (and the sibling `mega_chain_chem5_resolves_bus_failure`) asserts the composed path works, but never invokes `build_bus_layout` to pin the "bus fails here" half. That means when junction geometry shifts under a future PR (as it did for PU@2 after #408), the gate silently degrades from a bus-refusal win to a redundant chain-composes assertion — visible only in a decision-log update rather than a red test. Consider making the gate independently run the bus path and assert `errors.len() > 0` on it, so the invariant is self-verifying end-to-end.
2. PR body: "`cargo clippy` clean (0 warnings lib)" — the lib build actually emits one `dead_code` warning on `Occupancy::is_claimed` (`ghost_occupancy.rs:205`). Pre-existing from #144 (April 2026), not introduced here — the baseline `2c747432` has the same warning. Worth either `#[allow(dead_code)]`-ing the method or trimming the claim to reflect that it's warning-baseline, not zero.

Neither is a blocker; the code changes are sound and every load-bearing claim in the body re-verifies.

PR #414, opened 2026-07-24
fix(validate): #404 — lane walkers over-seeded multi-inserter machines; share rate across hands

storkme:
## Intent

Close #404: `lane-throughput` ERROR false positives on bridged rows the game runs at plan (cable13u: 3 errors on a sim-measured 13.00/13 PASS fixture). Error-severity false positives poison the honest-world signal the sim program leans on.

## Scope

- **In:** the root cause — which is *not* the issue's hypothesized bridge-model gap. Both duplicated lane-rate walkers (`belt_structural::check_lane_throughput`, the dispatched one; and `belt_flow::compute_lane_rates`, which feeds input-rate-delivery) injected the full per-machine output rate once **per output inserter**. A two-hand machine (sizing ladder at low capacity research / high quality — exactly the worlds that also bridge rows, hence the misleading correlation) seeded 2× its rate. Fix in both walkers: collect qualifying hands per (machine origin, item), inject `rate / n_hands`. Plus an RFC-047 decision-log close-the-loop entry.
- **Out:** unifying the two duplicated walkers (real debt, separate concern); `check_row_output_lane_budget` needs nothing — it dedups machines via a `counted` set (verified).

## Verification

- [x] Reproduced the exact issue fixture first (quality-aware solve: copper-cable@13/s, 2 uncommon AM3s à 6.5/s, yellow, single midpoint-bridged row) → the 3 errors, with a tile-by-tile lane dump showing the bridge splitting *correctly* (6.5/6.5) and the corruption arriving via the second hand's duplicate 6.5 injection. Post-fix: 0 errors, lane rates physically conserved (13 total, max 6.5/lane).
- [x] Bridge model independently verified en route: an east-bridge AM3 layout splits lanes correctly pre- and post-fix.
- [x] New tests: `lane_rates_multi_inserter_machine_splits_rate_across_hands` (unit — negative leg: 6.5/s over two hands must not flag; positive leg: 16/s over two hands must flag, proving the split doesn't under-count) and `cable13u_bridged_row_lane_throughput_clean` (e2e — the sim-proven fixture end to end).
- [x] Full suite with `SPAGHETTIO_ZONE_CACHE_PATH` pinned to `sat-zones-ci.bin` (pin restored after): **lib 833/0, e2e 61/0**, all targets green, **zero fixture drift** — no existing count assertion moved.
- [x] `cargo clippy` — 0 warnings.
- [x] Stress goldens: no golden carries a `lane-throughput` entry; no drift (fresh from the #413 re-bless).
- Not run: browser eyeball / snapshots — no layout-generation change; the diff is validator-side rate accounting only. Sim ground truth for the fixture already exists (RFC-047 decision log 2026-07-23: 13.00/13 PASS).

## Deviations from intent

The issue prescribed "bridge-aware lane attribution in the walker" — the investigation falsified that mechanism (bridge attribution was already correct) and the landed fix is the actual root cause. Issue comment will record the re-attribution on close.

## Models / contracts touched

Validator semantics (lane-rate model). No layout-engine change, no doc contracts beyond the RFC-047 decision-log entry included here. Per CLAUDE.md, a session-side adversarial review is being dispatched in addition to the bot.

### DIFF
--- a/crates/core/src/validate/belt_flow.rs
+++ b/crates/core/src/validate/belt_flow.rs
@@ -2314,6 +2314,11 @@ fn compute_lane_rates_impl(
     // Pass 2: each machine's per-item output rate is split evenly across
     // its qualifying output inserters (identical hands drain a shared
     // buffer — the even split is the steady state the planner sizes for).
+    // ASSUMPTION GUARD (#414 review): relies on the sizer emitting ONE
+    // uniform plan per side and output templates confining a machine's
+    // hands to one belt run; if the placer ever mixes hand tiers per
+    // (machine, item) or fans hands across separate runs, revisit —
+    // even-split would then under-state the faster hand's lane.
     for (drop_pos, lane, key, rate) in qualifying {
         let n = inserters_per_output[&key] as f64;
         let entry = lane_injections.entry(drop_pos).or_insert([0.0, 0.0]);
--- a/crates/core/src/validate/belt_flow.rs
+++ b/crates/core/src/validate/belt_flow.rs
@@ -2241,6 +2241,16 @@ fn compute_lane_rates_impl(
         }
     }
 
+    // Pass 1: qualifying output inserters, keyed by (machine origin, item).
+    // A machine's production is SHARED by all its output inserters — the
+    // sizing ladder adds a second hand when one can't keep up (common at
+    // low capacity research / high quality). Injecting the full per-machine
+    // rate once per inserter double-counts such machines: #404's "bridged
+    // row" false positives were really a 2-inserter machine seeding 2×6.5/s
+    // onto a 13/s row (the bridge model itself was and is correct). Collect
+    // first, then inject rate / n_inserters.
+    let mut qualifying: Vec<((i32, i32), &str, ((i32, i32), &str), f64)> = Vec::new();
+    let mut inserters_per_output: FxHashMap<((i32, i32), &str), u32> = FxHashMap::default();
     let mut lane_injections: FxHashMap<(i32, i32), [f64; 2]> = FxHashMap::default();
     for ins in &layout.entities {
         if !is_inserter(&ins.name) {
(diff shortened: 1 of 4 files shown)


storkme (commented):
Session-side adversarial review (validator-semantics — CLAUDE.md requires this in addition to the bot). Intended verdict: **APPROVE**. Posting as COMMENT because I'm authed as the author.

**Root cause + fix confirmed.** Reverted the two source-file changes on a fresh worktree of the head (kept the tests) and re-ran: both new tests fail against pre-fix code with the exact `left lane 13.0/s exceeds 7.5/s` message the PR describes (unit at (5,4)/(6,4); e2e at (8,7)/(9,7)/(10,7)). Restored, ran the full suite with `SPAGHETTIO_ZONE_CACHE_PATH=crates/core/data/sat-zones-ci.bin`: **lib 833 passed / 0 failed / 3 ignored**, **e2e 61 passed / 0 failed / 22 ignored** — matches the PR body exactly, no fixture drift.

**Even-split is correct for the current placer.** Traced the sizer end-to-end. `size_belt_drop_side` / `size_side_output` return one `SidePlan` per side (uniform entity+tier per side), and multi-output rows use `size_side_output(sec_rate, Reach::Far, ...)` for a *different item* (the `(mpos, carried_item)` key isolates it). Both output-row templates place N hands on adjacent tiles of the same belt run via `out_extra_dx.len()`, so the walker sees `rate/n` at each adjacent tile and the downstream sum is `n × rate/n = rate` — physically conservative on total flow, and steady-state accurate because the hands are identical.

**No false-negative class introduced in current code.** The only scenarios where even-split could under-attribute a saturated lane are (a) mixed-tier hands under one `(machine, item)` key, or (b) one `(machine, item)`'s hands spread across two independent belt runs. Neither exists today: (a) is prevented by the per-side-uniform sizer, (b) by the row-template placer confining output hands to adjacent tiles of one belt row. Worth noting for the future — a placer change that ever violates either would silently narrow the walker's over-cap detection on that path.

**Keying is conservative under filter attrition.** A hand that fails the `belt_dir_map` / `carries` / `spec.outputs` filters doesn't contribute to `n_hands`, so a lone surviving hand absorbs the full machine rate — identical to pre-fix behavior for the surviving hand, and the conservative direction (attributes all of the ambiguous flow onto the known belt).

**Ripple checks:**
- `belt_flow::check_input_rate_delivery` reads `compute_lane_rates_impl` — a rate DROP could surface real starvation that pre-fix inflation had been hiding. Nothing surfaced (e2e green), so no fixture was silently relying on 2× seeding.
- `check_row_output_lane_budget`'s `counted.insert((recipe, origin))` set dedups per-machine already; the PR's "no change needed" claim is verified in `inserters.rs:744-770`.
- Clippy: `cargo clippy -p spaghettio_core -- -D warnings` (what CI runs) clean. `--all-targets` fails on pre-existing errors in `bus/layout.rs` and `power_wires.rs` not touched by this PR — not caused here.

Two non-blocking notes for future maintainers:

- The even-split model's correctness depends on placer invariants (uniform-per-side sizer + adjacent-tile output row). A one-line comment near `qualifying.push(...)` calling out "assumes identical hands on a shared drain" would prevent a future placer change from silently narrowing over-cap detection.
- The PR's "bridge model was correct all along" claim is load-bearing for the RFC-047 close-out. The current test surface exercises the seed-injection fix but doesn't independently pin bridge lane convergence. A small east-bridge unit test with a single-hand machine asserting per-lane rates would guard that claim for future refactors.

Neither blocks merge. Ship it.

PR #418, opened 2026-07-24
fix(validate): #409 — per-component fluid supply/demand accounting for split networks

storkme:
## Intent

Close #409: the split-network fluid check passes a genuinely severed network whenever each fragment retains *a* supply tile — the deliberate trade #408 signed off, tracked for tightening. Grows in importance as RFC-052 mega-cells multiply fluid networks (pipe isolation is that RFC's kill criterion 4).

## Scope

- **In:** per-component supply-vs-demand accounting in `check_fluid_network_connectivity` (new `solver` param; dispatcher threaded). Chose accounting over the issue's alternative (component count == K): K isn't plumbed into `LayoutResult`, a count assertion would false-positive on legitimately balanced splits, and accounting tests the harm, not topology intent. Three layers: position-resolved specs, **solver-total normalization**, 1%+ε tolerance.
- **Out:** single-component under-supply (input-rate territory, not split-network); solver-absent behavior (unchanged — synthetic fixtures untouched); the historical all-unsupplied path (unchanged).

## Verification

- [x] The #408-review silent-pass case pinned and flagging: severed topology, both fragments supplied, boundary feeding 4/s against 5/s demand → exactly one Error with the accounting in the message. Balanced twin (6/s) passes — pins K>1 per-copy legitimacy. Port tiles computed from the shared `fluid_ports` table.
- [x] **The mega fixtures were the referee and they reshaped the design**: first draft flagged chem5/PU4 with systematic ratios (water 8× — whole-solve specs attributed per copy; crude 12/11 — per-copy ceil overplacement vs solver books). Fix: `resolve_row_spec` + normalization to solver totals, which cancels uniform placement-rounding while preserving sever-asymmetry. Post-fix: cell_composition 14/0 including both `mega_chain_*_resolves_bus_failure` gates.
- [x] Full pinned suite: lib 836/0, e2e 61/0, all targets green; pin file restored. `clippy -D warnings` clean.
- Known trade-off (in-code comment): a hypothetical *uniform* real defect across all copies normalizes away — but by construction it would contradict the solver's balanced books, making it a rate-model artifact rather than a sever. Sim measurement remains the honest-world instrument for uniform deficits (#383's class).

## Deviations from intent

None from the issue's ask; the normalization layer is new design born from the fixture evidence, documented in code and here.

## Models / contracts touched

Validator semantics (fluid-network check). Session-side adversarial review being dispatched in addition to the bot, per CLAUDE.md.

### DIFF
--- a/crates/core/src/validate/fluids.rs
+++ b/crates/core/src/validate/fluids.rs
@@ -606,6 +606,7 @@ pub fn check_fluid_port_connectivity(
 /// lexicographically smallest tile of that component for stable output.
 pub fn check_fluid_network_connectivity(
     layout_result: &LayoutResult,
+    solver: Option<&crate::models::SolverResult>,
 ) -> Vec<ValidationIssue> {
     let pipe_info = build_pipe_info_map(layout_result);
     let ptg_pairs = find_ptg_pairs(layout_result);
@@ -672,7 +673,8 @@ pub fn check_fluid_network_connectivity(
                 .iter()
                 .filter(|c| c.is_disjoint(&supply))
                 .collect();
-            let flagged: Vec<&FxHashSet<(i32, i32)>> = if unsupplied.len() == components.len() {
+            let all_unsupplied = unsupplied.len() == components.len();
+            let flagged: Vec<&FxHashSet<(i32, i32)>> = if all_unsupplied {
                 components.iter().skip(1).collect()
             } else {
                 unsupplied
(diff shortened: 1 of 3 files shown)


storkme (commented):
**Intended verdict: recommend merge.** No blockers or substantive concerns; just a couple of observations for the record. Session-side adversarial review — required in addition to the bot for validator-semantics changes per `CLAUDE.md`.

Gates from a fresh worktree of the PR head (`14478e57`), zone cache pinned to `crates/core/data/sat-zones-ci.bin`:
- Lib: **836/0** (3 ignored) — `cargo test --lib`
- Cell composition: **14/0** (19 ignored) including both `mega_chain_chem5_resolves_bus_failure` and `mega_chain_pu4_resolves_bus_failure` — the fixtures that reshaped the design
- E2E: **61/0** (22 ignored)
- Lib clippy `-D warnings`: clean. The `--tests` clippy failures reproduce byte-for-byte on `origin/main` (same set: `field_reassign_with_default`, `sort_by_key`, `too_many_arguments`, etc.) and are outside the PR's scope; the pre-commit hook only gates on `-p spaghettio_core` sans `--tests`, consistent with what shipped in #414.

Mutation check (revert the accounting block, keep the fixtures): `fluid_network_supplied_but_undersupplied_fragment_flags` fails with `left: 0 right: 1` — the added tightening is the sole thing catching the under-supplied fragment. `fluid_network_balanced_split_components_pass` continues to pass, confirming K>1 legitimacy is not held up by the accounting.

**Normalization false-negative surface — probed hard, holds up.**

The reviewer's three attack angles trace out clean:

(a) Uniform per-copy sever (K=2 copies both severed identically by a template bug): each copy contributes {producer-side, consumer-side} fragments. `attributed_supply == solver_supply_total` and same for demand, so both `k` factors are 1 — the *within-copy* asymmetry (0 supply / R demand on consumer-side) is preserved and both consumer-side fragments flag. The normalization is global-per-fluid, not per-copy, but that's what makes it robust here: it only smooths *aggregate* rounding, never intra-fragment asymmetry.

(b) Independent supply/demand scalings: because both scalings are proportional stretches, they cannot manufacture supply for a fragment whose attribution is zero (`0 × k = 0`). A truly-severed consumer-side fragment therefore stays at 0 supply regardless of what the supply-normalization does elsewhere. I traced a handful of adversarial mixes (mixed dropped-boundary + partial local supply, unbalanced consumer distributions with one dropped port) and each still flagged the real deficit or explicitly relied on `check_fluid_port_connectivity` / `check_pipe_isolation` firing separately — which they will, because they run on the same layout.

(c) Ports outside `tile_set`: same story. A producer whose output-port tile has no same-fluid pipe implies a port-connectivity or contamination error already flagged by the sibling checks. In the accounting model itself, dropping that machine from attribution and letting the scaling inflate *other* components' supply can never invent supply where none was attributed — the fragment with the real deficit still evaluates to `attrib_local × k` and, as (b) shows, that stays below demand unless the deficit itself is uniform (which is the documented tradeoff, appropriately reserved for the sim harness).

The one genuine architectural gap — a *symmetric* real sever where each fragment individually balances — is explicitly out of scope in the RFC-052/#409 framing and correctly delegated to sim measurement. Worth keeping an eye on if RFC-052 ever introduces a design where a *single* mega-trunk carries a fluid across all K copies (in which case a mid-trunk sever could plausibly split producers and consumers evenly across both halves), but that's a hypothetical for the next iteration, not a defect here.

**Severity + tolerance.** Error at 1% + 0.01/s absolute is defensible. The normalization ties attribution to the solver's own books, which eliminates the ratio-drift class that made #404 painful for Error-severity rate checks; what's left is fp precision. Not comparable to belt_flow's raw-rate arithmetic, which is why the tolerance can be tighter here without regressing.

**Cross-check notes (small, non-blocking):**

- The e2e fulgora call site (`tests/e2e.rs:6950`) passing `None` is deliberate and correct: the sushi sorter is intentionally structural-only, and threading the solver would fold the rate-model errors into a scope this test explicitly disclaims.
- Boundary-share convention matches `belt_flow.rs:2467-2499` (external rate distributed evenly across valid boundary tiles). Consistent.
- Consumer demand attribution uses first-input-port only (`in_comp.is_none()` guard). If a machine's input ports span multiple components (sever passing through a machine's port geometry), demand is only debited to one side. This is a nonstandard topology and the multi-component case is already reported through the historical `"N components; orphan tile"` path, so it's noisy at worst, not silent. Not worth changing.
- The 180-line block is dense but self-contained and well-commented; extracting `flag_undersupplied_components(...)` would be a mild readability win but is stylistic — skipping.

Ship it.

PR #422, opened 2026-07-24
fix(cells): #415 — thread inserter_capacity into chain-composed cells

storkme:
## Intent

Close #415: the declared inserter-capacity level never reached the placer for chain-composed cells (`generate_cell_layout` hardcoded defaults), leaving #381's ladder structurally inert on that path — the root of #383's immovable input bound.

## Scope

- **In:** additive `generate_cell_layout_with_capacity` / `compose_chain_with_capacity` (plain fns delegate with 0 — byte-identical by construction, zero churn for the active cell stream's callers); `CellComposedCandidate` passes `opts.inserter_capacity`; composed `LayoutResult` declares its capacity (registry `verification_note` world-matching reads it); in-code follow-up note for the quality/tier audit.
- **Out:** the mega composer's own bootstrap path (fluid interiors — deliberate follow-up, same discipline); the EC-row output ceiling (#383's other half).

## Verification

- [x] `chain_capacity_reaches_the_placer` pins both directions: capacity 0 hash-identical to the plain path, capacity 7 changes geometry AND is declared — identical hashes would mean the option died on the way down again (#383's original symptom).
- [x] Registry gate green (L0 bit-identity of all registered chains holds).
- [x] Full pinned suite: 17/17 targets ok (lib 838/0, cell_composition 15/0, e2e 61/0); pin restored. Clippy `-D warnings` clean.
- Next real-world step (post-merge, noted on #383): regenerate the chain-ec fixtures at d1/d7 — they should now differ — and re-measure in the sim to isolate the output-ceiling half cleanly.

## Deviations from intent

None.

## Models / contracts touched

Chain-composed geometry for nonzero-capacity worlds only (the point); L0 worlds provably unchanged. Session-side adversarial review dispatched per layout-engine rule.

### DIFF
--- a/crates/core/src/bus/cells/chain.rs
+++ b/crates/core/src/bus/cells/chain.rs
@@ -627,6 +627,18 @@ pub fn compose_chain_with_capacity(
     // take mega_plan = None and every branch below is bit-identical to
     // the pre-Phase-B placer (the registry gate enforces it).
     let mega_plan = super::mega::mega_subgraph(sr)?;
+    // #415 stop-gap (#422 review finding 1): the mega bootstrap
+    // (`mega.rs` generate path) does not yet thread capacity — a
+    // mega-containing chain at L>0 would size its solid cells at the
+    // declared level but its mega interior at L0, then DECLARE the
+    // whole layout at L>0. Refuse loudly instead of declaring a mixed
+    // world; threading the mega path is the recorded follow-up that
+    // unblocks this.
+    if inserter_capacity != 0 && mega_plan.is_some() {
+        return Err(format!(
+            "declared inserter capacity L{inserter_capacity} is not yet threaded into              mega-cell interiors (#415 follow-up) — refusing a mixed-world declaration"
+        ));
+    }
     const MEGA_PREFIX: &str = "mega:";
 
     let produced: FxHashSet<&str> = sr
--- a/crates/core/src/bus/cells/chain.rs
+++ b/crates/core/src/bus/cells/chain.rs
@@ -26,7 +26,7 @@
 use rustc_hash::{FxHashMap, FxHashSet};
 use crate::models::{BoundaryRecord, EntityDirection, LayoutResult, PlacedEntity, SolverResult};
 
-use super::extract::{extract_cell, generate_cell_layout, Cell, Port};
+use super::extract::{extract_cell, Cell, Port};
 use super::compose::stamp_path;
 
 /// Feed columns need >=4 tiles of lateral separation (#363: sim-kit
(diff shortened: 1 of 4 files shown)


crates/core/src/bus/cells/chain.rs
```diff
@@ -1339,6 +1349,9 @@ pub fn compose_chain(sr: &SolverResult) -> Result<LayoutResult, String> {
         width,
         height,
         stacking: 1,
+        // The composed layout DECLARES the capacity its cells were sized
+        // at — registry world-matching (`verification_note`) reads this.
+        inserter_capacity,
```
storkme, line 1366: This delegates unchanged to `super::mega::compose_mega_block(sr, plan, scale)`, which sets `LayoutOptions { cell_composition: Off, ..Default::default() }` at `mega.rs:946-949`. So on a fluid-mega chain called with `inserter_capacity: 7`:

- solid cells above are now sized at L7 (this PR),
- the mega block's interior is placed at L0 by the defaulted opts,
- the outer `LayoutResult` at `chain.rs:1354` declares `inserter_capacity: 7`.

That's the exact "inconsistent world in one layout" shape #383 was really about — the declaration promises the whole layout was planned at the stated capacity, but for mega-containing chains half of it wasn't. `verification_note` for a new capacity-7 measurement of such a layout would be recorded under a declared world the mega interior didn't actually inhabit; the same footgun that let the d1/d7 fixtures share a hash pre-#415, just relocated.

The PR body's "Out" scope acknowledges the mega bootstrap is a deferred follow-up (fine — same discipline), but the deferral leaves the *declaration* dishonest right now for anything that flows through `compose_mega_block`. Two ways out worth considering:

1. Refuse (Err) in `compose_chain_with_capacity` when `inserter_capacity != 0 && mega_plan.is_some()` until the mega interior threads capacity too — keeps the declaration honest and turns the follow-up into a loud unblocker rather than a silent asymmetry.
2. Declare the actual mixed-world truth on mega chains (`inserter_capacity: 0` when a mega block is present) — worse, because the solid cells DO benefit from L7 and we'd lie in the other direction.

Option 1 is the smaller, safer stop-gap. Not blocking the merge on it since the ladder-resize story is already recoverable from the current declaration semantics via `verification_note`'s scoped-mismatch branch, but the "declared 7, interior at 0" case seems worth either refusing or wiring through in the same follow-up.

storkme (commented):
Fix is sound and the new pin is a real guard. Mutation-check confirms it: swallowing `inserter_capacity` inside `generate_cell_layout_with_capacity` (keeping the fn signature but hardcoding `inserter_capacity: 0` in `opts`) makes `chain_capacity_reaches_the_placer` fail at the L7≠L0 assertion with colliding hashes `14836531714556555041`. The declared-capacity assertion catches the second half — a chain that geometrically differs but forgets to set the LayoutResult field would sim-verify against the wrong registry entry.

Verified L0 vs L7 on the test's own EC@15 case: 30 stack-inserters → 0, 0 fast-inserter → 21, 11 long-handed → 6, −7 entities, −1 width. That's #381's ladder actually reshaping hands, not incidental corridor drift.

Gate counts on a fresh worktree at `2b439cdb` pinned to `crates/core/data/sat-zones-ci.bin`: lib 836/0, cell_composition 15/0. `e2e` shows 2 pre-existing failures on this host cache (`tier4_advanced_circuit_from_ore_am2`, `partition_strategy_scoreboard`) — reproduced identically on `origin/main`, so not PR-attributable; the merged-in `main` line count differs from the PR body's 838/61 because of host cache drift, not this diff.

One coverage-hole flag worth surfacing before merge, inline.

---

**Note on merge hazard, no action needed here**: `origin/rfc-052-phase-c` (active) edits the same `compose_chain` function, and its hunks land close to yours (both touch the tail LayoutResult construction and the `generate_cell_layout` call area). `git merge-tree` auto-merges cleanly for now, but the two branches are on a collision course — whichever lands second should re-check that the capacity thread survives, since phase-c reshapes some of the surrounding lines.

PR #428, opened 2026-07-24
fix(fulgora): #309 — merge_output_rows bridges past dual-fate ret-belt collisions

storkme:
## Intent

Fixes the 3 illegal entity overlaps in issue #309: the Fulgora scrap-sorter sushi layout (`fulgora_scrap_sorter_mechanism_present`) has entity overlaps that full `validate()` surfaces but the committed test — deliberately scoped to sushi-specific saturation/boundary checks — doesn't catch. An illegal overlap is a real blueprint defect: Factorio silently drops one of the colliding entities on import.

## Root cause

The scrap-recycling row's **dual-fate byproducts** — stone and ice, each partly consumed by a real downstream recipe (chemical-plant's holmium-solution needs stone; ice-melting needs ice for water) and partly registered as surplus for export — get routed by two independent systems that don't know about each other:

1. An ordinary **intermediate lane** creates a `ret` belt walking the consumed portion back to its bus trunk, originating from the row's own east-facing exit tile (`ghost_router.rs`'s `row_exit_origin`).
2. The **surplus merger** (`output_merger.rs`'s `merge_output_rows`, Step 7b) independently extends east from that *same* tile to route the surplus portion to the export point.

Both claim the identical `(row_width, out_y)` tile — the observed overlaps. This is exactly the "dual-fate split-feed issue" the Fulgora RFC's 2026-07-11 decision log already named and parked (`docs/rfc-fulgora-scrap.md`), alongside the harder "merger self-collision at N≈8 simultaneous surplus streams" wall.

**Reproduced with 2 overlap instances** in the current holmium-plate fixture (stone, ice), not the 3 reported when the issue was filed — plausibly solver/routing drift since 2026-07-19, but the same root-cause class; the fix is general (any tile, any item) rather than special-cased to 2 or 3 instances.

## Fix

`merge_output_rows`'s east extension already bridges around *other items'* south columns (`blocked_columns`) via an underground pair. It now treats any tile Step 4-6 (ghost-routed lanes) already claimed the same way — bridge underground around it instead of colliding.

The one wrinkle: when the very first tile of that bridge needs converting (the row's own last belt tile), that tile lives in the caller's immutable `row_entities` slice, which this module never sees — it's merged into the final layout only after `route_bus_ghost` returns (`layout.rs`). Rather than plumb mutable access through, a new `row_tile_overrides: FxHashSet<(i32,i32)>` on `GhostRouteResult` reports the tile, and `layout.rs` folds it into its **existing** splitter-eviction filter over `row_entities` (previously scoped to bus splitters only) — so the row's now-stale tile is dropped rather than double-placed.

## Folded from review

The adversarial review caught a real latent bug in the first version of this fix, and two nomenclature/cleanliness notes. All three folded:

1. **Latent self-overlap** (the substantive finding): the `x == rw` branch placed the UG exit at `run_end + 1` without checking it against `col_x` (the south-column reservation). When `col_x` sits exactly one tile past `rw` — a tight `min_merge_x`, e.g. this is the only/first item processed, not the fulgora fixture's own shape today but a real trap for a future single-item dual-fate case — the exit and the south column's own top tile land on the same coordinate. Reviewer reproduced it with a probe on the fixture the new regression test already builds. Fixed with a pre-pass before `merge_x` is finalized: mirror the main loop's blocked-run walk (bounded only by UG reach, since `col_x` isn't known yet) and widen `merge_x` just enough that every row's exit — bridged or not — lands strictly before its own south column. A no-op for every row whose `row_width` tile isn't blocked, i.e. byte-identical for every layout but this one; verified by temporarily disabling the pre-pass and confirming the strengthened test (below) fails exactly as predicted, then re-enabling it and confirming green.
2. **Test strengthened**: `test_merge_output_rows_bridges_pre_occupied_row_exit` now asserts no duplicate `(x, y)` among *all* emitted entities (not just the originally-blocked tile) — the assertion shape the reviewer's finding needed to be caught by the test in the first place.
3. **`merger_rate_and_tier` inlined back into `merge_output_rows`** — it had exactly one caller (itself) after an earlier design iteration removed the second call site; the split no longer bought anything.
4. **RFC decision-log nomenclature fixed**: the new `belt-junction` errors are a sibling of, not literally the same check as, the belt-loop/underground-belt noise (see the Scope bullet above).

## Scope

- **In:** the entity-overlap geometry fix; extending `fulgora_scrap_sorter_mechanism_present` to assert zero entity-overlap errors; a focused unit test for the new bridging path (`test_merge_output_rows_bridges_pre_occupied_row_exit`); an RFC decision-log entry recording exactly what is and isn't fixed.
- **Out — AM3 single-exit-bus cluster (holmium-plate row, zero output inserters/belt):** confirmed via `git stash` A/B comparison against the pre-fix baseline to be **100% pre-existing and untouched by this fix** — a third, unrelated defect in row/template generation upstream of ghost routing, in a completely different subsystem from the Step 7 merger-collision fix above.

  **Root cause, precisely**: `row_kind()` (`crates/core/src/bus/placer.rs` ~561) classifies any row with `solid_inputs == 0, fluid_inputs == 1, machine_size < 5` as `RowKind::OilRefinery` (dispatched to `templates::fluid_only_row`) without checking `solid_outputs` — that template is hard-coded fluid-in/fluid-out (its own doc comment: "fluid inputs sit at the NORTH edge... fluid outputs sit at the SOUTH edge"; the motivating case is lubricant, light-oil→lubricant on chemical-plant) and has no solid-output code path at all. holmium-plate (holmium-solution [fluid] → holmium-plate [**solid**] on assembling-machine-3) hits this branch and silently gets zero output inserter/belt — apparently the first Fulgora recipe shaped this way, so the gap was never exercised before.

  The straightforward-looking fix — add a `solid_outputs.is_empty()` guard and route non-empty-solid-output rows to `RowKind::FluidInput` instead — **does not actually work**: I traced `templates::fluid_input_row` (`templates.rs:2316`) and it unconditionally places a solid-**input** belt+inserter at y+2/y+3 from a mandatory `solid_item`/`solid_rate` parameter, alongside the machine-row and the output inserter+belt at y+7/y+8 that already exist and would be exactly what holmium-plate needs. holmium-plate has zero solid inputs, so redirecting to `FluidInput` would place a bogus, disconnected solid-input inserter rather than fix anything. A real fix needs a genuinely new template variant — fluid-input-only (no solid input stage) with the existing solid-output tail — not a one-line dispatch change, and `row_kind()`/the template functions are shared across every non-Fulgora recipe too, so the safe surface for that work is larger than this PR's scope. Left tracked on #309 rather than attempted here under time pressure against a heavily-shared, heavily-tested subsystem.
- **Out — the deeper dual-fate flow-rate split:** this fix resolves the *geometric* collision only. The `ret` lane is still sized off the row's full production (`lane_planner.rs`'s `item_to_rate` sums the whole row, not true downstream demand), and `check_input_rate_delivery` still warns 0.0/s delivered against the item's real demand (0.05/s stone, 0.025/s ice) — **byte-identical before and after this fix** (verified via the same A/B comparison), confirming this was already broken and is untouched. Real resolution needs the physical flow-split (a splitter at the sort point) the RFC's decision log already calls for.
- **Side effect, verified pre-existing not new:** fixing the overlap surfaced 4 new `belt-junction` "head-on" errors (the row's east-facing exit now visibly abuts the ret belt's west-facing bend — previously masked because the two colliding entities' directions weren't both visible to the checker when they shared one tile). Confirmed via the same A/B diff that this defect (the underlying geometry) is pre-existing, not newly introduced. Nomenclature: `check_belt_junctions` (head-on) is a *separate check* from `check_belt_loops`/`check_underground_belts` — these new errors are a **sibling** of the belt-loop/underground-belt false positives the sushi region already generates (~100 instances, per the test's own comment), not literally the same check firing. The test deliberately asserts zero entity-overlap only, not zero belt-junction, matching that existing scope boundary.

## Verification

- [x] `SPAGHETTIO_ZONE_CACHE_PATH=$(pwd)/crates/core/data/sat-zones-ci.bin cargo test --manifest-path crates/core/Cargo.toml` — full pinned suite, 951 passed, 0 failed (837 lib + 61 e2e + rest), zone-cache pin restored after (`git checkout --`).
- [x] `cargo clippy -p spaghettio_core -- -D warnings` (exact pre-commit hook invocation, lib target) — clean. (A broader `--all-targets` run fails on pre-existing, unrelated lints in `e2e.rs`/`layout.rs`/`power_wires.rs` doc comments — confirmed via the same A/B stash comparison to be present on unmodified `origin/main` too, not introduced here.)
- [x] `cargo build --features wasm -p spaghettio_core` + `cargo build --workspace` — clean; no wasm-bindgen-exposed signatures touched.
- [x] Entity-level snapshot inspection: dumped exact tile contents at every previously-overlapping coordinate before/after via a scratch example (`crates/core/examples/repro_309_fulgora.rs`, gitignored), confirming zero shared tiles and tracing the full `ret` belt path.
- [x] `git stash` A/B against the pre-fix baseline for every claim above (belt-junction new-vs-preexisting, AM3 pre-existing, input-rate-delivery byte-identical).
- [ ] Browser eyeball — **not applicable**: this fixture requires `allow_recycling`/scrap-chain netflow options that aren't exposed via the web app's URL/UI (`allow_recycling` doesn't appear anywhere in `web/src/`), only reachable via the test harness's `solve_netflow_with_options`.

## Deviations from intent

None from the assigned scope. The AM3 cluster and the flow-rate-split gap were investigated (root-caused to the same or a different subsystem, precisely as asked) and explicitly left tracked rather than folded in, per the task's own scope guard.

## Models / contracts touched

`GhostRouteResult` gained one field (`row_tile_overrides`); `merge_output_rows`'s signature gained two params (`existing_tiles`, `row_tile_overrides`) — both `pub(crate)`/internal to `bus/`, no public API change. Decision log updated in `docs/rfc-fulgora-scrap.md`.

https://claude.ai/code/session_01Q7TFG8si9QxRz6FrMmMnrp

### DIFF
--- a/crates/core/src/bus/ghost_router.rs
+++ b/crates/core/src/bus/ghost_router.rs
@@ -81,6 +81,15 @@ pub struct GhostRouteResult {
     /// the old path made retries fire only under a trace guard. The trace
     /// event is still emitted for the debugger; this is the control-flow copy.
     pub cap_coords: Vec<(i32, i32)>,
+    /// Tiles a routed bus entity claims that used to belong to a row
+    /// entity (`bus/placer.rs`'s `row_entities`, which this module never
+    /// sees — it's merged into the final layout by the caller). Populated
+    /// by `output_merger::merge_output_rows`'s row-exit bridging (RFC
+    /// Fulgora dual-fate items, #309): the row's now-stale tile there
+    /// must be dropped rather than double-placed alongside the new one.
+    /// `bus/layout.rs` folds this into its existing splitter-eviction
+    /// filter over `row_entities`.
+    pub row_tile_overrides: FxHashSet<(i32, i32)>,
 }
 
 /// A spec for one connecting belt run.
@@ -3304,6 +3313,21 @@ pub fn route_bus_ghost(
         true
     });
 
+    // Snapshot of every tile Step 4-6 (+ the `retain` cleanup above) has
+    // already claimed, threaded into every `merge_output_rows` call below
+    // as its `existing_tiles` param — a dual-fate byproduct's `ret` belt
+    // (an intermediate lane's producer connection, #309) can land on the
+    // exact row-exit tile Step 7's east extension is about to start from.
+    // Taken once here; Step 7's OWN entities (target/surplus/voider merges
+    // below) don't need to be visible to each other through this set —
+    // `blocked_columns` already covers that ordering within this step.
+    let existing_tiles: FxHashSet<(i32, i32)> = entities.iter().map(|e| (e.x, e.y)).collect();
+    // Tiles `merge_output_rows`'s row-exit bridging overrides — a row
+    // entity the caller must drop before merging `row_entities` with this
+    // function's returned `entities` (see
+    // `GhostRouteResult::row_tile_overrides`).
+    let mut row_tile_overrides: FxHashSet<(i32, i32)> = FxHashSet::default();
+
     // -------------------------------------------------------------------------
     // Step 7: Merge output rows for final products
     // -------------------------------------------------------------------------
@@ -3360,6 +3384,8 @@ pub fn route_bus_ghost(
                 merge_x_cursor,
                 &blocked_columns,
                 ctx,
+                &existing_tiles,
+                &mut row_tile_overrides,
             );
             blocked_columns.extend((item_merge_x - output_rows.len() as i32)..item_merge_x);
             merge_x_cursor = item_merge_x + 1;
@@ -3454,6 +3480,8 @@ pub fn route_bus_ghost(
             merge_x_cursor,
             &blocked_columns,
             ctx,
+            &existing_tiles,
+            &mut row_tile_overrides,
         );
         blocked_columns.extend((item_merge_x - output_rows.len() as i32)..item_merge_x);
         merge_x_cursor = item_merge_x + 1;
@@ -3542,6 +3570,8 @@ pub fn route_bus_ghost(
             merge_x_cursor,
             &blocked_columns,
             ctx,
+            &existing_tiles,
+            &mut row_tile_overrides,
         );
         blocked_columns.extend((item_merge_x - output_rows.len() as i32)..item_merge_x);
         merge_x_cursor = item_merge_x + 1;
@@ -3968,6 +3998,7 @@ pub fn route_bus_ghost(
         warnings,
         surplus_exits,
         cap_coords,
+        row_tile_overrides,
     })
 }
 
--- a/crates/core/src/bus/layout.rs
+++ b/crates/core/src/bus/layout.rs
@@ -840,6 +840,7 @@ fn layout_pass(
     // stream) so `run_layout_with_retry_inner` detects caps whether or not a
     // trace collector is active — see that function and package #3.
     let cap_coords = ghost_result.cap_coords;
+    let row_tile_overrides = ghost_result.row_tile_overrides;
     crate::trace::emit(crate::trace::TraceEvent::PhaseTime {
         phase: "ghost_routing".to_string(),
         duration_ms: t_ghost.elapsed().as_millis() as u64,
@@ -876,6 +877,11 @@ fn layout_pass(
             }
         }
     }
+    // Dual-fate byproduct row exits (RFC Fulgora, #309): `merge_output_rows`'s
+    // row-exit bridging spliced a UG entrance over the row's own last belt
+    // tile so the row's stale plain belt there doesn't double-place
+    // alongside it — same eviction mechanism as the splitter case above.
+    bus_occupied.extend(row_tile_overrides);
     let row_entities: Vec<PlacedEntity> = if bus_occupied.is_empty() {
         row_entities
     } else {
--- a/crates/core/src/bus/output_merger.rs
+++ b/crates/core/src/bus/output_merger.rs
@@ -13,37 +13,6 @@ use crate::bus::balancer::splitter_for_belt;
 use crate::bus::placer::RowSpan;
 use crate::bus::stacking_ctx::StackingCtx;
 
-/// Total rate + belt/underground tier for an item's merge cascade — split
-/// out so a caller can compute it once and reuse it, though today only
-/// [`merge_output_rows`] itself calls it.
-pub(crate) fn merger_rate_and_tier(
-    output_rows: &[usize],
-    item: &str,
-    row_spans: &[RowSpan],
-    max_belt_tier: Option<&str>,
-    ctx: &StackingCtx,
-) -> (f64, &'static str, &'static str, u32) {
-    use crate::bus::balancer::underground_for_belt;
-    use crate::common::{belt_entity_for_rate_stacked, ug_max_reach};
-
-    let total_rate = output_rows.iter()
-        .map(|&ri| {
-            if ri >= row_spans.len() {
-                0.0
-            } else {
-                row_spans[ri].spec.outputs.iter()
-                    .filter(|o| o.item == item)
-                    .map(|o| o.rate * row_spans[ri].machine_count as f64)
-                    .sum::<f64>()
-            }
-        })
-        .sum::<f64>();
-    let belt_name = belt_entity_for_rate_stacked(total_rate * 2.0, max_belt_tier, ctx.for_item(item));
-    let ug_name = underground_for_belt(belt_name);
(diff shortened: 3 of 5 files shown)


storkme (commented):
Verified the load-bearing claims via a direct A/B probe against `holmium-plate @ 0.25/s` (probe source + pre/post fix outputs archived locally):

- Entity-overlap: pre-fix 2 (at `(29,14)` and `(29,17)`) → post-fix 0. Fix works on the target fixture.
- 4 new `belt-junction` head-on errors: confirmed at `(28-29, 14)` and `(28-29, 17)` — the exact y-rows where the previous overlaps sat, so the "same geometry, previously masked because the two colliding entities' directions weren't both visible to the checker" framing is precise.
- AM3 cluster (`output-belt: 1, inserter: 1, belt-connectivity: 1`) is byte-identical pre- and post-fix. Correctly out-of-scope.
- `row_tile_overrides` is populated from exactly one site (`output_merger.rs:143`), gated on `existing_tiles.contains(&(x, out_y))`, so it stays empty for ordinary bus layouts and the `bus_occupied.extend` is a no-op there. No accidental eviction blast radius.

Pinned suite (`SPAGHETTIO_ZONE_CACHE_PATH=...ci.bin cargo test`): 0 failed. Clippy per the pre-commit invocation (`cargo clippy -p spaghettio_core -- -D warnings`): clean.

---

### Concern — latent overlap when `col_x == run_end + 1` (`output_merger.rs`, the `x == rw` branch, ~L127-176)

Reachable exactly for the shape the added regression test constructs (single-row, single-item, `min_merge_x = 0`, so `col_x = rw + 1`).

Trace with `existing_tiles = {(rw, out_y)}`, `col_x = rw + 1`:
1. `x = rw`, blocked. Inner while stops immediately because `run_end + 1 < col_x` is `rw+1 < rw+1` — false. So `run_end = rw`.
2. `x == rw` branch pushes a UG entrance at `(rw-1, out_y)` and a UG exit at `(run_end + 1, out_y) = (rw+1, out_y) = (col_x, out_y)`.
3. `x = rw + 2`, loop exits.
4. The south-column loop at line 192 then unconditionally pushes a south-facing `transport-belt` at `(col_x, out_y)` — same tile as the UG exit just placed.

Verified by adding a probe module that calls `merge_output_rows` on the same fixture as `test_merge_output_rows_bridges_pre_occupied_row_exit` and checks for duplicate tiles among the returned entities:

```
dup at (11, 2): [
  "underground-belt dir=East io=Some(\"output\")",
  "transport-belt  dir=South io=None",
]
```

The existing test misses it because its assertion is `entities.iter().all(|e| (e.x, e.y) != (row_width, out_y))` — it checks the *blocked* tile, not the tiles the fix newly places. A stronger assertion (no duplicate `(x, y)` among `entities`) would catch it.

Doesn't fire on the fulgora fixture: for the surplus/voider loops, `merge_x_cursor` has already been bumped past the widest row by the target-items loop (`ghost_router.rs` Step 7), so `col_x` sits far east of `rw` — the branch fires but `run_end` extends and the UG exit lands well before `col_x`. It's a trap for any future single-item / small-`min_merge_x` dual-fate case, or if the fix's pathway were exercised via the target-items loop with a lone external output whose producer row also emits a ret belt.

Two reasonable ways out: (a) stop the run-extension one tile short when the next iteration would place a UG exit at `col_x` (fall through to a plain belt at `col_x - 1` with the belt-to-south handoff intact); or (b) tighten the regression test to assert no duplicate tiles among the emitted entities and let the failure force the geometry fix.

---

### Note — `merger_rate_and_tier` split with one caller (`output_merger.rs:19-45`)

Helper is split out but has one caller today, and the doc comment says so ("though today only [`merge_output_rows`] itself calls it"). Either inline it, or note where the second caller is coming from. If it's speculatively split for a follow-up refactor, worth mentioning that follow-up in the doc.

### Note — nomenclature in `docs/rfc-fulgora-scrap.md` decision log (~L680)

`check_belt_junctions` (head-on) is a separate check from `check_belt_loops` and `check_underground_belts` — they're all "general belt validator noise" in the informal sense, but calling the new head-on errors the "SAME class" as belt-loop/underground-belt reads as if they're from the same check function. Worth being explicit (e.g. "head-on belt-junction errors — a sibling of the belt-loop / underground-belt false positives the sushi region already generates") since the next reader may only skim.

PR #645, opened 2026-08-14
chore(compaction): #632 A2 — delete the compaction/fold stack (owner call)

storkme:
## Intent

Delete the compaction/fold stack (`bus::compaction`, the `compact_layout`/
`fold_layout` `LayoutOptions` flags, and the `CompactTransform`/
`FoldTransform` `LayoutTransform` wrappers) — item A2 of the churn-reduction
campaign, issue #632. This was previously a park (owner-reserved), converted
to a kill by an explicit owner call: the session goal for this unit was
"complete B5 and A2 as described." Recorded here and will be posted to
#632's tracking comment by the coordinator (not this PR, per the campaign's
process — this PR does not touch the issue).

The underlying relocation research (topology-preserving dense repacking /
snake folding) never shipped past three falsified attempts across three
RFCs — RFC-057, RFC-058, and RFC-064 Phase 3 — each independently concluding
the transport-layer cost outweighed the density win. `compact_layout`/
`fold_layout` shipped opt-in, default false, and stayed unused; deleting them
removes 6,401 lines of dead-weight source plus every test whose only subject
was the transform, with the produce→validate→measure→verdict→rank framework
(`bus::candidate_runner::run_candidate_field`) preserved intact because the
live RFC-068 celldb campaign consumes it.

## Scope

**In:**
- `crates/core/src/bus/compaction.rs` deleted whole (6,401 lines). Its
  `ProductionEdge`/`ProductionSignature`/`ProductionMachine`/
  `ProductionBoundary`/`RATE_SCALE` cluster (+ private canonicalization
  helpers) relocated into `bus/transit.rs` first — its only surviving
  consumer, doc comments preserved.
- `bus::candidate_runner`: `CompactTransform`, `FoldTransform`, the
  `LayoutTransform` trait, `TransformOutcome`, `CandidatePlan::with_transform`,
  `compose_chain`, and the transform slot/plumbing throughout
  `CandidatePlan`/`run_plan`/`run_candidate_field` are gone. The framework
  itself — `run_candidate_field`, `CandidatePlan`, `FullSelectionCandidate`,
  `EvaluatedCandidate`, `FieldResult`, objective/verdict integration —
  survives unmodified in shape; `crates/core/tests/celldb_template.rs` calls
  it with zero code changes.
- `bus::layout`: `compact_layout`/`fold_layout` `LayoutOptions` fields, their
  `build_bus_layout` branches, `FOLD_SEARCH_ENTITY_THRESHOLD`, the
  `PlacedMachineSignature` test, and the stale-flag guard in
  `build_rotation_aware_row_layout[_selected]`.
- `bus::decomposition_search`'s inner-`LayoutOptions` struct literal (its
  `compact_layout`/`fold_layout: false` fields) and `bus::layout`'s now-
  orphaned `cover_unpowered` helper (its only caller was inside
  `compaction.rs`) — both scope surprises the original menu didn't name,
  found by `cargo build` after the module went away.
- `crates/wasm-bindings/src/lib.rs`: the `compact_layout`/`fold_layout`
  `Option<String>` params and wiring, dropped from all three
  `#[wasm_bindgen]` layout entry points — a breaking wasm JS API change.
- Web surface (six-site sweep, per the project's own "web controls need six
  sites" lesson): `web/src/state.ts` (`FormState` fields, URL extras keys,
  read/write sites), `state.test.ts`, `engine.ts`, `workers/engine.worker.ts`,
  `ui/sidebar.ts` (both checkboxes + every read/write/wire site).
- Test files: `crates/core/tests/fold_layout_knob.rs` deleted whole (329
  lines). `candidate_runner.rs`: the two fold/compact parity tests deleted;
  the two synthetic-transform ranking/verdict tests ported to synthetic
  `DecompositionCandidate` base producers (what they exercise — "a candidate
  whose score/verdict differs from the incumbent's" — never depended on the
  transform abstraction). `objective.rs`: the fold-ranks-above-native
  calibration test deleted; `measurement_is_total_or_it_refuses` ported off
  `compact_layout` onto a plain build. `connectivity_parity.rs`: five
  compaction-admission-filter tests deleted (see Verification below for
  which K65-1/K65-2 coverage that costs).

**Scope surprise not in the #632 A2 menu**: `crates/core/tests/
cell_composition.rs` carried its own ~2,200-line block of RFC-057
compaction/fold/manifold-routing research probes, including TWO
non-`#[ignore]` tests `cargo test` runs by default —
`mil5_multifold_holds_and_preserves_lanes` and
`rfc057_island_placement_keeps_terminals_disjoint`. Neither was in the
original dependency map. Found by `cargo build --test cell_composition`
after `compaction.rs`'s deletion produced 38 compile errors. `bus::
compaction`'s "local manifold routing"/"recipe cluster placement"/"rigid
island" subsystem (`plan_local_manifolds`, `build_local_manifold_graph`,
`route_local_manifold_edges`, `legalize_manifold_routes`,
`place_distributed_local_manifold_nodes`, `build_manifold_nets`,
`place_recipe_clusters`, `extract_rigid_islands`, ...) — distinct from
`CompactTransform`/`FoldTransform` — had **zero live `src/` consumers**
(grep-verified); it existed purely to be exercised by this file's probes,
so it went with them rather than being carved out as a surviving module.
Kept everything interspersed in the same region that wasn't compaction-
dependent: `export_mega_pu_for_sim`, `export_mega_chem_for_sim`,
`export_mega_chain_for_sim`, `probe_mega_cells`, `probe_mil5_errors`,
`export_chem5_for_rebless`. Every RFC-058/RFC-064-Phase-3 test past that
region is untouched.

**Out:** No changes to `verdict.rs`'s `never_worse`/`MatchTier`/`Policy`
machinery, `objective.rs`'s scoring logic, `bus::transit`, `bus::
row_rotation`, `connectivity.rs`, or any of `bus::cells` — all confirmed
live and untouched beyond doc-comment repoints.

**Size**: net −10,093 lines (33 files, +548/−10,641) — overwhelmingly
deletion, so the "split above ~400 added lines" norm doesn't bite in the
direction it's calibrated for; the ~548 added lines are almost entirely
relocated code (`transit.rs`) and RFC decision-log/followups-doc closure
notes, not new claim surface.

## Verification

1. `cargo test --manifest-path crates/core/Cargo.toml` — **1199 passed, 0
   failed, 83 ignored** (one clean invocation, summed across the 23 test
   binaries in that single run).
2. `cargo clippy --workspace -- -D warnings` — clean.
3. `wasm-pack build crates/wasm-bindings --target web --out-dir <abs path to
   web/src/wasm-pkg>` — succeeded (release, ~3 min).
4. `cd web && npx tsc --noEmit` — clean. `npm run test` — 3 files, 41 tests,
   all passed.
5. Post-deletion grep receipts: `grep -rn "bus::compaction\|CompactTransform\|
   FoldTransform" crates/ web/` returns only my own repointing comments
   (explicitly noting the 2026-08-14 deletion) inside `verdict.rs`,
   `connectivity.rs`, `common.rs`, `validate/mod.rs`, `bus/candidate_runner.rs`,
   `bus/layout.rs`, and the three edited test files — zero live code
   references remain.

Docs sweep: swept `grep -rn "compact_layout|fold_layout|search_snake_fold|
compact_validated_geometry|CompactTransform|FoldTransform|LayoutTransform"
docs/ crates/ web/ CLAUDE.md .github/` (83 hits, 18 files) plus a broader
`compaction` sweep for symbol-adjacent references. RFC-057, RFC-064, and
RFC-065 (the three RFCs whose own test/code surface changed) each got a
dated decision-log entry recording the deletion, owner-call provenance,
what specifically survives vs. is moot, and the revival path. `docs/
rfcs.md`'s RFC-057 row (was making a false present-tense "ships opt-in"
claim), `docs/status.md`, `docs/rfc064-phase2-followups.md`, `docs/
snake-fold-followups.md`, `docs/spare-compute-playbook.md` (its job grids
literally specify `compact=1`/`fold=1` — marked VOID inline), `docs/
compaction-retro-2026-07.md`, `docs/validator-trust.md`, and `docs/
rfc-058-band-packing.md`/`docs/rfc-062-multi-target-outputs.md` (each had
one dead style-example pointer to a deleted test) all got dated fixes.
Left untouched, confirmed unrelated by content check: `web/src/ui/
inspector.ts`'s `COMPACT_DIR` (a different sense of "compact"),
`crates/core/examples/probe_hotspots.rs`'s now-permanently-dead
`"compact-lane:"` segment-prefix match arm (harmless, debug-only probe
script, left as a known minor deviation — see below), `trace.rs`'s
`InterRowBand` doc comment (live consumer is an e2e stress scoreboard, not
compaction), and RFC-055/056/059/063/lane-aware-routing/throughput-
priority-merges/archive/* (generic "compaction" word usage in adjacent
RFCs or an unrelated SAT/sideload sense — verified with a targeted grep
for the actual deleted symbol names, zero hits).

## Deviations from intent

- **`skipped_transforms` removed from `EvaluatedCandidate`/plumbing beyond
  what the brief literally named.** Since no transform can ever run
  anymore, this field would always be empty — removed as vestigial
  "transform plumbing" rather than left as a permanently-dead pub field.
  No consumer referenced it (grep-verified).
- **`crates/core/examples/probe_hotspots.rs` left with one dead match arm**
  (`"compact-lane:"` segment-prefix classification) rather than edited.
  It's a debug/hotspot-classification probe script, not part of the
  shipped pipeline; the arm is harmless (will just never match) and
  touching it felt like scope creep beyond "delete the compaction/fold
  stack." Flagging in case the campaign wants it cleaned up separately.
- **Commit history has one cosmetic slip**: the first commit's message
  was written for what became the *second* commit's content (a `git add`
  pathspec on an already-staged deleted file errored and silently
  short-circuited a broader `add`, so the first commit only picked up the
  3 files staged earlier in the session). Both commits' actual diffs are
  correct and match their real content; only the label on the first one
  is off, noted in the second commit's message.
- Everything else matches the brief as given.

## Models / contracts touched

- **`docs/file-reference.md`, `docs/ghost-pipeline-contracts.md`**: neither
  mentioned `compaction.rs`/`candidate_runner.rs`'s transform layer, so
  neither needed edits (checked).
- **`bus::candidate_runner`**: contract narrows — `CandidatePlan` no longer
  carries a transform chain; `run_candidate_field` always verdicts field
  candidates at `MatchTier::Count`. No current consumer (celldb) used
  `MatchTier::Provenance`/`Positional` via a transform, so this is a
  no-op in practice, documented in the module's new header.
- **RFC-057, RFC-064, RFC-065 decision logs**: each amended in place (see
  Verification above) — this is itself "the doc update landing in the same
  PR" for those contracts.


storkme:
## Second-opinion review adjudication

Review took 3 attempts to produce output — the first two (re-run via `gh run rerun`) both hit the reviewer's own 6,000,000-token ceiling exploring this PR's diff on all 3 K-passes and returned "reviewer malfunction, not a clean bill of health" per the workflow's own fail-on-degraded design. The third attempt succeeded and produced 4 findings. Adjudicating each:

**Finding 1 (`cell_composition.rs` over-deletion) — ACCEPTED, real bug.** `export_chem5_for_rebless` was swept into the deletion by mistake (it never touched `bus::compaction` — confirmed via `git show origin/main`). `probe_pole_connectivity_census`'s own doc comment states its purpose is measuring the *normal* pipeline, and one of its two compaction-comparisons was computed and then explicitly discarded in the original (`let _ = (cd, cn);`). Both restored — `export_chem5_for_rebless` verbatim, `probe_pole_connectivity_census` with the compaction-specific measurement stripped and the surviving normal-pipeline/composed-vs-repaired measurements kept.

**Finding 2 (`production_signature_is_order_independent` lost) — ACCEPTED, real gap.** `ProductionSignature`'s only direct correctness pin lived in `bus::compaction`'s own test module and didn't move with the code. Ported verbatim into `transit.rs`'s existing test module.

**Finding 3 (`MatchTier::Provenance`/`Positional` now dead-but-inviolate) — accepted as a disclosed tradeoff, no code change.** Correct observation, already the intended shape: the brief scoped `verdict.rs` as KEEP beyond doc-pointer fixes, since `MatchTier` is generic infrastructure a future transform could still use. Deleting enum variants from a public API is a separate decision this PR wasn't authorized to make.

**Finding 4 (bookmarked `?compact=1`/`?fold=1` URLs silently degrade) — accepted as already covered, light touch added.** Already stated in this PR's body; added a source comment in `state.ts` next to `directInsertion` so a reader of the code sees it too, not just the PR description.

Absorption commit: 80e2c55c. Verification re-run clean: `cargo test --manifest-path crates/core/Cargo.toml` 1200 passed / 0 failed / 85 ignored (one invocation), `cargo clippy --workspace -- -D warnings` clean, `cd web && npx tsc --noEmit` clean.

PR #646, opened 2026-08-14
fix(validate): #632 B5 step 2 — belt_flow takes the lane-throughput dispatch

storkme:
## Intent

Step 2 of the B5 arbitration plan (#632): the one-line dispatch swap — `belt_flow::check_lane_throughput` replaces `belt_structural`'s at the validate dispatch — plus the adjudication of every corpus red the truth-telling check surfaces. **Stacked on #643** (the walker's three false-positive fixes); retarget to main when #643 merges.

## The enumeration arc (measured at each stage)

- Swap with no walker fixes: **26 reds**, including the RFC-059 big-pole teeth test shipping the 0.51/s-measured layout (the splitter-cap false positive refusing the good arm — the #519-inversion the sim-anchored pin exists to catch).
- After #643's three fixes: **7 reds**, all one family.
- After the adjudications here: **0** — suite 1242/1242.

## The adjudications (all cite #644, the engine-defect tracking issue)

| Fixture | Change | Receipt |
|---|---|---|
| stress_ec_30s | max_errors 0→140 (lane-throughput) | sim 92.1% delivered (post-lift bank) + meter 91.9% |
| stress_ec_60s_red | 1→218 | sim 90.7% delivered |
| stress_ec_40s | 13→201 (13 belt-dead-end unchanged +188 lane) | same family by construction |
| 30s_decomposed | both arms 0→140 | same layout family |
| tier5 PU am3 | exact pin {lane-throughput: 70} | meter 85.6% of plan (1.712/2.0), uniform choke |
| partition scoreboard | PU@2-ore-red (0,3)→(70,66); AC@5 kept (4,4) | AC@5's 4 covers recorded CI nondeterminism — local-0 tightening would re-arm the flake |
| stacking_ec_60s headline | arms tolerate lane-throughput only | S=1 at 60/s is where the deficit binds (16.1–16.8/s on 15/s lanes); any other category still panics |

The tier5 warnings golden moves to a new `assert_warnings_golden_allow_errors` variant — reserved for fixtures carrying an adjudicated exact error pin, never for silencing unadjudicated errors.

## What the swap buys

The dispatched lane check now models #519-style consumption and convergence: the corpus's real high-rate trunk deficits (silently shipped for months, meter/sim-confirmed) are Errors with named tiles, and the ec22-class "validator-clean at 75–92% of plan" state is no longer representable. The `belt_structural` twin is step 3's deletion.

Trust-table lane-throughput row rewritten in the same PR (the "which of its two implementations to believe is unsettled" caveat resolves).


storkme:
Round 6 absorbed:
1. **K1-1 flake tolerance [3/3] — ACCEPTED**: `decomposed <= pooled + 2` with the rationale in place.
2. **Decomposed arms category-scoped [nit] — ACCEPTED**, matching tier5's guard: any non-#644 category fails loudly on either arm.
3. **rate-stamp-semantics + meter handoff still name belt_structural as dispatched [minor] — ACCEPTED**: both bracketed with the settled state and pointer to the trust table (the historical analysis stands as record).
4. **Ceiling-at-measured absorbs no upward flake [minor] — DECLINED with a receipt**: every run today (nine full-suite/single-test invocations across two machines' worth of cache states) reproduced the identical counts (140/218/201/70) — these fixtures are not in the AC@5 nondeterminism class, and headroom above measured would erode the file's tighten-downward convention. If CI ever flips one, the failure message names #644 and the adjudication path.

storkme:
Round 7 adjudicated — one accepted, two closed as re-litigation of adjudicated ground:

1. **Ceiling discontinuity across the from-ore family [3/3] — the demanded verification already exists, the doc gap was real.** The full suite (nextest, CI's required `rust` check) is green on every head of this PR including this one — 22s/23s genuinely clear under the stricter walker; the deficit family *starts* at 30/s. That physics is now stated at the 30s ceiling (4e7157d9) so the discontinuity reads as measured, not un-updated.
2. **"Ceilings become permanent floors" [2/3] — DECLINED, third consecutive round.** This is the file's own long-standing baseline convention (`max_warnings` has worked exactly this way for months); #644 is the auditable ledger of every number, and the tighten path is the same review culture that tightened 60→0 and 11→0 in this very file's history. No new evidence was raised.
3. **"+2 for K1-1 but zero ceiling headroom is inconsistent" [2/3] — DECLINED; the asymmetry is principled.** K1-1 differences two independent noisy counts (noise doubles, and a spurious failure MISATTRIBUTES regression between arms); a ceiling compares one count to its own observation — reproduced exactly across nine invocations today — and a hypothetical flake fails one test with a #644-labeled message whose adjudication path is one line. Round 6 declined this with the stability receipt; nothing new was added.

Per the campaign's convergence rule (#639 round 4 precedent): rounds are now circling adjudicated ceilings philosophy with no new facts — merging on green.

PR #647, opened 2026-08-15
chore(validate): #632 B5 step 3 — delete belt_structural's lane walker; share the survivor

storkme:
## Intent

The B5 arbitration's close (#632; steps 1–2 = #643/#646, tracking issue #644): delete the losing walker and land the memoization promised in #646's review.

## Scope

- **Deleted** (~1,280 lines): `belt_structural::compute_lane_rates`, `check_lane_throughput`, `classify_belt_feeders`, `do_propagate`, `splitter_output_rates`, `LaneRates`, and their test section. Zero consumers since #646's dispatch swap (grep receipt: the one remaining mention was a doc-comment consumer list, cleaned).
- **Deleted with the model**: the two RFC-066 comparison probes (`probe_walkers.rs`, `probe_walker_shape.rs`) — their subject, the two-walker disagreement, no longer exists, and they fail the build once the walker is gone (`cargo test` compiles examples). RFC-066's port-or-delete rule applied; survivor-vs-ground-truth instrumentation is the meter (`crates/meter`) and the sim harness. **RFC-066's decision log carries the EXECUTED AND CLOSED entry** — its own sixth-pass consumer sweep served as this deletion's checklist, every item verified handled (models.rs/inserters.rs intra-doc links and prose were repointed in #646; mod.rs's plain-backtick list, the probes, and rate-stamp rule 2 are handled here).
- **Shared walker**: `validate()` computes `belt_flow::compute_lane_rates` once and passes it to new `check_lane_throughput_with` / `check_input_rate_delivery_with` variants (public wrappers unchanged) — the walk is the expensive part of both dispatched rate checks.
- Trust-table lane-throughput row closed out; three test-scoped A2-residue warnings in `objective.rs` swept.

## Verification actually run

- Full suite **1184/1184** (one clean invocation, 71s — the count and wall-clock reflect A2's merged deletions plus this PR's ~16 walker tests).
- `cargo clippy` clean on the lib; pre-commit clippy on staged files clean.
- Deleted-symbol sweep: remaining mentions are decision-log history (RFC-066, rate-stamp-semantics' bracketed record, the meter handoff's bracketed note) — each states the deletion inline.

## Deviations

The `objective.rs` residue sweep is A2 collateral rather than B5 scope — three one-line test-scoped warnings from the compaction deletion, taken here because they sat in the build output of every run.


storkme:
Review absorbed (12c5eeb9):

1. **Over-reached test deletion [major] — ACCEPTED, all ten restored verbatim** from the pre-deletion commit, with a provenance comment. The bot's reading was exactly right: they were interleaved inside the deleted lane-test span, and `check_tap_splitter_priority` — a dispatched Error check — had been left with zero coverage. Suite is 1194/1194 with them back. This is the campaign's deletion-instrument lesson yet again, in its purest form: the span looked like "the walker's tests" and the grep that would have caught the live-check names inside it was never run.
2. **RFC-066's probe references [minor] — ACCEPTED, solved at the source**: the RFC now carries an EXECUTED-AND-CLOSED status header that covers every reproduce-command and table reference in its body in one place (deleted probes named, blame point given, surviving instruments listed), instead of seven scattered brackets.
3. **CPU-vs-latency [minor] — ACCEPTED as a wording fix**: the shared-walker comment now states the distinction; suite wall-clock receipts at merge time show no regression (71s, flat against the pre-memoization runs on the same tree).

storkme:
Round 2 absorbed (docs-only): status.md's 'still unarbitrated', the meter handoff's 'both exist', and rfc-inserter-sizing's three-sites follow-up all state the settled/deleted reality now. The WASM-path note needs no change by its own analysis — single-threaded, one walk is strictly better than two — and the comment already carries the CPU-vs-latency distinction; noted for whenever a WASM latency pass happens.

PR #648, opened 2026-08-15
fix(validate): #644 — UG crossing exits are not external-input sources (fourth walker FP class)

storkme:
# Intent

First PR of the #644 fix campaign — and it re-scopes the campaign. Probing the flagged "choke tiles" at tile level (the step the B5 adjudication skipped) found a **fourth walker false-positive class**: UG crossing exits counted as external-input graph sources. This PR fixes it and re-adjudicates everything downstream of the artifact.

## The bug

`belt_flow`'s external seeding treats any belt tile with no upstream surface feeder as a graph source. A UG crossing **exit** has no surface feeder — its flow arrives through the pair, via the topo sort's inheritance from behind the paired entrance. Counting it as a source:

- **broke demand attribution** for the item (Σ over sources ≠ solver total, e.g. copper-ore 75 ≠ 45 on stress-ec30), which forced the even-split fallback → every **real** trunk under-seeded (9/s instead of 15/s);
- **double-counted at the exit** (seed + pair inheritance) → 18/s modeled out of a tunnel carrying 9/s in.

Everything downstream of a crossing then read over-cap. `SPAGHETTIO_LANE_WALK_STATS=1` before/after: copper-ore `sources=[(22,15), (22,8), (5,0), (4,0), (3,0)] Σ=75 consistent=false` → `sources=[(5,0), (4,0), (3,0)] Σ=45 consistent=true`.

## What this re-adjudicates

**Every swap-era lane-throughput corpus red was this artifact**: ec30 140→0, ec60-red 218→0, ec40 188→0 (its 13 belt-dead-end stay), decomposed both arms 140→0, partition PU@2 (70,66)→(0,0), tier5 70→0. The true lane rates on those layouts sit exactly **AT** cap — zero headroom. Ceilings tightened to hard zeros with `lane-throughput → 0` category pins (any reappearance fails loudly); the now-unreachable #646 ceiling scaffolding is folded away.

The same under-seeding fabricated IRD starvation reads: tier5 golden 32→13, tier4 11→7, quality_ec45 inline 14→2 — the 12 cleared there are exactly the "even-split seeding" class its own 2026-07-31 adjudication comment already suspected.

**What does NOT change**: the sim/meter deficits (92.1%/90.7% delivered on stress-ec30/60s-red, 85.6% produced on tier5 PU) are real and stay open on #644 — reattributed to **zero-headroom provisioning** (rows/trunks planned at exactly 100% of nominal per-lane capacity), which flow conservation correctly does not flag. That's the engine fix the rest of the campaign targets. Residual validator signal for the family: `input-rate-delivery` + `row-input-belt-margin` warnings.

`validator-trust.md`'s lane-throughput row and an RFC-066 decision-log retraction addendum record the re-attribution; `manifest_carries_per_category_validator_state` re-pointed to ec@10-from-ore (gear@10 now validates clean).

# Verification

- New regression tests, both mutation-checked in TDD order: `ug_exit_with_fed_pair_is_not_an_external_source` (RED pre-fix: head seeded 6.0 of 12.0, conservation broken through the pair) and `orphan_ug_exit_still_seeds_as_external_source` (the guard's boundary — an exit with no paired entrance keeps seeding).
- Tile-level inspection per the layout-change protocol: stress-ec30's flagged tap dumped entity-by-entity pre-fix ([4.5,4.5] in → [9.0,9.0] out of the UG pair) and post-fix (conserved [7.5,7.5] trunk-to-row).
- Full core suite green — verified by EXIT CODE (0; all 26 binaries; the first version of this claim came from an aborted-run grep, caught by the adversarial review), lib clippy clean, wasm-pack build clean.
- Warning-golden re-bless run across the whole suite: exactly two files drifted (tier4, tier5), both adjudicated above.
- **Shipped geometry DOES change on ec30/ec60-red** — the phantom errors had been steering candidate selection since #646; removing them flips both winners BACK to the 2026-08-07 post-lift bank layouts (entity-count-exact: 3369 and 4967), i.e. the exact artifacts the 92.1%/90.7% sim receipts were measured on. The sprawls current main ships are 3-day-old phantom-steered accidents, never sim-anchored. tier5/ec40 layouts are byte-identical (adversarial-review golden check). RFC-069 Phase-0 sims added: ec22 99.4% delivered (PASS), tier5-PU axis-declared 95.6% delivered (the meter 85.6% was an axis-mismatched export).

# Deviations

- The B5 adjudication's "the walker's flags predict the measured deficit" claim is retracted as right-for-the-wrong-mechanism; recorded in the RFC-066 addendum rather than rewriting closed history.
- AC@5 partition pin deliberately NOT tightened (Pool reads 0 locally) — the 2026-05-02 CI nondeterminism adjudication on that exact case stands; note added at the pin.


storkme:
Review absorbed in 2b304baa. Adjudication, finding by finding:

**Fixed (real catches):**
- **`behind != ug_out` mirror clause** — real. The guard now carries the topo-sort wiring's exact predicate including `behind != pos`, plus a comment stating the shared assumption (structural presence, not fed/same-item — deliberately identical on both sides so the skip fires iff inheritance fires). Unreachable today, mirrored by construction now.
- **status.md:48 tier4 "pins 11"** — real records-outlive-state; trued to 7 with the #644 note. The bot missed the same class one row down: tier5's "pins 32" → 13, fixed in the same commit.
- **S=1 stacking comment** — the *finding's mechanism* was wrong but it surfaced a real staleness: the "16.1–16.8/s on 15/s fast lanes" comment described the pre-fix walker's phantom readings. Fresh probe on this branch: **S=1 and S=2 both measure ZERO lane-throughput errors** (`S=1: 9 issues {W input-rate-delivery: 4, W row-input-belt-margin: 5}`). The tolerance block is retired — `validate()` must succeed again on both arms — and with it the basis for the "live Error-level anchor" objection: there is no surviving over-cap fixture, and validator-trust.md's "no live anchor" claim stands as written.

**Refuted with receipts (no change):**
- **Hard-zero pins are float-fragile / contradict AC@5's (4,4)** — the check flags only `rate > cap + 0.01` (`belt_flow.rs`, `check_lane_throughput_with`). These layouts' lanes sit exactly AT cap — 0.01 *below* the flag threshold by construction; no rounding wobble crosses a 0.01 gap. The #646-era ceilings/tolerances guarded counts of *above-cap* tiles flipping near thresholds (the 140-count regime), which no longer exists. AC@5's (4,4) is junction-era *layout* nondeterminism — different category, and its pin is deliberately left loose with a dated note in this PR. The epsilon receipt is now in the ec30 pin comment.
- **"the 60s PU fixture's pin is documented as 'NOT on CI'"** — no such text exists anywhere in this PR or the repo (`grep -rn "NOT on CI"` returns nothing); hallucinated.
- **Nit "no negative test confirming the inherited-only path still seeds"** — `ug_exit_with_fed_pair_is_not_an_external_source` asserts tail total == 12.0 == head total, which *is* the inheritance-path positive check: post-fix the exit's entire flow arrives by inheritance and the assertion fails if it doesn't.
- **CONTRIBUTING.md example drift** — the file does not exist in this repo.

**Deferred to RFC-069 (real question, not this PR's scope):**
- "zero-headroom doesn't explain PU's 14% vs EC's ~8%" — agreed this needs measurement, and it's exactly RFC-069 Phase 0: a 432k-warmup sim of `tier5-pu2-am3` is running now to anchor the meter's 85.6% and decompose the loss. This PR retracts a false error attribution; it does not claim the zero-headroom account is complete for PU.

storkme:
Round 2 absorbed in 816dfdc3.

**Fixed (real catches):**
- **Second stale tolerance block (1/3-pass finding — the best catch of this round)**: round 1 fixed the closure but missed the per-arm screening at the call sites, leaving a dead `non_lane1` assert under a stale "carries real lane-throughput errors" comment. Both arms' screening is retired; `run()` returns the layout only.
- **tier5 doc comment** still described the removed known-deficit ceiling — trued up.
- **Guard comment precision**: now states that `behind != pos` lives in the inheritance *wiring* (`behind_to_ug_outputs`), that the processing block carries no such clause, and that `build_ug_pairs` geometry keeps `behind == pos` unreachable — instead of "mirrors EXACTLY".
- **blueprint.rs fixture comment**: "needs more than one issue category" downgraded to "preferred" — the only asserted requirement is non-clean, as the review noted.

**Refuted with receipts:**
- **"removes the only positive lane-throughput anchor; the check could fire on nothing"** — the firing side is anchored at unit level, mutation-checked: `splitter_tiles_rate_at_splitter_tier` asserts the YELLOW variant *must still flag* ("16/s on one YELLOW lane must still flag over-cap") and `ug_input_tiles_rate_at_ug_tier` likewise pins its flagged direction. A corpus fixture manufactured to stay over-cap would be a synthetic defect kept alive for the test's sake; the unit anchors are the right instrument. The trust-table row already instructs treating any new corpus mass as a finding, not background.
- **"any lane landing between cap and cap+0.01 trips the hard zero"** — backwards: the comparison is strictly `rate > cap + 0.01`, so that interval is exactly the band that does NOT trip.
- **Item-blind guard as "the one real correctness hole"** — deliberate and now documented in place: structural presence is the predicate *shared with inheritance*; tightening only the seed side (item/fed checks) would create the de-seed-without-inherit divergence the guard exists to prevent. No engine layout, celldb entry, or balancer template routes external flow underground past an unrelated abutting belt; if one appears, `SPAGHETTIO_LANE_WALK_STATS=1` shows the missing source.
- **Nit "re-attribution depends on unproven premise about IRD warnings"** — the zero-headroom re-attribution rests on the sim/meter receipts plus tile-level rate verification ([7.5, 7.5] exactly at cap), not on IRD deltas; the IRD drops are separately explained by the under-seeding mechanism (seed-stats before/after in the PR body). The Phase 0 sims (ec22 boundary, tier5-PU) are running to further decompose the PU loss.

storkme:
Round 3 absorbed in 9df84e69, and this closes the re-litigation tail per the convergence rule (#639 round-4 precedent): three of the five findings restate ground adjudicated in rounds 1–2 with no new evidence.

**Fixed (real catch, 2/3-pass):** the pre-existing comment above the sources loop ("simpler and safer to seed every graph source independently") directly contradicted the guard below it — exactly the future-reader trap the review described. It now states the orphan-only rule and names the refuted reasoning.

**Closed as re-litigated (no new evidence):**
- *[critical] item-blind guard* — round 2's finding escalated, same content. One addition worth answering: "Phase 3 then inherits behind's foreign/zero rates" describes **pre-existing inheritance behavior** this PR doesn't touch; in that hypothetical topology the pre-fix walker was MORE wrong (foreign inheritance *plus* a phantom seed). The guard strictly reduces error there. The topology itself (external flow entering underground past an unrelated abutting belt) exists in no engine layout, celldb entry, or balancer template; an independent adversarial review with a live checkout is verifying exactly this claim in parallel.
- *[major] "deletes an adjudicated Error signal"* — the signal never measured the deficit; it measured the phantom sources (tile-level receipts in the PR body). Adjudicated in rounds 1–2 and the #644 comment.
- *[major]/[minor] hard zeros vs AC@5, third statement* — the epsilon receipt stands (`> cap + 0.01`, strict). New receipt for the "CI nondeterminism could produce a different layout" variant: the 22s/23s/25s siblings of the SAME from-ore construction have carried `max_errors: 0` pins through CI for months without a flake — the 2026-05-02 nondeterminism record is specific to AC@5's junction-class errors, and that pin keeps its documented looseness in this PR.

**Refuted:**
- *"no mutation test exists"* — "mutation-checked" is this repo's established term (B5, #643) for a test verified to FAIL against the unfixed code, which both new tests were, in TDD order: `ug_exit_with_fed_pair_is_not_an_external_source` was run red first ("head of the external run must seed the full 12/s, got 6.00" — the failure is quoted in the PR body) and asserts conservation (head == tail through the pair). The docs' claim describes verification that actually ran.

Further rounds that re-litigate the guard's shared-predicate design or the zero pins without a concrete reproducing topology or a CI flake receipt will be noted and closed without code change.

storkme:
**Session-side adversarial review complete** (independent agent, isolated worktree, 167 tool calls — required in addition to the bot for validator-semantics changes). It found **two blockers the bot rounds missed**, both now absorbed in 4de86fec:

1. **The round-1 guard was wrong: presence ≠ feeding.** It skipped a UG exit whenever *any* belt tile sat behind the paired entrance — but a perpendicular, different-item belt can abut the entrance without feeding it. The RFC-064 sci2 rotation artifact has exactly that shape; the guard de-seeded its iron-ore run to 0.0/s (whose pre-fix attribution reconciled *exactly*, Σ=11.000=total — the deleted seed was real), fabricating 15 starvation warnings and turning `rfc064_packed_router` red. The corrected guard requires the behind belt's direction to step onto the entrance tile; new regression test `ug_exit_behind_crossing_belt_still_seeds` pins the crossing shape.
2. **My "full suite green (all binaries)" claim was an aborted-run miscount** — cargo stops after a failing binary and my grep pattern also matched "0 failed" substrings. Re-verified by exit code: **0, all 26 binaries green** with the corrected guard (`rfc064_packed_router` 4/4 included).

**Its major selection-flip finding resolves in the PR's favor, with receipts.** The fix flips the ec30/ec60-red stress winners — and the flipped-to winners are **entity-count-exact matches of the 2026-08-07 post-lift bank** (3369 and 4967 entities): the layouts the 92.1%/90.7% sim receipts were measured on. The sprawls current main ships (4964/5566 entities, meter 96.6%/90.8%) are **3-day-old artifacts of the phantom errors steering selection** (they only exist in the #646→this-fix window) and were never sim-anchored. So this PR does not ship unanchored geometry — it *restores the measured winners*. The sprawl class is still interesting (its meter readings suggest sprawl may buy delivery — the RFC-069 headroom hypothesis by accident) and is recorded as a Phase-0 follow-up there, reproducible from any pre-merge SHA.

Also verified by the agent: all five iron-ore phantoms were UG exits (no second source-detection hole); celldb (0/8 entries contain UG belts) and balancer templates don't depend on UG-exit seeding; `build_ug_pairs`' missing reach cap noted as latent (0 over-reach pairs across the family today); clippy clean.

**Bonus finding from RFC-069's Phase-0 sims (also answers round 3's "PU 14%" objection):** kit-clean, axis-declared tier5-PU sims at **95.6% delivered** — the meter's 85.6% was measured on an axis-mismatched export. The family's sim-anchored ledger: ec22 −0.6% (PASS — the 22-vs-30 boundary is real under measurement), tier5-PU −4.4%, ec30 −7.9%, ec60-red −9.3%.

storkme:
Round 4 absorbed in b041b8c5 (comment-only — the trivial-delta gate should skip). Per the convergence rule invoked in round 3, this closes the tail.

**Recorded (the round's one substantive item):** the inheritance edge's presence-only predicate vs the seed guard's stricter feeds-the-entrance predicate is a real asymmetry — but it is **pre-existing and unchanged by this PR** (pre-fix modeling double-counted strictly MORE in that topology: phantom seed *plus* the same leak). Now documented at the wiring site with the census receipt (zero non-feeding behind tiles in any engine layout; two in the RFC-064 sci2 artifact) and the deferred symmetric fix named as a #644 follow-up. "Re-fabricating the #644 class" it is not: no corpus layout has the topology, and this PR strictly reduces modeled flow in it.

**Closed under the convergence rule (no new evidence):**
- *item-match on the guard* — third statement of the shared-blindness design point, documented twice in code.
- *hard-zero pins* — fourth statement. One correction of fact: "exact warning counts (13/9)" misreads the code — `max_warnings` are **ceilings**, not exact pins. The lane pins' epsilon receipt and the 22s/23s/25s siblings' months of flake-free CI at `max_errors: 0` stand.
- *"corpus-wide reversal on synthetic evidence, no instrument run backs the claim"* — the receipts are in this thread: tile-level entity dumps (4.5/4.5 → 9.0/9.0 conservation violation pre-fix, 7.5/7.5 conserved post-fix), seed-stats reconciling exactly (Σ=45/45, 30/30), the flipped winners matching the sim-anchored bank **entity-count-exact** (3369, 4967), and two fresh 432k-warmup sims (ec22 99.4%, tier5-PU 95.6%).
- *lock-step comment* — the comment names the WIRING as carrying the clause, which it does; the processing block's lack of it and the geometric unreachability are both stated in the same comment.
- *nit: assert a distinct-category count on the manifest test* — declined: that converts a stated preference into a hard pin on a fixture's warning profile, the exact over-pinning this PR spent three rounds retiring.

PR #649, opened 2026-08-15
docs(rfc): RFC-069 — achievable-duty provisioning for the zero-headroom family (#644)

storkme:
# Intent

Design container for the #644 engine fix, opened on #648's re-attribution: the family's real deficit is zero-headroom provisioning (chains planned at exactly 100% of nominal per-lane capacity), not lane over-capacity.

Phase 0 is already run and logged in the decision log: **ec22 delivers 99.4%** (PASS — the 22-vs-30 boundary is real under measurement, not just under the validator) and **tier5-PU axis-declared delivers 95.6%** (the meter's 85.6% receipt was an axis-mismatched export). Family ledger: ec22 −0.6% / tier5-PU −4.4% / ec30 −7.9% / ec60-red −9.3%, all 432k-warmup converged sims.

Root site identified in the design: `floor(in_lane_cap/rate) × 2` row caps admit exact equality, so integral divisions produce zero-headroom rows **by construction** (copper-plate: exactly 24 furnaces = exactly 15.0/s = exactly one full yellow belt); trunks follow rows 1:1 on the external path, so a duty factor in the row caps reshapes the whole chain.

K69-1..4 kill criteria (mechanism, no-regression, strategy-confound honesty, footprint cost with an owner escalation bar). The #519 margin-probe rejection is explicitly discounted as instrument-contaminated — it was measured with the pre-#648 walker whose phantoms penalized the very taps the margin created; its geometric end-to-end insight survives as a design constraint.

# Verification
Docs-only PR — the RFC's own verification plan governs the implementation PRs. Phase-0 sim receipts banked at `~/spaghettio-corpora/i644-phase0/`.

# Deviations
None.


storkme:
Review absorbed in 4c2c2a96 — all five findings were real doc-precision catches (the Motivation predated Phase 0's own results): tier5's Motivation bullet now leads with the kit-clean 95.6% sim and retracts the 85.6% axis-mismatched headline; K69-1 is pinned to the single bar **96.0% delivered** (registry row matches); the calibration span reads 90.7–95.6 sim-anchored; the expected-shape passage separates the cap (20/row = 83%) from the even-split realized load (18/row = 75%) and now carries the first measured Phase-1 artifact (81×168 / 3,797 entities, +1.3% bbox, 0 errors); the tier5 receipt cites the harness delivered-d% column as authoritative per meter-divergence.md's own delta-column rule.

PR #650, opened 2026-08-15
feat(placer): RFC-069 Phase 1 — planning_duty knob; ec30 gate cleared at 99.4% delivered (#644)

storkme:
# Intent

RFC-069 Phase 1: the `planning_duty` knob (default-off, bit-identical at 1.0) plus the experiment trail that found and measured the first working engine lever for the #644 deficit family.

**The headline receipt: ec30 at `--duty 0.6` sims at 99.4% delivered** (29.82/30, warmup 432k, converged, kit-clean, 168/170 machines working) — versus the 92.1% the default engine ships. The gate (≥96.0% delivered, pre-registered in the RFC decision log) cleared with +3.4pp to spare.

## What the knob does

`LayoutOptions.planning_duty` (default 1.0 = bit-identical; full suite green by exit code, all 26 binaries). At duty < 1, HS dual-input consumer rows are capped at ONE input₀ trunk's block: `floor(belt_cap × duty / input₀_rate)` machines per row. At 0.6 on yellow this gives EC rows of 2 — the engine then deterministically reproduces the shape of the layout that measures 99.4% (dims/entities exact: 106×222 / 4,934).

## The experiment trail (all receipts in the RFC decision log)

- **Phase 1a** (duty scaling the general row caps) — **K69-1 FIRED at both pre-registered duties**: 84.4% delivered, 7.7pp WORSE than baseline. Reverted; the harmful mechanism (backpressure from a nominal-provisioned collection fabric behind extra producer rows) is documented in the cap functions.
- **Phase 1b** (HS block=3) — sim invalidated by a harness kit artifact (overlapping drain banks, logged as a follow-up); the meter refutes independently at 85.0%. Dead.
- **Phase 1c** (HS block=2 via duty 0.6) — **gate cleared at 99.4%**.
- The physics that discriminates 2 from 3: per-pickup extraction fraction. An EC machine draws 4.5/s ≈ 30% of a yellow belt per pickup (a furnace draws 4%), so a 3-machine block's tail extracts from a belt already ⅔ depleted — the RFC-054 gappy-belt mechanism. Blocks of 2 kill the within-row gradient.
- Four instruments converge on the mechanism (sim census, walker IRD warnings, meter `attribute`/`cable_probe` occupancy gradients, three-artifact row census): the ec30 deficit was the EC consumer fan-in, not the at-cap ore rows (which the 99.4% artifact shares verbatim with the 92.1% one).

## What this PR does NOT do

- Does not flip any default: the engine still ships the 92.1% ec30 until Phase 2's family rollout adjudicates the corpus (pins, goldens, scoreboards) with sim receipts.
- Does not claim the family is solved: **ec60-red does not respond to this lever** (meter 90.8% unmoved despite its EC stage reshaping; two open questions logged in the decision log) — its binding site is different, per K69-3's family-divergence expectation.
- `sim_export` gains `--duty` for the A/B workflow; the wasm boundary pins 1.0 until the default flips.

# Verification

- Full core suite by exit code: 0, all 26 binaries, at default duty (bit-identical).
- Unit pin `planning_duty_caps_hs_rows_at_one_trunk_block`: duty 1.0 → output-bound 10-machine HS rows (the 92.1% shape); duty 0.9 → block-capped rows, machine count preserved.
- Gate sim receipts banked at `~/spaghettio-corpora/i644-phase0/` (ec30-duty06-p1c = the 99.4% run; ec30-sprawl = the independent same-shape 99.4% run; the dead arms' reports alongside).
- Lib clippy clean.

# Deviations

- The RFC's original Phase-1 design (duty in the general row caps) was killed by its own gate and REVERTED within this branch — the decision log carries the full arc, including two pre-registered-then-failed attempts. This PR ships the surviving lever only.

### DIFF
--- a/crates/core/examples/sim_export.rs
+++ b/crates/core/examples/sim_export.rs
@@ -173,7 +173,9 @@ fn main() {
             "--duty" => {
                 duty = need(i)
                     .parse()
-                    .unwrap_or_else(|_| usage("--duty must be a float in (0, 1]"))
+                    .ok()
+                    .filter(|d: &f64| *d > 0.0 && *d <= 1.0)
+                    .unwrap_or_else(|| usage("--duty must be a float in (0, 1]"))
             }
             "--quality" => {
                 let q = need(i);
--- a/crates/core/examples/sim_export.rs
+++ b/crates/core/examples/sim_export.rs
@@ -134,6 +134,7 @@ fn main() {
     let mut di = DirectInsertion::Candidate;
     let mut claim: Option<DiClaimOrder> = None;
     let mut belt: Option<String> = None;
+    let mut duty: f64 = 1.0;
     let mut quality = QualityTier::Normal;
     let mut stacking: u8 = 1;
     let mut research_productivity: std::collections::BTreeMap<String, f64> =
@@ -168,6 +169,12 @@ fn main() {
                 })
             }
             "--belt" => belt = Some(need(i)),
+            // RFC-069 Phase 1: planning-duty knob for the K69-1 sim A/B.
+            "--duty" => {
+                duty = need(i)
+                    .parse()
+                    .unwrap_or_else(|_| usage("--duty must be a float in (0, 1]"))
+            }
             "--quality" => {
                 let q = need(i);
                 quality = QualityTier::from_name(&q)
@@ -261,6 +268,7 @@ fn main() {
     let mut opts = LayoutOptions {
         direct_insertion: di,
         max_belt_tier: belt,
+        planning_duty: duty,
         quality,
         stacking,
         research_productivity,
--- a/crates/core/examples/sim_export.rs
+++ b/crates/core/examples/sim_export.rs
@@ -267,6 +267,14 @@ fn main() {
         std::process::exit(1);
     });
 
+    // The fitted duty values are TIER-RELATIVE (bot review round 2):
+    // block = floor(in_lane_cap(tier) × 2 × duty / rate), and with no
+    // --belt the cap resolves to the express default (belt_cap 45), so
+    // --duty 0.6 silently computes the measured-DEAD block 6 instead of
+    // the gate-clearing block 2. Require the tier to be explicit.
+    if duty < 1.0 && belt.is_none() {
+        usage("--duty < 1 requires an explicit --belt (the fitted duty is tier-relative; the RFC-069 gate receipts are on transport-belt)");
+    }
     let mut opts = LayoutOptions {
         direct_insertion: di,
         max_belt_tier: belt,
--- a/crates/core/src/bus/cells/extract.rs
+++ b/crates/core/src/bus/cells/extract.rs
@@ -56,6 +56,12 @@ pub fn generate_cell_layout_with_capacity(
     // (candidate → generate cell → candidate → …; found as a stack
     // overflow the moment the default flipped).
     let opts = layout::LayoutOptions {
+        // planning_duty deliberately NOT inherited (bot review round 2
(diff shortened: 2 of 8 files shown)


storkme:
Review absorbed in 845aefb8: `--duty` now rejects values outside (0,1]; the two stale "gate still open" comments are reworded (the gate CLEARED at 0.6 — default-off is Phase-2 scope, matching the decision log); the budget site documents the fitted-value and stacking-blind limits (conservative direction — S>1 + duty<1 under-caps rather than crediting unmeasured ×S). Refuted with a receipt: `item0_rate` = max-rate solid input **is** input₀ by the HS convention (the internals sort inputs by rate desc and hand the highest to the trunks — see the "Skip input₀ (highest rate)" site). The PR-title framing point is covered in the body's "What this PR does NOT do": nothing ships by default; the 99.4% is the gated lever's receipt.

storkme:
Round 2 absorbed in ee87f7d6 — both 3/3 majors were correct and are fixed: `--duty < 1` now requires an explicit `--belt` (the fitted duty is tier-relative; the express default silently computed the measured-dead block 6), and the unit test now pins **duty 0.6 → blocks of 2** — the gate-clearing config — replacing the dead 0.9→3 arm whose docstring misattributed the 99.4% receipt (good catch; the dead shape is deliberately not enshrined). Also: +1e-9 before the block floor (the fitted products land on integer boundaries by ULP luck), the LayoutOptions doc carries the tier-relative and caller-validation caveats, and the cells/mega sub-solves are documented as **deliberately** duty-1.0 — cell geometry is candidate-independent by design (the sim registry keys on it), so inheritance there would fork the registry; if the default flips, cells get a cell-scoped duty decision, not an inherited one.

storkme:
Round 3 absorbed in ad7b63c8, and per the convergence rule this closes the tail — the round was all-minor and the remaining items re-litigate settled scope. Fixed: the test pins **exactly 10×2** (`== 2`; a degenerate 20×1 could pass `<= 2` — good catch) and the baseline label now says output and input₁ TIE at 10 (both floor(7.5/1.5)×2). Doc precision: values ≤0 clamp conservatively; **>1 and NaN literally behave as 1.0** (`NaN < 1.0` is false, so the cap gate never fires — the review's 'bypass entirely' is the designed no-op, now stated). Declined with rationale: moving the needs-tier check into `build_bus_layout` — duty<1 with tier=None is *unmeasured*, not invalid, and the engine should not hard-fail configurations Phase 2 hasn't adjudicated; validation lives at the experiment surface (the CLI) until shipping semantics exist. The 1e-9 direction is deliberate (stabilizes exact-boundary products; no real belt/rate tuple sits 1e-9 *below* an integer) and the duty-block-vs-block_size epsilon asymmetry is accepted as practically unreachable. The dead-band point is the RFC's own recorded Phase-2 question (duty fraction vs explicit block bound).

storkme:
Round 4 absorbed in 61eaef8d (comment-only — the trivial-delta gate should skip). All three real items were errors in my own absorb-round text, now corrected: block 3's dead-band is **[0.9, 1.0)**, not [0.675, 1.0) — block 2 (the gate-clearing shape) spans [0.6, 0.9); the 1e-9 comment's direction was garbled (a down-slip floors to an unmeasured block **1**, not 3); and '≤0 degrades conservatively' now reads 'maximal fragmentation — the measured direction of safety for delivery, costly in footprint'. The tier-relative-guard items restate round 3's declined scope (blocking non-yellow tiers at the experiment surface would prevent exactly the Phase-2 family measurements the RFC schedules) and stay closed.

PR #651, opened 2026-08-15
feat(placer): RFC-069 Phase 2 — the block cap reaches every candidate path; ec60-red at plan (100.0%)

storkme:
# Intent

RFC-069 Phase 2, first increment: make the input₀ block cap reach every candidate path, not just the HS branch — and clear the ec60-red gate with it.

**Receipts (both pre-registered in the RFC decision log before the runs returned):**
- **ec60-red at `--duty 0.6`: sims AT PLAN — 60.00/60.00 produced (+0.0%), PASS**, converged, 338/340 machines working. From the 90.7% baseline to 100.0%.
- ec30's gate-cleared shape holds under the extended cap: same dims/entities, meter identical to four decimals per stage (96.6% class → its banked 99.4% sim receipt).

## The root cause this fixes

ec60-red's non-response to #650's lever was root-caused via the `DecompositionChosen` trace: at duty 0.6 the capped HS candidate balloons (6,572 entities), its **density score** loses, and NATIVE wins — whose vertical dual-input EC rows never consulted the HS-branch cap. The lever was hostage to a delivery-blind objective. Rather than touch the objective, the cap now applies to dual-input rows (≥2 solid inputs) on **every** candidate path, so selection chooses among duty-shaped variants on cost as before.

Scope discipline: single-input producer rows stay untouched — the measured 1a harm was shrinking those; the measured 1c win was capping dual-input consumers. The split is now structural (input count), not fitted.

## Known costs and parked items

- ec60-red's duty artifact costs **+28% entities / +64% bbox** vs baseline (K69-4 watched, under its bar; the trade ships nothing by default — the knob stays 1.0).
- **tier5 is promising but router-blocked**: at duty 0.6 the meter reads PU at 100% of plan, but the export carries 9 belt-dead-end + 1 unresolved-junction ERRORS — the ghost router can't yet wire the fragmented deep chain. Parked with receipts; its gate runs only on an error-free artifact.

# Verification

- Pre-registered sim gates, 432k warmup, kit-clean: ec60-red PASS at plan (banked at `~/spaghettio-corpora/i644-phase0/ec60red-duty06-b/`).
- Default duty 1.0 remains bit-identical (the cap is behind `planning_duty < 1.0`); `planning_duty_caps_hs_rows_at_one_trunk_block` green.
- Meter cross-checks on ec30/ec60-red/tier5 as quoted; lib builds clean.

# Deviations

None — this is the RFC's recorded Phase-2 path, receipts in its decision log.

### DIFF
--- a/crates/core/src/bus/layout.rs
+++ b/crates/core/src/bus/layout.rs
@@ -115,7 +115,8 @@ pub struct LayoutOptions {
     /// bit-identical to pre-RFC. The ec30 gate CLEARED at 0.6 (99.4%
     /// delivered vs the 92.1% default — RFC-069 decision log); the
     /// default stays 1.0 because Phase 2 must first settle the family
-    /// semantics (ec60-red/tier5 constructions don't consult this cap)
+    /// semantics (the cap now reaches every candidate path's DualInput
+    /// rows — Phase 2; tier5 remains router-blocked)
     /// and adjudicate the corpus-wide re-ranking a flip causes. Engine
     /// policy, not a user axis. Two caller-owned caveats: the fitted
     /// values are TIER-RELATIVE (the block budget prices off the
--- a/crates/core/src/bus/placer.rs
+++ b/crates/core/src/bus/placer.rs
@@ -3172,7 +3172,39 @@ pub fn place_rows(
             }
         } else {
             let ob = belt_entity_for_rate_stacked(output_rate, max_belt_tier, out_stack);
-            max_machines_for_belt_both_lanes(spec, ob, max_belt_tier, out_stack)
+            let cap = max_machines_for_belt_both_lanes(spec, ob, max_belt_tier, out_stack);
+            // RFC-069: the input₀ block cap applies to DUAL-input rows on
+            // every candidate path, not just HS — otherwise the selection
+            // objective (density-scored, delivery-blind) routes around the
+            // duty shape by picking an uncapped candidate (measured on
+            // ec60-red: at duty 0.6 the capped HS variant balloons and
+            // NATIVE wins with uncapped 7×6 EC rows; decision log).
+            // Scope: ≥2 solid inputs only — the measured 1a harm was
+            // shrinking SINGLE-input producer rows, and the measured 1c
+            // win was capping dual-input consumers.
+            let solid_inputs = spec
+                .inputs
+                .iter()
+                .filter(|i| !i.is_fluid && i.rate > 0.0)
+                .count();
+            if planning_duty < 1.0 && solid_inputs >= 2 {
+                let item0_rate = spec
+                    .inputs
+                    .iter()
+                    .filter(|i| !i.is_fluid && i.rate > 0.0)
+                    .map(|i| i.rate)
+                    .fold(0.0_f64, f64::max);
+                if item0_rate > 0.0 {
+                    let belt_budget =
+                        effective_in_lane_cap(max_belt_tier) * 2.0 * planning_duty;
+                    let block = (((belt_budget / item0_rate) + 1e-9).floor() as usize).max(1);
+                    cap.min(block)
+                } else {
+                    cap
+                }
+            } else {
+                cap
+            }
         };
 
         // RFC-062 Phase 2: a final (target) item that is ALSO consumed by
--- a/crates/core/src/bus/placer.rs
+++ b/crates/core/src/bus/placer.rs
@@ -3180,8 +3180,14 @@ pub fn place_rows(
             // ec60-red: at duty 0.6 the capped HS variant balloons and
             // NATIVE wins with uncapped 7×6 EC rows; decision log).
             // Scope: ≥2 solid inputs only — the measured 1a harm was
-            // shrinking SINGLE-input producer rows, and the measured 1c
-            // win was capping dual-input consumers.
+            // shrinking SINGLE-input rows (ore/plate producers), and the
+            // measured 1c/2 wins were capping dual-input rows with a
+            // high-draw input₀ (EC-class). For dual rows whose input₀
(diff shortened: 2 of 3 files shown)


storkme:
Review absorbed in d62dce5e — both majors were right. The native branch now has its own pin, and writing it taught two things worth recording: (1) the yellow-belt version of the routed-around shape **cannot exist** — the native dual path's existing input limit already caps EC-on-yellow rows at 2 (`floor(7.5/4.5)×2`, no input₀ exemption outside HS), so the pin runs on FAST belts where the limit is 6 (the real ec60-red mechanism: duty 1.0 → 4×6 bit-identical, duty 0.6 → exactly 6×4); (2) the placer balances rows under the cap (20 machines over cap-6 gives 4×5, not 6+6+6+2), so exact-fill counts are required to pin the arithmetic. Scope wording corrected in the branch comment: the discriminator is dual-input rows with a **high-draw input₀** — low-draw dual rows compute a large block and the `min()` is a structural no-op, and fluid+solid duals have one SOLID input and never enter the branch, so the 'unmeasured population' the review worried about is untouched by arithmetic, not by luck. The duty-0.9 dead-band point restates the documented field caveat (block 3 = [0.9, 1.0) on EC-yellow) and Phase 3's shipping-semantics call owns it. (Meta: this round's posted comment contains leaked assembly-prompt text — flagging for the review-bot forensics ledger.)

storkme:
Round 2 absorbed in 4c0301f8 — **both majors were right and materially improved the change**: (1) the `>= 2`-solids guard did reach TripleInput and FluidDualInput rows (whose kind definition has TWO solid inputs — my 'fluid+solid duals never enter' claim was structurally wrong), and (2) my 'low-draw ⇒ no-op' arithmetic was wrong: for an input-bound dual row the block is ≈ duty × the existing cap, so it always bites at duty < 1. The cap is now factored into ONE helper (`duty_input0_block`) with one comment set, gated on exact `RowKind::DualInput` **and** input₀ draw ≥ 10% of the full-belt budget — a threshold that sits between the measured-win fractions (EC: 30% yellow / 15% fast) and the measured-harm class (2–7%), recorded in the RFC decision log as Phase 3's shipping-semantics question. On the confound worry: the *measured* fixtures are unconfounded either way — ec30/ec60-red's only DualInput recipe is EC — and the re-export under the tightened gate reproduces identical row structures and validator state (byte delta = a 4-tile router micro-swap, 3 belts ↔ 1 UG pair, the repo's recorded layout-nondeterminism class; throughput-neutral, receipts attach to the row-structure class). The stacking-blind note now lives in the shared helper (one KNOWN-LIMITS block covers both paths), the DI-fragmentation interaction is noted there as pre-existing-and-accepted, the stale field-doc line is fixed, and the two extra input iterations are gone with the refactor.

storkme:
Round 3 absorbed in a7d9494e — all four real items fixed: the stale 'only at the HS site' comment names both DualInput sites; both_lanes' doc header no longer contradicts its own Input-limit paragraph (the ×2/B7 account is operative and the helper's budget builds on it); the helper doc states that DualInput rows are capped regardless of producer/consumer role, names the out-of-main-path escapes (multi-solid-output single_lane rows; cell/DI sub-solves, deliberate), and records the 10% threshold's tightening of Phase-1b HS behavior as intentional; and — the round's best item — `duty_input0_block_threshold` now pins the fitted boundary directly (1.4/s no-fire vs 1.6/s → block 5 on yellow, the kind gate, duty-1.0 and NaN no-ops), so a threshold edit can no longer silently flip row classes. The stacking-asymmetry and DI-fragmentation notes live in the helper's shared KNOWN-LIMITS block; further rounds re-litigating the fitted threshold's existence (vs its Phase-3 ownership, which the RFC records) will be noted and closed. Suite green by exit code on the refactor (26/26 binaries).

storkme:
Round 4 absorbed in ca2234b3 — the [2/3] was right about the word: the 10% gate **NARROWS** the fired population vs Phase-1b (low-fraction HS rows return to pre-RFC sizing, which is unmeasured under either rule — restoring the status quo outside measured scope is the principle, not a reopened deficit: Phase-1b's cap on that population was never receipted). The EC-on-express exact-10% tie (4.5/45) is now pinned as-firing in the boundary test, with Phase 3 owning whether the tie belongs in scope; 'input₀' is documented as the max-rate solid input (the bound lands on the binding input regardless of listing order). The sum-of-inputs observation is inherited both_lanes arithmetic, inert for the measured class — closed scope. This closes the review tail per the convergence rule: four rounds, each absorbed, with the remaining items restating the fitted threshold's Phase-3 ownership.

storkme:
Round 5 absorbed in bf6f5558 (docs-only — trivial-delta skip expected). The [1/3] was real: the Phase-2 decision-log entry still carried the pre-tightening '≥2 solid inputs' wording — corrected with provenance (code + helper doc authoritative). The [3/3] major is **refuted on its mechanism claim**: the 1a harm was measured on SINGLE-input rows (ore/plate furnaces), and its mechanism was extra rows against a nominal collection fabric — a row-count effect, not a producer-role effect; the physics discriminator this PR ships (per-pickup draw fraction) is role-blind because the tail-starvation mechanism is role-blind. The 'high-draw DualInput producer' population's flagship member is EC itself in deep chains (EC rows are simultaneously producers and consumers there), which is exactly tier5's recorded state: meter-100% at the reshaped rows, gated on the router unblock before its sim. Nothing unmeasured ships — the knob defaults to 1.0. Also recorded this round: the **owner call resolving K69-4's reserved trade-off — correctness over footprint** — Phase 3 ships the duty shape on by default once family semantics settle, with duty-1.0 as the compact escape hatch. The convergence close from round 4 stands; merging on green.

Issue #652, opened 2026-08-15
Router: stale-occupancy zone commits + the duty-reshape mega-zone class (gates RFC-069's default flip, #644)


storkme:
Split from #644's flip campaign (RFC-069 Phase 3): the duty reshape multiplies junction zones, and two pre-existing router defects now gate the default flip. Full probe receipts in RFC-069's decision log (2026-08-15 entries) and on branches `fix/644-tier5-router` + `feat/644-phase3-default-flip`.

**1. Zones are solved against a stale occupancy view (the flip's hard blocker — ac7-HS ships an unpaired-UG Error).** Sequence, traced five layers deep: cluster A commits and claims tiles → cluster B's zone was solved (or cache-replayed) against pre-A occupancy → B's commit-time collision filter silently drops the claimed tiles (tile-by-tile, route-blind) → B's surviving chain ships invalid (an orphan UG entrance at the zone's boundary column, in the measured case). Post-hoc repairs were prototyped and measured to only *move* the invalidity: pair-atomicity turns the orphan-UG Error into a belt-dead-end Error (the upstream hop points at the dropped tile); route-level pruning then dead-ends the external feeder into a hollowed zone. The honest fix is upstream: **re-solve (or re-validate) a zone against current occupancy at commit time** — cache hits make this cheap — or make solve/commit ordering atomic per cluster neighborhood.

Kept from the prototype spike (merged into the flip branch, correct regardless): UG-pair atomicity + never-emitted-partner pruning in `prune_dangling_sat_entities`; an `OppositePairUg` span/axis/bbox guard in eviction.

**2. The tier5 mega-zone class** (branch `fix/644-tier5-router`): the reshaped deep chain forms a 13×7 3-channel crossing wanting 1,547 SAT vars vs the 700 hang-guard ceiling (raising to 1,600 clears 4/5 junctions at negligible cost locally but re-exposes the measured WASM hang risk — a real hang was observed at 756 vars, so var-count is not the hardness discriminator and a blanket raise is not shippable); the fifth cluster is a structural 68-tile interleaved-flow region that no budget clears (item-conflict skips; the retry-with-gaps already fires and doesn't shrink it). Candidate directions: zone splitting at clustering time, channel reduction via routing around, or placer-side de-stacking of the crossing flows.

[...]

PR #653, opened 2026-08-15
fix(router): #652 PR-1 — conflict-aware zone commit + UG-pair atomicity + eviction span guard

storkme:
# Intent

First increment of #652 (the router unit gating RFC-069's default flip): three validity hardenings extracted from the flip campaign's probe spike. Each is correct independent of the flip; the corpus is **bit-neutral** at today's defaults (full suite by exit code: 0, all 26 binaries).

1. **Conflict-aware zone commit** — the commit-time collision filter dropped solution entities tile-by-tile with no route awareness; when the existing entity at a claimed tile DIFFERS (name/direction/io), the solution was produced for different surroundings (cache signature alias / stale claim view) and dropping tiles mutilates routes — the mechanism that shipped an unpaired-UG validator Error on ac7-HS under the duty reshape (root-cause receipts on #652 and in RFC-069's decision log). Conflicting overlaps now reject the whole cluster into the existing cap/retry machinery; identical overlaps stay a benign dedup.
2. **UG-pair atomicity in `prune_dangling_sat_entities`** — a pair lives or dies together, and an entrance whose partner was never emitted is pruned (invalid geometry by construction).
3. **`OppositePairUg` span/axis/bbox guard in eviction** — a 2-tile "jump" beyond the tier's UG reach, off-axis, or exiting the bbox refuses and falls through to the contiguous A* recipes.

Post-hoc pair- and route-level repairs at the commit site were prototyped and measured to only relocate the invalid geometry (orphan-UG → belt-dead-end → hollowed-zone cascade — receipts in the RFC decision log); reject-and-retry is the design that holds.

# Verification

- Full core suite by exit code: 0, all 26 binaries (bit-neutral at default duty — the hardenings only bite on the invalid-geometry classes).
- The motivating defect measured end-to-end on the flip branch: the ac7-HS orphan-UG Error is eliminated by (1) (the cluster caps and re-lays); its remaining red is a different crowding-class defect tracked on #652.
- lib clippy clean (pre-commit hook).

# Deviations

None — #652's remaining scope (stale-zone re-solve semantics, the tier5 mega-zone class, SAT-budget determinism) is explicitly not attempted here.

### DIFF
--- a/crates/core/src/bus/eviction.rs
+++ b/crates/core/src/bus/eviction.rs
@@ -479,15 +479,16 @@ fn route_evicted_spec(
     let path = match recipe {
         EvictionRecipe::OppositePairUg => {
             // 2-tile path triggers `render_path`'s UG-emission branch.
-            // GUARD (#644 flip campaign): render_path emits the pair for
-            // ANY span, but the game's UG reach bounds a legal jump —
-            // an over-reach "pair" ships as an unpaired-UG validator
-            // Error (found on ac7-HS under the duty reshape: the second
-            // hop's exit landed beyond reach, leaving an orphan
-            // entrance at the crossing). Also require both endpoints on
-            // one axis (a diagonal 2-tile "jump" is not a UG) and the
-            // exit inside the bbox. Refusing here falls through to the
-            // A* recipes, which route contiguously.
+            // DEFENSE-IN-DEPTH GUARD: `select_specs` →
+            // `opposite_pair_fits` already enforces span/axis validity
+            // for the specs that reach this recipe (review traced the
+            // active pipeline and found this guard currently
+            // unreachable) — it stays as a second line so a future
+            // change to spec selection cannot silently reopen the
+            // over-reach class: render_path emits the pair for ANY
+            // span, and an over-reach "pair" ships as an unpaired-UG
+            // validator Error. Refusing falls through to the A*
+            // recipes, which route contiguously.
             let span = (exit.0 - entry.0).abs() + (exit.1 - entry.1).abs();
             let axis_aligned = entry.0 == exit.0 || entry.1 == exit.1;
             let max_span = crate::common::ug_max_reach(belt_name) as i32 + 1;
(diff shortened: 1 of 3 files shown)


storkme:
Review absorbed in 88713e06 — full suite EXIT=0, all 26 binaries. Fixed: the conflict predicate includes `carries` (the 3/3 was right — same shape carrying a different item is exactly the built-for-other-surroundings class); the conflict branch mirrors the capped-solve branch's bookkeeping (cluster tiles into `remaining_crossings`, so a conflicting crossing is owned by the retry pass or the unresolved-junction reporter — never silently dropped); the prune rejects a zone as degenerate when it would remove an entity sitting ON a non-interior boundary port (the port-hole edge); the eviction guard is reworded as documented defense-in-depth (your pipeline trace was right — `opposite_pair_fits` already enforces this for selected specs; the guard stays so spec-selection drift can't silently reopen the class); and two targeted tests pin the partner-less-in prune and the boundary-port rejection. Refuted: the nearest-output binding is the game's own pairing semantics — a UG entrance pairs with its nearest same-direction exit, and a farther collinear output belongs to a different hop, so binding-first is correct, not over-pruning.

storkme:
Round 2 absorbed in e1293206. The 1/3 major was right: the reject now runs immediately after solve returns — BEFORE `corridor_handled`/`template_regions`/residue-release — so its state matches the capped-solve branch exactly (the first placement left a stale CrossingZone region and handled-tile marks on `continue`). The symmetric partner-less UG-**out** case is adjudicated too. On the test finding: a mutation probe confirmed your suspicion outright — two successive fixture constructions both went vacuous (one killed by base keep, one by the boundary-port degenerate guard; this BFS deliberately doesn't walk sideloads, so a straight-line partner-less entrance always strands its chain first). The vacuous test is **deleted** rather than kept as false confidence; the branch's live value — multi-channel zones where cross-channel propagation keeps a stranded entrance reachable, the ac7-HS field shape — is documented at the branch with `parity_tier4_ac7_horizontal_stack` named as its regression anchor. Both must-verify items share one receipt: parity 13/13 and full suite EXIT=0 across all 26 binaries — the conflict path (carries predicate included) fires nowhere in today's corpus at defaults, so no Error→gap behavior shift ships with this PR.

PR #654, opened 2026-08-15
fix(router): #652 PR-2 — deterministic cost descent (stacked on #653)

storkme:
# Intent

Second #652 increment, **stacked on #653** (same file). Removes the SAT cost-descent loop's wall-clock deadline; the existing iteration cap (4) becomes the sole budget — bounded and reproducible.

This is the measured root of the long-recorded layout nondeterminism: the deadline made completed-iteration count (and therefore the chosen zone solution) depend on machine speed and build profile. Debug and release built **different layouts for identical inputs** (ac7-HS under the RFC-069 reshape) — the recorded 2026-05-02 AC@5 CI-flake class ("the layout pipeline *should* be deterministic, but isn't on CI").

`effective_budget_ms` survives only as the cache's UNSAT-vs-Timeout classification threshold (a bookkeeping bias between two refusal flavors, noted on #652).

# Verification

- Full suite by exit code: 0, all 26 binaries, **zero pin/golden drift**, ~2min wall (normal range — at most 4 extra bounded solves per zone).
- Direct receipt: debug and release now build **identical** ac7-HS layouts (109×124 / 2877 entities both; they diverged before this change).

# Deviations

None. Retarget note: if #653 merges first, retarget this to main before deleting the base branch (stacked-PR discipline).

### DIFF
--- a/crates/core/src/bus/junction_sat_strategy.rs
+++ b/crates/core/src/bus/junction_sat_strategy.rs
@@ -1466,6 +1466,20 @@ impl JunctionStrategy for SatStrategy {
             // round-2 all-or-nothing version was a strict quality
             // regression on oversized zones). Pure function of zone
             // size and cost — reproducibility untouched.
+            //
+            // HONEST LIMIT (round 4): when the clamp actually bites
+            // (natural cap > max_cap_for_bound) and the clamped solve
+            // is UNSAT, that proves nothing about the unexplored band
+            // (max_cap_for_bound, best_cost-1] — the zone keeps its
+            // base cost even if a cheaper layout exists there. That
+            // band is unencodable within the bound, so no deterministic
+            // scheme can probe it; forfeiting descent depth on
+            // oversized zones is the accepted price of boundedness
+            // (on the UNSAT branch the graduated clamp gains nothing
+            // over an outright skip — its gain is the SAT branch).
+            // The interactive "Improve region" pass (region_reimprove,
+            // wall-clock-budgeted, preview-only) can still find those
+            // layouts; it does not write back to the zone cache.
             const MAX_DESCENT_AUX_VARS: u64 = 200_000;
             let tiles = zone.width as u64 * zone.height as u64;
             let max_cap_for_bound = (MAX_DESCENT_AUX_VARS / (11 * tiles.max(1))) as u32;
--- a/crates/core/src/bus/junction_sat_strategy.rs
+++ b/crates/core/src/bus/junction_sat_strategy.rs
@@ -1396,17 +1396,25 @@ impl JunctionStrategy for SatStrategy {
             initial_cost.expect("entities_opt is Some here, so initial_cost is Some");
 
         // Cost descent: re-solve with a tighter cost cap until either
-        // UNSAT (current best is optimal at this cap), wall-clock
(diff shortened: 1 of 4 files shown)


storkme:
Review absorbed in d307f5ea (rebased onto the updated base — stacked-PR housekeeping). The major's answer is **accept-with-receipts, now recorded in the code**: descent re-solves — including the UNSAT-terminal one — run on zones that already passed the `MAX_ZONE_SAT_VARS ≤ 700` gate, and that ceiling's own calibration found **variable count, not clause count, to be the hang discriminator** (the observed hang: 756 vars / 7,871 clauses, while 20k+-clause zones solved in milliseconds); cost caps add only clauses, so descent solves sit inside the calibrated-safe var range. The residual risk is accepted with the note as the record — a deterministic per-solve conflict budget remains the ideal if varisat ever exposes one. Both doc drifts fixed: the env var and `cost_descent_budget_ms` are documented as classification-only (including the inverted incentive the old doc described backwards — small budgets classify Timeout, which RE-ATTEMPTS; large budgets classify Unsat, which is permanent), and `region_reimprove`'s header now states the two `descend` callers' deliberately different stop conditions (reproducibility for the pipeline, wall-clock responsiveness for the interactive pass).

storkme:
Round 2 absorbed in 11828bcf — **the major was right and the first rationale is retracted**: `encode_cost_cap` feeds 11·tiles literals into the Sinz encoder which allocates n·k auxiliary variables, so capped solves run at tens of thousands of vars and the 756-var base-encoding calibration doesn't apply to them in either direction (verified at sat.rs before replying — the review caught a genuine factual error in my accept-rationale). The corrected account, now in the code: the old wall-clock deadline was checked **between** iterations only and never interrupted a running solve (no watchdog natively or in WASM), so old and new both bound *attempts*, not solve duration; the iteration cap bounds attempts deterministically at the order the 25ms deadline allowed on release hardware; corpus-wide, every capped solve — 30k+-aux instances included — completes in milliseconds (Sinz chains propagate cheaply). Added on top: a **deterministic encoding-size breaker** (skip descent when 11·tiles·cap exceeds 200k estimated auxiliaries — a pure function of zone size and cap, so reproducibility is untouched). The region_reimprove stale-shared header is fixed as caught. The UNSAT-vs-Timeout machine-speed classification nit stands recorded on #652. Debug and release still build identical layouts on the divergence case with the breaker in place.

storkme:
Round 3 absorbed in be9a1a9c. The graduated-breaker suggestion was better design than my round-2 all-or-nothing and is implemented: descent caps may legally jump, so an oversized natural cap is **clamped down** to fit the 200k-aux bound — a deeper deterministic cut on oversized zones instead of abandoned descent, still a pure function of zone size and cost. The cache lookup's Timeout comparison now uses `effective_budget_ms` (hoisted above the lookup so record and compare share the same value — your re-solve-forever scenario was right); the stale 'keeps the loop short' comment is gone and the contradictory 'same class' sentence is rewritten to match the corrected hang note. The classify-from-solver-return item is real but pre-existing and solver-API-sized (`solve_crossing_zone_per_channel` collapses timeout and UNSAT into one `None`) — recorded on #652 as part of its determinism scope rather than expanded here. Receipts: lib green; debug==release holds on the divergence case (111×162 / 3419 both).

PR #655, opened 2026-08-15
fix(router): #652 design half — fail-severed unresolved crossings

storkme:
## Intent

#652 design half, part 1: make the crossing-failure path fail SAFE. The 2026-08-15 diagnosis (recorded on #652) traced **all** of ac7-HS-duty-0.6's 120 errors and the tier5 duty fixture's ~100 to four unresolved crossings shipping **flat sideload-merges** — one item's flow dumped onto one lane of another item's legal trunk, fanning out as ~90 downstream lane-throughput errors per site.

*(History: the PR originally also carried a local conflict-retry re-solve. Review round 1 + the session-side adversarial review independently proved it a deterministic no-op — `refresh_forbidden`'s surface-belt exemption filters the augmented obstacles back out (`tile_is_forbidden_kind = !is_surface_belt`), the port-tile exemption runs before the obstacle test, and the zone-cache signature excludes obstacle sets so the re-solve replays the cached solution. An A/B receipt showed the entire error delta comes from the fail-sever alone. The retry was removed; the retry-with-teeth design is recorded on #652.)*

## Changes

1. **Fail-severed fallback** (`ghost_router.rs`, final pass of `route_bus_ghost`, after Step 7 + tier renames; new `CrossingSevered` trace event): `ghost:`-segment surface belts that point into a belt carrying a DIFFERENT item are dropped when the feed faces an unresolved-cluster tile or either side belongs to an unresolved spec (route ∩ `remaining_crossings`). The flow dead-ends one tile short and fails visibly (belt-dead-end errors, starved consumers, the unresolved-junction region) instead of poisoning a legal trunk. Never severed: committed trunk/balancer/tapoff/crossing/merger geometry (mixing there still reaches `belt-item-isolation`), UG halves (severing an output orphans its input, and `build_ug_pairs` would re-pair to the next output down the lane — inventing a tunnel; measured: every sever decision on both duty shapes was a plain transport-belt, so the arm bought nothing).
2. **Context-conflict trace precision** (`ghost_router.rs`): the #653 conflict-cap path now reports the conflicting tile coordinates.
3. **e2e pin** `tier4_ac7_duty06_unresolved_crossings_fail_safe`: zero lane-throughput errors on the flip shape, plus two anti-vacuity guards — `unresolved > 0 ⇒ severed > 0`, and a liveness assertion (`severed > 0 ∨ unresolved > 0`) so the pin fails loudly if the shape ever stops exercising the fail-safe path (ac7-HS at duty 1.0 has zero errors of any kind, so the lane assertion alone would be vacuous).

## Verification

Fixtures (both at `planning_duty: 0.6`):
- **ac7-HS** = advanced-circuit @ 7/s, AM2, inputs {iron-plate, copper-plate, coal, water, crude-oil}, yellow belt, HorizontalStack.
- **tier5-pu2** = processing-unit @ 2/s via netflow multi, AM3, from-ore inputs {iron-ore, copper-ore, coal, water, crude-oil}, research productivity {plastic-bar: 0.1, processing-unit: 0.1}, red belt, default row layout.

Results (errors / warnings):
- ac7-HS: **120 → 17 errors** (lane-throughput 111 → **0**; isolation 6 unchanged = the pre-existing balancer-over-trunk overlap recorded on #652; +8 belt-dead-end; 3 unresolved-junction). **Warnings 44 → 96** (36→77 input-rate-delivery, 8→14 belt-flow-reachability, +4 belt-flow-path, +1 orphan-belt-segment) — that is the design: starvation is now *visible* where mixing used to hide it.
- tier5-pu2: **100+ → 20** (the accumulating 24→77/s lane cascade eliminated; 15 belt-dead-end + 5 unresolved-junction point at the mega-zone).
- On other unresolved-crossing shapes total **errors can go UP** (session-side review measured a PU AM2 reconstruction at 42 → 48: 10 real mixing sites removed for +6 honest dead-ends, lane-throughput 0 both sides). That is the intended trade — mixing converted to visible failure.
- Corpus at defaults: **byte-identical to main** — independently verified by the session-side review via the suite's SHA-256 layout-hash goldens, not just error counts. Full suite green by exit code (`CARGO_EXIT=0`, all 26 binaries), clippy (CI invocation) clean, WASM check clean.
- Sever accounting: 9 `CrossingSevered` events = 9 net entity removals on ac7 (the final-pass placement means the trace can no longer over-report severs of already-doomed duplicates).
- Determinism: session-side review ran three in-process + separate-process builds — identical FNV digest over the full entity vector and the ordered sever list.

Run per the layout-engine protocol (isolated worktree, opus agent). Verdict: **SHIP-WITH-FIXES** — all fixes absorbed across rounds 1–2: retry removed (its A/B receipt: disabling the retry alone reproduces the branch byte-identically), sever ghost-scoped + surface-belt-only + relocated after Step 6/7, pin liveness guards, scope-gap comments (splitter second tile, row belts), stale-comment truth. Its under-sever enumeration found exactly the 6 recorded residual adjacencies and nothing else (PU census 10 → 0).

## Recorded residuals (on #652, not in scope here)

- **Retry-with-teeth**: forbidden-override through `GrowingRegion` + cache-signature extension; note the ac7 conflict tiles may be zone *ports*, where obstacle-override is structurally insufficient.
- **Balancer-over-trunk overlap**: the 6 remaining isolation errors (EC balancer stamped over the plastic-bar trunk; fragments survive under UG-span holes).
- Same-item lane-override at failed crossings: declined with the pin-to-build named.

### DIFF
--- a/crates/core/src/bus/ghost_router.rs
+++ b/crates/core/src/bus/ghost_router.rs
@@ -4287,6 +4287,12 @@ pub fn route_bus_ghost(
             seg.and_then(|s| s.strip_prefix("ghost:"))
                 .is_some_and(|k| unresolved_keys.contains(k))
         };
+        // Known scope gaps, deliberate (session-side review): only a
+        // splitter's PRIMARY tile is keyed here (a feed into the second
+        // tile is invisible to this map), and row-template belts live
+        // in `row_entities`, not `entities`, so a ghost feed into a row
+        // belt cannot sever — the latter is fine per the owner-confirmed
+        // "the engine never side-feeds row input belts".
         let belt_carries_at: FxHashMap<(i32, i32), (Option<String>, bool)> = entities
             .iter()
             .filter(|e| {
@@ -4309,10 +4315,14 @@ pub fn route_bus_ghost(
             if !is_ghost {
                 return true;
             }
-            let is_feeder = e.name.ends_with("transport-belt")
-                || (e.name.ends_with("underground-belt")
-                    && e.io_type.as_deref() == Some("output"));
-            if !is_feeder {
+            // Surface belts ONLY (session-side review): severing a
+            // UG-output would orphan its input, and build_ug_pairs
+            // re-pairs greedily to the NEXT output down the lane —
+            // inventing a tunnel nobody designed, worse than the
+            // error. Measured across both duty-0.6 shapes: every
+            // sever decision was a plain transport-belt, so the
+            // UG-out arm bought nothing and carried the hazard.
+            if !e.name.ends_with("transport-belt") {
                 return true;
             }
             let Some(item) = e.carries.as_deref() else {
--- a/crates/core/src/bus/ghost_router.rs
+++ b/crates/core/src/bus/ghost_router.rs
@@ -3012,24 +3012,114 @@ pub fn route_bus_ghost(
             &pending_crossings,
         );
 
-        let Some(sol) = junction_solver::solve_crossing(
-            cluster.as_slice(),
-            &keys_at_tile,
-            &routed_paths,
-            &hard_for_junction,
-            &junction_hard,
-            &unreleasable_obstacles,
-            &spec_belt_tiers,
-            &spec_items,
-            &spec_exit_dirs,
-            &spec_kinds,
-            &entities,
-            strategies,
-            &pending_crossings,
-        ) else {
-            // Capped: `solve_crossing` returned `None`, matching the
-            // `JunctionGrowthCapped` event it emitted at this cluster's seed.
-            // Record it for the retry loop (control flow, not the trace stream).
+        // CONFLICT-RETRY solve (#652 design half): `junction_hard` is
+        // snapshotted BEFORE this cluster loop, so a later cluster's
+        // zone can be built blind to entities earlier clusters
+        // committed — SAT then returns a solution that collides with
+        // committed geometry. The conflict check below (#653) detects
+        // that; the fix here re-solves ONCE with the conflicting tiles
+        // added to the strict obstacle set (an obstacle tile is exactly
+        // what UG spans tunnel under, so "taken" is expressible and the
+        // re-solve can still cross). A second conflicting attempt caps
+        // the cluster for the layout-level retry — which is what the
+        // pre-fix path did on the FIRST conflict; since the row-widened
+        // second layout pass hits the same stale-snapshot conflict
+        // deterministically, that shipped the crossing unresolved as a
+        // flat sideload-merge (the 2026-08-15 #652 diagnosis measured
+        // ~90 of ac7-HS-duty-0.6's 111 lane-throughput errors as the
+        // downstream shadow of ONE such merge at (9,97)).
+        // Identical-entity overlaps remain a benign dedup (skipped at
+        // stamping, as before); only DIFFERING overlaps
+        // (name/direction/io/carries — #653 review, 3/3) conflict.
+        let conflict_tiles = |sol_entities: &[crate::models::PlacedEntity],
+                              occ: &crate::bus::ghost_occupancy::Occupancy|
+         -> Vec<(i32, i32)> {
+            sol_entities
+                .iter()
+                .filter(|ent| {
+                    let tile = (ent.x, ent.y);
+                    let claimed = matches!(
+                        occ.claim_at(tile),
+                        Some(crate::bus::ghost_occupancy::Claim::Template { .. })
+                            | Some(crate::bus::ghost_occupancy::Claim::RowEntity { .. })
+                    );
+                    claimed
+                        && occ.entity_at(tile).is_some_and(|existing| {
+                            existing.name != ent.name
+                                || existing.direction != ent.direction
+                                || existing.io_type != ent.io_type
+                                || existing.carries != ent.carries
+                        })
+                })
+                .map(|ent| (ent.x, ent.y))
+                .collect()
+        };
+        let mut sol_opt: Option<junction_solver::JunctionSolution> = None;
+        let mut conflict_obstacles: FxHashSet<(i32, i32)> = FxHashSet::default();
+        for attempt in 0..2u8 {
+            let strict_aug;
+            let strict: &FxHashSet<(i32, i32)> = if conflict_obstacles.is_empty() {
+                &junction_hard
+            } else {
+                let mut aug = junction_hard.clone();
+                aug.extend(conflict_obstacles.iter().copied());
+                strict_aug = aug;
+                &strict_aug
+            };
+            let Some(s) = junction_solver::solve_crossing(
+                cluster.as_slice(),
+                &keys_at_tile,
+                &routed_paths,
+                &hard_for_junction,
+                strict,
+                &unreleasable_obstacles,
+                &spec_belt_tiers,
+                &spec_items,
+                &spec_exit_dirs,
+                &spec_kinds,
+                &entities,
+                strategies,
+                &pending_crossings,
+            ) else {
+                // Capped: `solve_crossing` returned `None`, matching the
+                // `JunctionGrowthCapped` event it emitted at this
+                // cluster's seed. Fall through to the cap below.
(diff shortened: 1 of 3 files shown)


storkme:
**Round-1 absorption** (e2e0e86f). Both majors bought; two minors bought; one declined with the record.

- **Major 1 (retry is a no-op) — confirmed and absorbed by REMOVAL.** Verified against the code: `refresh_forbidden` (junction_solver.rs ~545-574) exempts surface-belt tiles from both obstacle sets by design (the benign identical re-derivation + stamping dedup depends on it), and the zone-cache signature excludes obstacle sets, so the re-solve replays the identical cached solution — matching what the probe observed (same two tiles conflicting on both attempts). The retry loop is gone; the conflict check stays with tile coordinates in the trace; a conflicted cluster caps for the layout-level retry as before. The retry-with-teeth design (forbidden-override through `GrowingRegion` + cache-signature extension, plus the port-tile caveat) is recorded on #652. This also dissolves **minor 1**: with no local retry, every emitted `CrossingZoneSkipped` is a genuine cap again — the scoreboard semantics are unchanged from main.
- **Major 2 (sever not scoped as documented) — bought.** The pass now severs only `ghost:`-segment entities; committed trunk/balancer/tapoff/crossing/merger geometry is never dropped regardless of mixing (that mixing still reaches `belt-item-isolation`). Measured cost of the restriction: zero — all nine ac7 severs and all tier5 severs were already ghost segments; error counts identical (ac7 17, tier5 20, lane-throughput 0 both).
- **Minor 2 (stale claims vs later phases) — bought by relocation.** The sever pass moved to the end of `route_bus_ghost`, after Step 7 and the tier renames: no later phase can stamp over a severed tile (no resurrection, no `AlreadyClaimed` panic path), and `existing_tiles` snapshots taken from `entities` can no longer disagree with it. The comment now states what is actually downstream of the pass.
- **Minor 3 (pin vacuity) — bought.** The test now also asserts `unresolved == 0 || severed > 0`, so "zero lane errors because the sever never engaged while crossings remain unresolved" fails loudly. The same-item lane-override half is **declined with a record on #652**: a same-item join is not unambiguously invalid geometry, no measured fixture exhibits the class, and auto-severing legitimate joins is the harmful direction — the record names the pin to build if an instance ever appears.
- **Nit (dead code) — fixed** (the differing item is bound by the match; the unreachable `unwrap_or_default` is gone). **Nit (suite may turn red)** — refuted with receipt: full suite green post-change, `CARGO_EXIT=0`, all 26 binaries.

Receipts on e2e0e86f: ac7-HS duty 0.6 → 17 errors (lane 0 / iso 6 / dead-end 8 / unresolved 3); tier5 duty 0.6 → 20 (dead-end 15 / unresolved 5); pin green standalone (20 s); full suite green by exit code.

storkme:
**Round-3 adjudication.** One major bought (real coverage gap), one refuted with a receipt the round asked for, two comment items bought, the rest declined as deliberate scope with the enumeration receipt.

**Bought — major 1 (last-tap alias).** Verified: the unified-spec pass removes both `trunk:{item}:{x}` and the last tap's `tap:{item}:{x}:{y}` from `routed_paths` (folding into `flow:{item}:{x}`), while the tap's belts keep their original `ghost:tap:…` segment — so last-tap feeders were invisible to the two spec-scoped clauses and severed only on exact crossing-tile adjacency. `unresolved_seg` now aliases `tap:{item}:{x}:{y}` → `flow:{item}:{x}`. Fixture impact: zero (ac7 unchanged at 3410 entities / 17 errors / 9 severs — its severs all fire via clause 1), which is exactly why the review's structural reading mattered: the gap was invisible to the measured shapes.

**Refuted — major 2 (dead-end baselines could break).** The round says "needs a full-suite run; I couldn't execute one" — it has run, three times across the absorption rounds, on exactly these heads: `CARGO_EXIT=0`, all 26 binaries including `processing_unit_2s_am2_fast_belts_validation_baseline`, plus the SHA-256 layout-hash goldens (which the session-side review verified byte-identical to main — stronger than error-count equality). The hypothesized regression does not occur; the row-template explanation the round itself offers is the correct one.

**Bought — minor 6's documentation kernel + the whitespace nit.** The `carries == None` early-keep now carries its rationale in place (undecidable difference; severing on "unknown" would tidy legitimate joins; an unstamped mixing target still reaches the adjacency checks). Stray blank line removed.

**Declined — minors 1/2/3/4 (scope-radius family).** All four describe geometry the pass deliberately does not touch, and the operative receipt is the session-side review's post-fix enumeration: across both duty-0.6 shapes, **every** differing-item belt→belt adjacency that survives is one of the 6 recorded balancer-over-trunk residuals (ac7 census 15 → 6, PU census 10 → 0). Row-template targets rest on the owner-confirmed no-side-feeds invariant; ghost splitter feeders don't exist (ghost routes stamp belts and UGs only, and UG feeders are excluded for the orphan-pair hazard with a 225/225 receipt); the reverse committed→ghost merge is empty in the same enumeration. The pass's contract is stated at the block: it kills the measured cascade class without silently tidying mixing elsewhere — over-widening it is how novel router bugs would get hidden from the validator.

**Declined — minors 7/8 (pin liveness trade-offs).** Both describe the documented intent: the pin going red on a future improvement (or on a fixture drifting to unresolved-without-mixing) is the §7 mechanism working — a human re-points it, and the comment says exactly that in caps. A pin that silently tolerates the path going unexercised is the ten-incident failure class this repo documents.

**Declined — minor 9 (pre-sever bounds/regions).** The unresolved-region markers on severed tiles are correct output — the marker's purpose is "a crossing failed here", which remains true after the feeder is dropped; `max_y`/`merge_max_x` being computed pre-sever affects only slack in the returned bounds. Cosmetic, noted.

**Nit (wasm/tsify).** Receipted: the session-side review ran `cargo check -p spaghettio_wasm --target wasm32-unknown-unknown` → exit 0 on a head that already contained `CrossingSevered`.

Receipts on this push: ac7 3410 entities / 17 errors / lane 0 / 9 severs = 9 removals; pin green with both guards (21 s); full suite re-running on the final head, result posted before merge.

storkme:
**Round-4 adjudication (convergence — all minors, merging on this).**

- **Bought: the stale title** (minor 6) — retitled to match shipped behavior; the body already carries the retry post-mortem as history.
- **Refuted — minor 5 (liveness reads pass-1 severs):** `build_bus_layout` discards pass-1 trace events on retry (`layout.rs` — "Discard pass-1 events from the collector so `result.trace` reflects only the retried (final) pass", with the sink detached so streaming consumers never see them either). The drained `CrossingSevered` count is final-pass only; the scenario cannot occur.
- **Refuted — minor 2 (no `trunk:` alias):** trunk belts are committed geometry with un-prefixed `trunk:{item}` segments — they are excluded from severing *by round 1's own major* (never drop committed geometry), and no `ghost:trunk:…` segment exists for an alias to match. Asking the sever to treat trunks as droppable is the reverse of the demand this PR already absorbed.
- **Declined — minor 1 (alias over-broadens to non-last taps):** true as stated, and the direction of the error is conservative: it can only sever *additional differing-item merges* from taps in a column whose unified flow is unresolved — mixing that is invalid geometry regardless, leaving dead-end evidence rather than hiding anything. Narrowing it would require threading per-lane `last_tap_y` into the pass for zero measured benefit.
- **Declined — minor 3 (same-tile last-wins):** the pass now runs after the Step-6 occupancy sync (round-2 relocation), which is precisely the pass that resolves overstamped duplicates; the residue the session-side review measured (35 distinct-carry duplicates, all resolved in favor of the survivor) existed only at the pre-relocation position. Latent by the round's own measurement.
- **Noted — minor 4 (chained severs can double dead-ends):** benign and measured-not-derived; the 8/15 dead-end figures are counts from the shipped layouts, and a chained feed dead-ending one tile earlier is the same visible failure.

Merging: required check green on 739c4127, mergeStateStatus CLEAN, session-side adversarial verdict SHIP-WITH-FIXES fully absorbed (rounds 1–3), suite green on every absorption head.

PR #656, opened 2026-08-15
fix(router): #652 mega-zone unit — stop region growth when every SAT attempt is over the var ceiling

storkme:
## Intent

#652 mega-zone unit: stop region growth when it is provably futile. Full-zone SAT variable count is monotonic in zone size (growth only adds participating specs, so tiles and channels both grow), so a zone whose every SAT attempt was refused for exceeding `MAX_ZONE_SAT_VARS` cannot be rescued by growing the *full* zone — yet the growth loop kept expanding: tier5's mega-zone grew 935 → 1020 → 1547 vars while refused at 935 the whole way, reaching the tile cap at 91 tiles and leaving a 68-tile orphan field for #655's fail-sever pass to clean up.

## Changes

1. `JunctionStrategyContext`/`SolveCtx` gain per-iteration counters (`sat_invocations` / `sat_ceiling_refusals`); every `SatStrategy` rung increments them (including eviction's filtered inner attempts), in a pre-pass **above** the zone-cache lookup so cold and warm runs grow identically (the #654 divergence class). The user-visible gate (trace event + synthetic-Timeout record) stays in its original position — trace and cache behavior are byte-identical to before this unit.
2. The growth loop treats an iteration with **no candidate, no walker veto tiles, `iter ≥ 1`, and every SAT invocation over-ceiling** as SAT-hopeless. It then runs **one bonus attempt round at the current size with `growth_iter` forced to `MAX_GROWTH_ITERS − 1`** — eviction is final-iteration-gated and is the only strategy that can *shrink* the SAT instance (its filtered zones can be under the ceiling at any region size), so it gets its attempt at the smallest refused size, which is strictly more favorable than after futile growth. Only if that round also yields no candidate does the loop emit the new `JunctionGrowthCapped { reason: "sat_var_ceiling" }` and stop. The `iter ≥ 1` floor means local template strategies always see at least one grown region before the verdict.

*(Review history: the first version stopped on any full-zone refusal; round 1 identified that eviction's final-iteration gate made the "under-ceiling attempt justifies growth" refinement inert and that a plain stop denied eviction entirely — the bonus-round design is the fix. An earlier claim that v1 "measurably regressed a corpus fixture" is **retracted**: that red came from a bare-cache suite run of the kind since proven to trip the same pin on unmodified main — see the verification finding below.)*

## Verification

- **Blast radius (session-side review, corrected from the original body):** the rule fires **15 times across 10 distinct cluster seeds** on the tier5 probe — most firings live in the layout-level retry's *truncated pass-1 trace*, which is why the probe's final trace shows only two `JunctionGrowthCapped` events. Outcome verified **byte-identical to main in both error and warning populations** on the probe; the containment claim holds, but the original "one cluster" framing under-counted per `docs/validator-reporting.md`.
- tier5 duty-0.6 probe: the (29,179) cluster caps at **iter 1 / 55 tiles** with the honest `sat_var_ceiling` reason — its bonus eviction round also fails, correctly (was: grown to the 91-tile cap). The (25,173) mega-zone still reaches `tile_cap` — its iterations carry item-conflict veto tiles and the rule defers to them; that zone's real fix (zone-split / channel-reduction) stays recorded on #652. Error counts unchanged (15 belt-dead-end + 5 unresolved-junction): this unit is containment, not resolution.
- Full suite **the CI way** (`SPAGHETTIO_ZONE_CACHE_PATH=<copy of sat-zones-ci.bin>`): `CARGO_EXIT=0`, all 26 binaries — on both the original and the round-1 head.
- **Verification finding, recorded in `docs/test-suite-followups.md` (new section):** the e2e warning pins are *cache-relative* — on unmodified main (308633b6), `tier4_advanced_circuit_from_ore_am2` reproduces its pinned `input-rate-delivery 7` **only** against the committed CI cache snapshot; both this host's drifted live cache and the embedded-cache baseline yield `belt-detour 1 + input-rate-delivery 8`. A bare local `cargo test` currently trips that pin on ANY branch including main (receipts: main-source warm, cold, single-test, and full-suite runs all reproduce the trip; branch and main byte-identical on the fixture). The followups section documents the local protocol and flags `ci.yml`'s stale wall-clock rationale comment for the next workflow-file touch.
- Known gap (recorded on #652): no automated test yet asserts the `sat_var_ceiling` stop fires — a synthetic `solve_crossing` fixture must exceed 700 vars while clearing the item-conflict fast-fail; sketch in the #652 record.

Verdict **SHIP-WITH-FIXES** (all fixes documentation-only, absorbed): claims A/B verified exactly (including branch-vs-main under an identical CI-cache copy: the (29,179) flip is real, everything else byte-identical); the v1 retraction verified *harder* — literal v1 semantics also passes CI-way, and instrumentation showed arm (c) was never load-bearing on the corpus fixture (the veto guard was); counter hygiene, template blind spot (structurally closed: the perp template only fires on 1-tile regions), determinism, WASM, and clippy all clean; monotonicity empirically confirmed (0 decreasing var-count steps across 463 series). Stale source records rewritten, blast radius corrected above, the eviction 200 ms wall-clock residual added to the followups doc.

## Deviations

The stop rule was redesigned once during review (see history note above); no other deviations. The tier5 mega-zone's actual resolution is out of scope per the #652 plan.

### DIFF
--- a/crates/core/src/bus/eviction.rs
+++ b/crates/core/src/bus/eviction.rs
@@ -231,6 +231,8 @@ impl JunctionStrategy for EvictionStrategy {
                 strict_obstacles: ctx.strict_obstacles,
                 placed_entities: ctx.placed_entities,
                 unreleasable_obstacles: ctx.unreleasable_obstacles,
+                sat_invocations: ctx.sat_invocations,
+                sat_ceiling_refusals: ctx.sat_ceiling_refusals,
             };
 
             let sat_started = web_time::Instant::now();
--- a/crates/core/src/bus/junction_sat_strategy.rs
+++ b/crates/core/src/bus/junction_sat_strategy.rs
@@ -1139,6 +1139,21 @@ impl JunctionStrategy for SatStrategy {
         // the record side writes.
         let effective_budget_ms = env_descent_budget_ms()
             .unwrap_or(self.constraints.cost_descent_budget_ms);
+        // Ceiling-aware growth-stop bookkeeping (#652): counted on EVERY
+        // encounter — including warm-cache paths that early-return below
+        // — so the growth loop behaves identically cold and warm (a
+        // flag set only on the cold path made growth depend on cache
+        // state, the #654 divergence class). The user-visible gate
+        // (trace event + synthetic-Timeout record + refusal) stays in
+        // its original place downstream of the cache lookup, so trace
+        // and cache behavior are byte-identical to before this unit.
+        ctx.sat_invocations.set(ctx.sat_invocations.get() + 1);
+        let sat_var_count = crate::sat::crossing_zone_sat_var_count(&zone);
+        if sat_var_count > MAX_ZONE_SAT_VARS {
+            ctx.sat_ceiling_refusals
+                .set(ctx.sat_ceiling_refusals.get() + 1);
+        }
+
         let cache_result = crate::zone_cache::lookup_zone_result(
             &zone,
             &channel_reaches,
@@ -1285,8 +1300,9 @@ impl JunctionStrategy for SatStrategy {
         // the structurally-infeasible zones, but a balanced-but-hard zone could
         // still hang the unbounded base solve. Refuse zones over the ceiling and
         // record a synthetic Timeout so re-encounters skip via the normal cache
-        // path. See MAX_ZONE_SAT_VARS.
-        let sat_var_count = crate::sat::crossing_zone_sat_var_count(&zone);
+        // path. (The growth-stop counters for this refusal are maintained in
+        // the pre-pass above the cache lookup — see there.) See
+        // MAX_ZONE_SAT_VARS.
         if sat_var_count > MAX_ZONE_SAT_VARS {
             trace::emit(TraceEvent::CrossingZoneSkipped {
                 tap_item: zone
--- a/crates/core/src/bus/junction_solver.rs
+++ b/crates/core/src/bus/junction_solver.rs
@@ -1037,8 +1037,17 @@ pub struct JunctionStrategyContext<'a> {
     /// size-hopeless for SAT", not "who emitted the refusal trace").
     /// When an iteration yields no candidates, no walker veto tiles,
     /// and EVERY SAT invocation was over-ceiling, growth is futile:
-    /// var count is monotonic in zone size (tier5 measured 935 -> 1547
-    /// while refused at 935 the whole way). The equality test (rather
+    /// var count is monotonic in zone size — tiles grow by
+    /// construction, and the channel count cannot DROP because an
+    /// absorbed/engulfed spec's flow is re-expressed as INTERIOR
+    /// boundaries (`interior: true` in the zone builder) which
+    /// `channel_info_from_boundaries` still counts; growth converts
+    /// perimeter crossings to interior ones, never removes a channel.
+    /// (Empirical check: 0 decreasing var-count steps across 463
+    /// (seed, variant) series on the tier5 probe. If zone construction
+    /// ever stops emitting interior boundaries for engulfed specs,
+    /// re-derive this lemma.) Tier5 measured 935 -> 1547 while refused
+    /// at 935 the whole way. The equality test (rather
     /// than any-refusal) matters because the SINGLE-SIDE VARIANT zones
     /// are one row/column larger than the primary — a variant can sit
     /// over the ceiling while the primary is under, and an
@@ -1308,7 +1317,8 @@ pub fn solve_crossing(
             // in the trace — a visible iter jump that marks the
             // ceiling-stop bonus round.
             //
-            // Two accepted asymmetries (session-side review): the
+            // Three accepted asymmetries (session-side review + round
+            // 3): the
             // bonus round re-runs the full-zone SAT rungs too (they
             // re-refuse at the ceiling pre-solve — four cheap checks,
             // duplicate skip events under the faked iter label; only
@@ -1319,7 +1329,11 @@ pub fn solve_crossing(
             // veto manufactured under the faked final-iter context
             // would contradict the stop's premise (measured: zero
             // discarded vetoes across all 15 bonus rounds on the tier5
-            // probe).
+            // probe). Third: the bonus round runs on the PRIMARY
+            // region only, not the four variants — eviction's value is
+            // shrinkage, and the primary is the smallest region on the
+            // table; variant regions are strictly larger, so their
+            // filtered zones are strictly less likely to fit.
             if iter + 1 < MAX_GROWTH_ITERS {
                 if let TryOutcome::Solved(sol) =
                     try_solve_on_region(&region, MAX_GROWTH_ITERS - 1, None, &solve_ctx)
--- a/crates/core/src/bus/junction_solver.rs
+++ b/crates/core/src/bus/junction_solver.rs
@@ -1025,6 +1025,22 @@ pub struct JunctionStrategyContext<'a> {
     /// relies on release-in-footprint for trunk/tapoff cleanup. Computed
     /// once per `solve_crossing` call by the caller.
     pub unreleasable_obstacles: &'a FxHashSet<(i32, i32)>,
+    /// SAT invocations this growth iteration (every `SatStrategy`
+    /// rung, INCLUDING eviction's filtered inner attempts). Paired with
+    /// `sat_ceiling_refusals` below to drive the ceiling-aware growth
+    /// stop (#652 mega-zone class). Reset by the growth loop per iter.
+    pub sat_invocations: &'a std::cell::Cell<u32>,
+    /// How many of those invocations REFUSED the zone because its SAT
+    /// variable count exceeds `MAX_ZONE_SAT_VARS`. When an iteration
+    /// yields no candidates, no walker veto tiles, and EVERY SAT
+    /// invocation was over-ceiling, growth is futile: full-zone var
+    /// count is monotonic in zone size (tier5 measured 935 -> 1547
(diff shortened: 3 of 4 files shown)


storkme:
**Round-1 adjudication — the critical and major 2 are bought together, and they exposed a narrative error of mine that I'm retracting explicitly.**

**Retraction first.** The PR body claimed v1 of the stop rule "measurably regressed `tier4_advanced_circuit_from_ore_am2` … whose zone grows until eviction's filtered solve wins." That claim was **environmentally confounded**: v1's red came from a bare-cache suite run, and we subsequently proved (receipts in `docs/test-suite-followups.md`, added in this PR) that the same bare run trips that pin on **unmodified main**. v1 was never tested the CI way; the eviction mechanism story was an inference, not a measurement. The PR body is rewritten accordingly. What remains true and *is* the operative design input is this round's structural finding: **eviction is final-iteration-gated** (`eviction.rs`: `growth_iter + 1 < MAX_GROWTH_ITERS ⇒ None`), so on iterations 0–3 the invocation/refusal counters see only full-zone rungs sharing one var count — making v2's "under-ceiling attempt justifies growth" clause inert exactly as major 2 states.

**Bought — critical + major 2, one restructure.** The stop no longer returns `None` directly. When the all-over-ceiling condition fires it now (a) requires `iter >= 1` (never stops on the seed-sized region — local template strategies get at least one grown region first), and (b) runs **one bonus attempt round at the current size with `growth_iter` forced to `MAX_GROWTH_ITERS - 1`**, opening the final-iteration gate so eviction — the only strategy that can *shrink* the SAT instance — gets its shot at the smallest refused size (strictly more favorable than after futile growth: the filtered zones grow with the region too). Only if that round also produces no candidate does the loop emit `sat_var_ceiling` and stop. The bonus round routes through `pick_cheapest_candidate` for identical trace side effects, and its iter-jump is visible in the trace by design. Measured: tier5's (29,179) still caps at iter 1 / 55 tiles (its bonus eviction attempt also fails, correctly); (25,173) unchanged; error counts unchanged; CI-way full suite green (receipt below).

**Bought — major 3 in part.** Conceded: no test asserts the stop fires. A synthetic `solve_crossing` fixture needs a dense multi-channel crossing that clears the item-conflict fast-fail while exceeding 700 vars — constructible but fiddly; recorded on #652 with the fixture sketch rather than blocking this PR on it. The tier5 probe measurement is the current receipt; the trace reason string is greppable in snapshots.

**Refuted — minor 1 (91 vs 64).** Both numbers are real: `MAX_REGION_TILES` is 64 and the trace reports `region_tiles: 91` *at the cap* because the check fires after a growth step that overshot (veto-directed growth adds whole rows/columns). The comment cites the measured trace, not the constant.

**Refuted — minor 2 (channel-count non-monotonicity).** Growth only ever ADDS participating specs (absorption and encountered→participating promotion); channels are the participating specs' items, so channel count is monotone non-decreasing under growth and the var count with it. No power-of-two boundary can be crossed downward.

**Nits.** "REFUSED the zone" docstring: reworded (the count answers "is this zone size-hopeless", including warm-cache encounters). "Computed twice": refuted — one live call site (`junction_sat_strategy.rs:1151`); the downstream gate reuses the binding. Encoder memoization: declined with receipt — `CrossingEncoder::new` is arithmetic plus one `Vec` of ≤64 tile-layout entries; sub-microsecond on the scales this path sees.

storkme:
**Session-side adversarial review absorbed** (verdict SHIP-WITH-FIXES; full report summarized in the PR body's new review section). The highlights that changed this PR:

- **Its C-verdict strengthened my retraction with the receipt I couldn't produce**: literal v1 semantics *passes* the CI-way suite, and instrumentation shows zero iterations on the corpus fixture where v1 and v2 diverge — the two mixed iterations (`inv=20/refusals=16`, `inv=8/refusals=4`) are protected by the **veto guard**, not arm (c). Arm (c)'s honest justification is the **variant-zone case** (single-side variants run one row/column larger than the primary, so a variant can be over-ceiling while the primary is under) — and that is what the two source records now say. The stale docstring + comment that still asserted the retracted v1-eviction measurement as fact are rewritten (the records-outlive-state class, caught exactly where the repo's own docs predict it).
- **Blast radius corrected**: 15 firings across 10 cluster seeds on the tier5 probe (pass-1 trace truncation hid 13 of them from my measurement), outcome byte-identical to main in error *and* warning populations. Containment holds; the "one cluster" framing under-counted.
- **A real wall-clock residual surfaced**: eviction's live 200 ms recipe budget survives #654 and gates layout-changing solves — the followups doc's "not wall-clock" sentence was an overclaim and now names it.
- Bonus-round asymmetries (full-rung re-runs; discarded bonus vetoes — zero observed across all 15 firings) documented in place.

Also noted from the report, no action needed: counter hygiene clean at both construction sites, the template blind spot structurally closed (perp template fires only on 1-tile regions, so growth cannot flip it), determinism/WASM/clippy clean, and the environmental cache claim reproduced on unmodified main in both directions.

Receipts on this head: one scoped code change (the `deferred` flag below) plus comment/doc true-ups; tier5 probe byte-identical through every change this round; CI-way full suite green (`CARGO_EXIT=0`, all 26 binaries).


---

**Bot round-2 adjudication (same push):**

- **Bought — major 1 (deferred-exit blind spot).** Correct and constructible: a deferred-exit discards a *valid* solution and `break`s with no veto tiles, and growth is that path's documented remedy — an iteration mixing a deferral (template-first variant) with ceiling refusals (other variants) could have stopped wrongly. `TryOutcome::Continue` now carries a `deferred` flag; the stop requires `!deferred_any`. Tier5 behavior byte-identical (receipt above); this closes the gap structurally rather than empirically.
- **Declined — major 2 (bonus round doesn't close the v1 shape).** The "eviction wins only after growth" scenario's single alleged instance was the environmental confound (see the retraction — literal v1 passes CI-way); the session-side instrumentation found zero mixed iterations unprotected by the veto guard across 463 series, and with this round's `deferred_any` the stop now defers to every signal class that can justify growth (candidates, vetoes, deferrals, under-ceiling attempts, seed-size). What remains is a hypothetical with no observed instance and a measured cost on the other side (futile growth to the cap on hopeless zones); if a fixture ever exhibits it, the #652 test-sketch record is where its pin goes.
- **Declined — minor 1 (duplicate bonus-round skip events).** Documented in place; the events carry the faked `iter = MAX_GROWTH_ITERS - 1` label, so any counter that cares can distinguish them, and the cache write is idempotent as the round itself notes.
- **Declined — minor 2 (var-count on Solved hits).** `CrossingEncoder::new` is arithmetic plus one ≤64-entry Vec; the counter must count warm paths or the growth loop diverges by cache state (the #654 class — stated in the pre-pass comment).
- **Acknowledged — minor 3 (coverage).** Same item as the adversary's MINOR-7 and the existing #652 record; the sketch there now also covers the deferred-exit suppression case this round added.

storkme:
**Round-3 adjudication — and convergence.** This round's findings are new hypotheses against standing structural and empirical receipts; the two real kernels are comment additions (pushed). Merging on the green head.

**Refuted — critical (channel count can drop under growth).** Structurally impossible in the current zone builder: an absorbed/engulfed spec's flow is re-expressed as **interior boundaries** (`interior: true` at the four emission sites in `junction_sat_strategy.rs`), and `channel_info_from_boundaries` counts ALL boundaries, interior included — growth converts perimeter crossings to interior ones, it never removes a channel, so `n_channels` (and with tiles, `total_vars`) is monotone non-decreasing. Empirical cross-check already on record from the session-side review: **0 decreasing var-count steps across 463 (seed, variant) series**. The monotonicity lemma, its structural dependency, and the re-derive instruction are now stated at the counter docstring — if zone construction ever changes that invariant, the comment names exactly what to re-prove.

**Declined — major 1 (bonus round primary-only).** Deliberate, now documented as the third accepted asymmetry: eviction's value is *shrinkage*, the primary is the smallest region on the table, and every variant is strictly larger — their filtered zones are strictly less likely to fit under the ceiling. Running five evictions per stop would multiply the bonus cost for a shape class with zero observed instances.

**Refuted — major 2 (non-SAT strategy denied a larger region).** Round-1's critical, re-raised; structurally closed and receipted then: `PerpendicularTemplateStrategy` is the only non-SAT strategy and it fires **only on 1-tile regions** (session-side review verified the precondition), so growth cannot flip it from fail to success; the `iter ≥ 1` floor additionally guarantees one grown region before any stop. There is no lane-splitting strategy in the junction strategy list.

**Declined — minor 1 (skipped variants contribute no counts).** A variant that cannot expand its bbox is evidence growth in that direction is *impossible*, and one over `MAX_REGION_TILES` is the tile-cap's own verdict — neither is evidence of hope, and the primary's refusal remains the operative signal.

**Refuted — minor 3 (blast radius beyond tier5 unmeasured).** The CI-way **full suite** ran green on every head of this PR — that is precisely a corpus-wide measurement: every SHA-256 layout-hash pin and every warning-population golden passed under the CI cache snapshot. An unmeasured-fixture golden trip is not a hypothesis left open; it is the thing the suite run tests.

**Declined — minor 2 (iter-jump in trace tooling) and the var-count nits.** The jump is the deliberate, documented marker of the bonus round; the var-count cost is one arithmetic pass plus a ≤64-entry Vec, and counting warm paths is what keeps growth cache-state-independent (the #654 class). Coverage nit: recorded on #652 (the sketch now covers the stop, the bonus round, and the deferred-exit suppression).

Receipts: comment-only delta on this push; CI-way suite green on the prior head (code-identical); tier5 probe byte-identical through all three rounds. Required check green, mergeStateStatus CLEAN — merging per the convergence rule: three rounds, each attacking previously-litigated ground from a new angle, zero surviving behavioral findings since round 2's deferred-exit fix.

PR #657, opened 2026-08-16
feat(router): #652 retry-with-teeth — forbidden-override conflict retry, wired and measured

storkme:
## Intent

Implement the retry-with-teeth design recorded on #652: give the ghost router's context-conflict branch a re-solve in which the conflicting committed tiles are forbidden **regardless of entity kind** — the naive strict-obstacles retry was a measured no-op because `refresh_forbidden`'s surface-belt exemption filtered the tiles back out and the zone cache replayed the identical solution.

This PR ships the mechanism plus the diagnosis it enabled. It does **not** yet reduce ac7's error count — the honest outcome and the reason are below, and the next lever is recorded on #652.

## Scope

- `GrowingRegion::forbidden_override` (new field): tiles inserted into `forbidden_tiles` unconditionally at the end of every `refresh_forbidden` — both the surface-belt exemption and the port exemption lose. Threaded through `from_crossings` / `solve_crossing`; empty for all pre-existing callers (fixture replay, blame diagnostics).
- ghost_router context-conflict branch: conflict check extracted into a closure; ONE retry with the conflict tiles as the override; new `CrossingConflictRetried { x, y, conflict_tiles, resolved }` trace event either way; cap-path skip reason now names the retry.
- Env-gated diagnostics under `SPAGHETTIO_DEBUG_CONFLICT_RETRY` (same pattern as `SPAGHETTIO_BLAME_JUNCTIONS`): primary conflicting entities vs committed, retry outcome, and a SAT pre-pass line (zone dims/vars/forced_empty/boundaries) — kept deliberately, they are the instruments for the follow-up unit.

Two pieces of the recorded design turned out **unnecessary**, which shrank the diff:
- **No cache-signature extension**: override tiles reach `CrossingZone::forced_empty`, which is already in the signature's `|F:` section — a retried zone is a distinct cached problem for free.
- **No port-tile special case**: a non-interior boundary on a `forced_empty` tile is documented encoder UNSAT (forces growth), and `topology_boundaries` re-expresses ports on forced tiles as *interior* boundaries when an item-matched entity sits there.

## Measured outcome (ac7-HS duty 0.6, the target fixture)

The retry fires on both conflicted clusters (seeds (15,93)/(15,96)) and engages end-to-end: fresh signature, `forced_empty` populated, 2 interior boundaries emitted, and a direct encoder call (local repro example) proves the retry zone **SAT** — with exactly the needed shape (deliver into the committed belts at (17,96)/(17,97), stamp nothing on them).

It still returns None in the pipeline, by a composition of three individually-correct mechanisms:
1. The conflict is a **flow-compatible identity mismatch**: the primary wants `underground-belt output East copper-cable` on a committed `transport-belt East copper-cable` — same item/direction/tier. Every feasible solution must either re-stamp a committed continuation belt (identity conflict) or deliver into it via the interior-boundary path.
2. On the interior-boundary retry, SAT's channel-fungible solution **reassigns which same-channel drop column carries which flow** (the zone has a 6-column copper fan in one channel); the walker veto correctly rejects the broken routed path (`flow:copper-cable:14 breaks at (14,94)`).
3. The veto's remedy is **growth**, which pushes the zone over `MAX_ZONE_SAT_VARS` → `sat_var_ceiling` cap.

Error population is byte-identical (17: 6 iso + 8 dead-end + 3 unresolved-junction). The recorded next lever (#652 comment): wire the existing, tested, currently-unwired `sat::solve_crossing_zone_with_pins` into the veto path — pin the vetoed path's committed in-zone tiles and re-solve the **same** zone instead of growing.

## Verification actually run

- CI-way full core suite (cache-pin protocol from `docs/test-suite-followups.md`): **CARGO_EXIT=0**. Every layout-hash pin and warning golden holds — this doubles as the corpus receipt that no shipped layout changes anywhere (a successful retry would flip a hash pin).
- Clippy clean on touched files (pre-commit hook + `--all-targets` run).
- ac7 probe (`i652_ac7_error_classify`, local example): error classes byte-identical before/after; retry firing + failure mode confirmed via the new trace event and env-gated diagnostics.
- Direct encoder repro (`i652_retry_zone_repro`, local example): retry zone SAT at cap None (30 entities, ~4ms), UNSAT at caps 0/1/2 — the ladder's lower rungs are legitimately refusing.

## Deviations

- The recorded design said "cache bypass or signature extension" — neither shipped, because the signature already covers `forced_empty` (explained above; the design comment in the conflict branch documents it).
- No new pin test: behavior is corpus-invariant by the suite receipt, and the retry's success case has no exhibiting fixture yet. The follow-up pin-re-solve unit is where a behavioral pin lands (it will have an exhibiting fixture: ac7).

### DIFF
--- a/crates/core/src/bus/ghost_router.rs
+++ b/crates/core/src/bus/ghost_router.rs
@@ -3022,20 +3022,19 @@ pub fn route_bus_ghost(
         // benign dedup (skipped at stamping); CONFLICTING overlaps
         // reject the whole cluster into the cap/retry machinery.
         //
-        // NO local re-solve is attempted (review on the fail-safe PR
-        // measured the obvious one as a deterministic no-op): adding the
-        // conflict tiles to `strict_obstacles` is filtered back out by
-        // `refresh_forbidden`'s surface-belt exemption (SAT is allowed
-        // to re-stamp surface belts — the benign identical re-derivation
-        // depends on it), and the zone cache's signature excludes the
-        // obstacle sets, so an unchanged zone replays the identical
-        // cached solution. A retry with teeth needs a
-        // forbidden-regardless-of-kind override threaded through
-        // `GrowingRegion` plus a cache-signature extension — recorded on
-        // #652. Until then a conflicted cluster caps for the
-        // layout-level retry and, failing that, is owned by the
+        // On conflict, ONE retry-with-teeth is attempted (#652): the
+        // conflicting tiles are re-solved as a `forbidden_override` —
+        // forbidden regardless of entity kind, so neither
+        // `refresh_forbidden`'s surface-belt exemption nor the port
+        // exemption filters them back out (the exemptions are what made
+        // the naive strict-obstacles retry a measured no-op, together
+        // with cache-signature replay — the override reaches the zone's
+        // `forced_empty`, which IS in the signature, so the retried
+        // zone is a distinct cached problem). A cluster whose retry
+        // fails (no solution, or a solution that still conflicts) caps
+        // for the layout-level retry and, failing that, is owned by the
         // unresolved-junction reporter + the fail-sever pass.
-        let Some(sol) = junction_solver::solve_crossing(
+        let Some(mut sol) = junction_solver::solve_crossing(
             cluster.as_slice(),
             &keys_at_tile,
             &routed_paths,
@@ -3049,6 +3048,7 @@ pub fn route_bus_ghost(
             &entities,
             strategies,
             &pending_crossings,
+            &FxHashSet::default(),
         ) else {
             // Capped: `solve_crossing` returned `None`, matching the
             // `JunctionGrowthCapped` event it emitted at this cluster's seed.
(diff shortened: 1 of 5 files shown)


storkme:
**Round-1 adjudication.** Two kernels bought (pushed on this head), the rest declined or refuted with receipts.

**Bought — major 1, the coverable part (zero test coverage).** A `GrowingRegion`-level unit test now pins the override's semantics: `forbidden_override_defeats_exemptions_and_tracks_growth` asserts (control) that the port + surface-belt exemptions win without an override, that an override tile in-bbox defeats BOTH exemptions, that out-of-bbox override tiles contribute nothing, and that growth re-inserts them when the bbox expands over them. The `resolved == true` commit path stays unpinned deliberately: no exhibiting fixture exists (that is the PR's own finding), and the follow-up pin-re-solve unit recorded on #652 is where the behavioral pin lands — it will have ac7 as its exhibiting fixture.

**Bought — minor 3 (mixed provenance in the failed-retry trace).** Correct: `conflicts` held the retry's tiles while `tap_x/y` reported the primary footprint. The reason string now names which attempt the sampled tiles came from and includes the retry footprint when one exists; `tap_x/y` deliberately keep reporting the primary footprint (that is the committed geometry snapshot forensics will find).

**Refuted — major 2 ("wired and measured" contradicted by the no-op outcome).** The no-op outcome on ac7 is not an admission the review extracted — it is the PR's stated content (Intent section, second paragraph; the Measured outcome section; the commit message). "Measured" means the claim ships with its own falsification receipts. A finding that restates the PR's disclosure as a contradiction is not actionable.

**Bought in revised form — major 3 (cap/zone accounting).** The observation was right and my first-pass remedy assessment was wrong; the session-side adversary supplied the decisive evidence this round lacked: the junction debugger (`web/src/ui/junctionTrace.ts`) groups growth series **purely by seed tile**, so an unmuted retry doesn't inflate counts alongside the primary — it **overwrites the primary's records in place** and the cluster outcome becomes whichever terminal event landed last. Not muting corrupts exactly the forensics it was meant to preserve. Fixed per the `blame_unsolvable_cluster` precedent: the retry solve is now wrapped in `trace::with_muted`; observability stays via the `CrossingConflictRetried` summary event + the env-gated diagnostics (eprintln, unaffected by muting). Trace accounting thereby reverts to the pre-PR shape entirely.

**Declined — minor 1 (retry cost).** Bounded by construction: one retry per conflicted cluster, inner growth capped by `MAX_GROWTH_ITERS` + the var ceiling, and conflicts are rare (two clusters on ac7, zero on the rest of the corpus). "Cap immediately" — the pre-PR behavior — is not free either: it ships the cluster to the layout-level retry and the fail-sever pass. The cost is the recorded design's price and it bought the diagnosis.

**Declined — minor 2 (in-situ commit bypasses the layout-level retry).** The retry's solution passes exactly the screening any PRIMARY solution passes — the same `conflicts_of` closure (same claim classes), plus the walker veto and deferred-exit checks inside `solve_crossing`. Pre-PR primary commits were no stricter; the retry does not widen any blind spot, it narrows one (a class of clusters that previously always capped can now commit a solution that cleared the identical checks).

**Nits.** The forced-zone boundary dump is env-gated diagnostics by design and prints for any forced zone, not only retries — retry zones are identified by sequence (they follow a `PRIMARY-CONFLICT` line); narrowing the gate would cost visibility into non-retry forced zones for no measured gain. Gold-hashed fixtures going red if a retry ever succeeds is the pin system working as intended, and the PR body's Deviations section already names it.

---

**Session-side adversarial review (required for layout-engine changes): SHIP-WITH-FIXES, absorbed.** Independent worktree review of 20de4e54. It re-ran the CI-way suite itself (26/26 ok — and independently confirmed `CrossingConflictRetried` fires **zero** times across the corpus, hardening the invariance claim beyond my receipt), verified the override threading structurally (single implementation inside `refresh_forbidden` covers all four call sites; `Clone` carries the override through both the variant and eviction clones; bbox-assignment-before-refresh makes growth-then-cover correct), verified retry-loop safety by direct read, ran the wasm32 check (exit 0; `std::env::var` is always `Err` on wasm so the diagnostics are dead code client-side), grep-verified all eight eprintln sites are gated, and confirmed all three design-comment claims against `zone_cache.rs`/`sat.rs`/`junction_sat_strategy.rs`. Its one MAJOR — the junction-debugger per-seed collision above — is fixed on this head. Scope note it flagged honestly: it did not re-run the ac7 local examples (the SAT-solves-but-walker-vetoes narrative), which stand on this session's own probe receipts.

Receipts on this head (b159a2af): CI-way full suite green (`CARGO_EXIT=0`) at each of the three commits; new `GrowingRegion` unit test green; clippy clean on touched files.

storkme:
**Round-2 adjudication — and convergence.** Two new kernels, both absorbed (one comment, one recorded tail); the rest re-litigates ground round 1 and the session-side adversary already settled, with no new evidence. Merging on the green head per the convergence rule.

**Bought — minor 4 (stale primary solved-event on a resolved retry).** New and correct: on the (currently unreachable) success path, the primary's unmuted solve series describes the discarded geometry. The reconciliation rule is now stated at the muting site: `JunctionCommitted` carries the committed (= retry's) entity list, `CrossingConflictRetried{resolved: true}` marks the provenance switch, and per-strategy retry forensics stay in the env-gated diagnostics until the #652 veto→pin unit gives retries a success path worth first-class tracing (also answers the nit).

**Bought as recorded tail — minor 7 (fixture replay cannot reproduce a retry).** Correct: `dump_region_fixture` records no override, so a retry zone can't be replayed. Recorded on #652 as part of the veto→pin unit's tooling scope — that unit is the work that will need the replay path, and a fixture-format change doesn't belong in a PR whose retry never yet succeeds.

**Refuted — majors 1 & 2 and minor 5 (retry never succeeds / success path untested / "latency adder, not a fix").** These restate the PR's own Intent and Measured-outcome sections — the no-op outcome on ac7 is this PR's published finding, not an extracted admission. On major 2's specific claim that the success path "commits a different footprint... handled only implicitly": there is no novel commit machinery — `sol` is rebound before the one shared stamping path every primary solution takes, and that path reads `sol.footprint` wherever it points; nothing downstream is footprint-shape-dependent. A synthetic end-to-end fixture that manufactures a conflicting-then-resolvable zone through the full router would cost more than the follow-up unit that makes ac7 itself the exhibiting fixture — the behavioral pin lands there, as the PR's Deviations section already commits to.

**Declined — minor→major 3 (resolved-retry cascade).** The stale-`junction_hard`-snapshot blind spot is the pre-existing, #653-documented architecture for ALL zone commits; the conflict check is its designed mitigation and every retry commit passes the identical screen. The retry participates in that mechanism, it doesn't create it — and each cluster's retry is bounded to one. Worth re-examining when retries start succeeding; noted for the pin unit.

**Refuted — minor 6 (debug flag floods, "fires in WASM").** The WASM half is factually wrong: `std::env::var` always returns `Err` on wasm32-unknown-unknown, so the branch is dead code client-side — verified by the session-side adversary's wasm32 check on this PR. The native half was adjudicated in round 1: the pre-pass line is one eprintln per invocation and the boundary dump fires only for forced zones, which are rare; the flag is a diagnostics gate, not a retry marker.

**Declined — minor 8 (dead cache weight).** One Unsat/Timeout record per conflicted-cluster retry signature, in an append-only cache that accumulates dead records across layout changes by documented design. The requested "confirm byte-identical on a conflict-carrying layout" already exists: the ac7 probe measured identical error populations warm and cold across every head of this PR, and both suite runs (mine and the adversary's independent one) pin the corpus.

Receipts on this head (eac2ac4c): comment-only delta over b159a2af; CI-way suite green at every code-bearing head; adversary's independent suite run 26/26 ok with zero `CrossingConflictRetried` firings corpus-wide. Three review rounds (bot ×2 + session-side adversary), zero surviving behavioral findings since the muting fix — converged.

PR #658, opened 2026-08-16
feat(router): #652 flow-compatible commit upgrade — UG outputs may replace their committed continuation belts

storkme:
## Intent

Eliminate the ac7 context-conflict class — the flip-gating defect where two crossing clusters could never commit. The #653 identity-based conflict check made the shape structurally uncommittable: every feasible solution must place a UG **output** exactly ON a committed surface belt that already carries the identical flow (same item, direction, tier) — the committed belt IS the flow's continuation into the neighboring, earlier-committed crossing zone.

Both alternatives were implemented and **measured structurally unsolvable** before this carve-out was chosen (receipts on #652):
- The forbid-and-grow retry (#657) engages end-to-end but the override makes the zone grow into the var ceiling.
- The recorded veto→pin re-solve was prototyped this session and falsified by encoder bisect: pinning any committed east-row belt UNSATs the zone (the EC crossing then has no dive column under the straight-feed UG rules), and the walker-clean topology provably requires the UG to surface on the committed tile. The prototype is not in this diff; design + falsification are recorded on #652.

## Scope

- `flow_compatible_ug_upgrade` (ghost_router.rs): the one accepted shape — solution UG output over committed plain surface belt, matching direction/item/tier (name-prefix compare). Conservative guards, each `false` ⇒ keep the conflict: behind-tile must be inside the solution footprint (nothing can jam against the UG exit's back — the zone re-derives that upstream); no committed belt-like entity outside the footprint may output into the tile from a side (sideloads onto a UG exit lose the near-side lane).
- `conflicts_of` carve-out: flow-compatible tiles are not conflicts (closure now takes the footprint as an argument so the retry path evaluates against its own footprint).
- Stamping upgrade path: `Occupancy::release_tile` (new; orphan semantics per `place()`'s existing note) + purge of the superseded belt from the entity list, then normal placement. The old Template/RowEntity skip would have stranded the solution's flow underground (kept UG-in, vanished UG-out).
- e2e: the ac7 fail-safe pin gains a `context_conflicts == 0` assertion; doc comment updated (liveness now carried by the remaining mega-cluster).

## Measured outcome (ac7-HS duty 0.6)

| | before | after |
|---|---|---|
| total errors | 17 | **14** |
| unresolved-junction | 3 | **1** |
| context-conflict skips | 2 | **0** |
| zones committed | 24 | **25** |
| severed feeders | 9 | 7 |

Remaining on the fixture: 6 balancer-over-trunk isolation errors + the 21-tile iter-capped mega-cluster near (15,123) (owns the 7 dead-ends) — both pre-existing classes tracked on #652/#644.

## Verification actually run

- CI-way full core suite (cache-pin protocol): **CARGO_EXIT=0**, twice (pre- and post-pin-extension). Corpus exactness argument: the #657 session-side adversary independently proved zero context-conflicts fire anywhere in the CI corpus, so the carve-out has nothing to change there — and every layout-hash pin and warning golden held.
- Extended ac7 pin test green in isolation (CI-way).
- Probe receipts (`i652_ac7_error_classify`, local): the table above; both formerly-conflicted clusters commit; layout shrinks slightly (3407 → 3397 entities) because the layout-level conflict retry no longer fires.
- Encoder bisect receipts (`i652_retry_zone_repro`, local): pin-UNSAT localization that falsified the pin design.
- Clippy clean on touched files.

## Deviations

- The unit this branch set out to build (veto→pin re-solve, recorded on #652) is deliberately absent: its target shape was falsified mid-implementation and it would have shipped with zero exhibiting fixtures. Falsification receipts and the code sketch are on #652 for the day a pinnable shape appears.
- No standalone unit test for the predicate: the load-bearing receipt is the e2e assertion on the exhibiting fixture plus the corpus invariance; a synthetic `Occupancy` harness would restate the predicate's own arms.

### DIFF
--- a/crates/core/src/bus/ghost_occupancy.rs
+++ b/crates/core/src/bus/ghost_occupancy.rs
@@ -359,6 +359,17 @@ impl Occupancy {
         Ok(idx)
     }
 
+    /// Drop the claim at `tile`, orphaning its entity (see `place`'s
+    /// orphan note — the backing entity stays in `self.entities` for
+    /// index stability but no claim points to it). Returns true if a
+    /// claim was removed. Used by the flow-compatible commit upgrade
+    /// (#652): a crossing-zone solution may replace a committed
+    /// same-item, same-direction, same-tier surface belt with the UG
+    /// output that continues the identical flow.
+    pub fn release_tile(&mut self, tile: (i32, i32)) -> bool {
+        self.claims.remove(&tile).is_some()
+    }
+
     /// Drop every `GhostSurface` claim whose tile lies inside `zone`.
     /// The corresponding entities become orphaned (see `place`'s note).
     /// Used by SAT before claiming tiles for its own solution.
--- a/crates/core/src/bus/ghost_router.rs
+++ b/crates/core/src/bus/ghost_router.rs
@@ -3080,45 +3080,64 @@ pub fn route_bus_ghost(
             }
             continue;
         };
-        let conflicts_of = |sol_entities: &[PlacedEntity],
-                           footprint: crate::bus::junction::Rect|
-         -> Vec<(i32, i32)> {
-            sol_entities
-                .iter()
-                .filter(|ent| {
-                    let tile = (ent.x, ent.y);
-                    let claimed = matches!(
+        // Classify each solution entity landing on a committed
+        // Template/RowEntity tile: identical → benign dedupe (neither
+        // list), flow-compatible UG-output-over-continuation-belt on a
+        // Template claim → upgrade (see `flow_compatible_ug_upgrade`;
+        // RowEntity targets are never upgrades — row belts live in
+        // `row_entities` with no eviction path), anything else →
+        // conflict. Evaluated ONCE here — the stamping loop consumes
+        // the approved `upgrades` set instead of re-evaluating, because
+        // the release pass mutates occupancy between the two sites and
+        // a re-evaluation can flip (measured: a stamp-time flip dropped
+        // a UG output and shipped unpaired-UG + belt-loop errors).
+        let classify = |s: &junction_solver::JunctionSolution| -> (
+            Vec<(i32, i32)>,
+            FxHashSet<(i32, i32)>,
+        ) {
+            let mut conflicts: Vec<(i32, i32)> = Vec::new();
+            let mut upgrades: FxHashSet<(i32, i32)> = FxHashSet::default();
+            for ent in &s.entities {
+                let tile = (ent.x, ent.y);
+                let is_template = matches!(
+                    occupancy.claim_at(tile),
+                    Some(crate::bus::ghost_occupancy::Claim::Template { .. })
+                );
+                let claimed = is_template
+                    || matches!(
                         occupancy.claim_at(tile),
-                        Some(crate::bus::ghost_occupancy::Claim::Template { .. })
-                            | Some(crate::bus::ghost_occupancy::Claim::RowEntity { .. })
+                        Some(crate::bus::ghost_occupancy::Claim::RowEntity { .. })
                     );
-                    claimed
-                        && occupancy.entity_at(tile).is_some_and(|existing| {
-                            let differs = existing.name != ent.name
-                                || existing.direction != ent.direction
-                                || existing.io_type != ent.io_type
-                                || existing.carries != ent.carries;
-                            // #652 flow-compatible carve-out: a UG output
-                            // continuing a committed belt's identical flow
-                            // is an upgrade, not a conflict (see
-                            // `flow_compatible_ug_upgrade`). ac7 proof: the
-                            // walker-clean topology REQUIRES the UG to
-                            // surface exactly on the committed continuation
-                            // belt, and the pin/retry alternatives are
-                            // measured structurally unsolvable there.
-                            differs
-                                && !flow_compatible_ug_upgrade(
-                                    &occupancy,
-                                    existing,
-                                    ent,
-                                    footprint,
-                                )
-                        })
-                })
-                .map(|ent| (ent.x, ent.y))
-                .collect()
+                if !claimed {
+                    continue;
+                }
+                let Some(existing) = occupancy.entity_at(tile) else {
+                    continue;
+                };
+                let differs = existing.name != ent.name
+                    || existing.direction != ent.direction
+                    || existing.io_type != ent.io_type
+                    || existing.carries != ent.carries;
+                if !differs {
+                    continue;
+                }
+                if is_template
+                    && flow_compatible_ug_upgrade(
+                        &occupancy,
+                        existing,
+                        ent,
+                        s.footprint,
+                        &s.participating,
+                    )
+                {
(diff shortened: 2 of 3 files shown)


storkme:
**Round-1 adjudication (bot + session-side opus adversary, absorbed together on c49779f8).** Both reviews converged on the same soundness gap from different angles, and my first absorb attempt found a third failure mode the hard way — the full story is in the commit message; the short version:

**Root cause bought (bot majors 1–2, adversary majors 2–3, + a measured self-regression).** The upgrade predicate ran at TWO times against DIFFERENT occupancy states (the release pass mutates claims between conflict check and stamping). My first claim-aware strengthening flipped at stamp time and silently dropped an approved UG output — hermetic A/B caught it as 14 → 16 errors (unpaired-UG + belt-loop). Design is now **single-evaluation**: `classify` computes `(conflicts, upgrades)` once at check time; stamping consumes the approved set. The behind/side guards are claim-aware via a conservative `released_at` mirror of the release pass (participating-spec ghosts + trunk Permanents released; Template/RowEntity/SatSolved and preserved balancer/tapoff Permanents persistent). Note the reviews' "smallest fixes" were each individually insufficient: the bot's strict-persistence check and the adversary's direct behind-test both over-reject at check time (the ac7 behind tile legitimately holds a *releasable* participating ghost belt) — the release-mirror + single-evaluation combination is what survives both the soundness argument and the fixture.

**Bought — adversary major 1 (RowEntity duplicate).** Upgrades are `Claim::Template` only; row belts live in `row_entities` with no eviction path. **Bought — adversary major 3 (splitter second tiles)**: `entity_covering` resolves covering splitters via `splitter_second_tile`. **Bought — adversary minors 4–7**: doc splice fixed (the predicate had stolen `ug_endpoint_conflicts`' doc comment), e2e `JunctionCommitted > 0` vacuity guard added, factorio-mechanics citation corrected to a U7 analogy marked conservative, "REQUIRES" softened to "only walker-clean topology found (argued, not enumerated)", purge comment names both purged entities.

**Refuted — bot major 3 (same-item lane splicing).** Channel fungibility is the SAT model by design; per-path identity is guarded by the walker veto, which runs before the conflict check ever sees a solution — a splice that breaks any routed path's continuity never reaches commit. The adversary independently verified the walker-relevant behavior (0 back-fed UG outputs, 0 duplicate anchors on the shipped layout).

**Declined — bot minor 2 (retain keyed by coordinate).** The local entities vec is the shipping authority; a coordinate purge removes any would-be duplicate, and an absent entity means nothing would have shipped anyway. The adversary confirmed both entities at each upgrade tile (crossing belt + stale ghost orphan) are correctly purged.

**Adversary's not-verified list, answered where possible:** the in-footprint side-feed question is now moot (the guard no longer skips in-footprint neighbors — persistence decides, not footprint membership). The "unique topology" universal is softened in the doc as requested. The sim-anchor gap stands honestly: the four upgraded tiles are new shipped geometry with no ground-truth run; ac7 duty-0.6 sim anchoring is part of the #644 flip campaign's gate, where this fixture's family already has the RFC-069 sim protocol.

Receipts on c49779f8: hermetic probe (fresh CI-cache copy) reproduces the approved baseline exactly — 14 errors, 0 conflict skips; CI-way full suite green; ac7 pin green in isolation; clippy clean.

storkme:
**Round-2 adjudication.** Three kernels bought (pushed on 3ba11543), the major refuted with a concrete counter-mechanism, the rest declined with receipts.

**Refuted — major (forward tile never validated).** The proposed fix — reject when the forward tile is in-footprint and released — would wrongly reject upgrades whose forward flow the SOLUTION itself carries, and measurably regress the exhibiting fixture. The forward adjacency is owned by other layers, not this predicate: in-zone, SAT's own adjacency rules require a UG output's expelled flow to be received (a solution entity or an interior sink); at the zone edge, off-grid output is only legal at boundary ports, where the committed continuation persists by construction; and the walker re-walks every affected path through the composite shadow before any candidate reaches the conflict check. The session-side adversary's independent probe on the shipped layout (0 back-fed UG outputs, 0 duplicate anchors, all layout-level receipts reproduced) is the empirical half. This ownership split is now stated in the predicate's doc (also answering minor 4).

**Bought — minor 1 (pipe trunks).** Correct and in the unsafe direction: the release pass refuses pipes regardless of segment, so a `trunk:` pipe persists and `released_at` must not count it released. Fixed.

**Bought — minor 3 (double-apply defence).** The stamping loop now CONSUMES the approved set (`remove`, not `contains`) — a hypothetical second solution entity at the same tile falls through to the ordinary skip path instead of re-applying the release. (Unreachable today — `place` would panic first — but the defence costs one word.)

**Bought — nit (negative-only pin).** The e2e pin gains its positive half: `unresolved <= 1` (pre-upgrade the fixture shipped 3). A silently-dead carve-out now trips a count, not just an absence, and the `JunctionCommitted > 0` vacuity guard from the previous round covers the trace-stream side.

**Declined — minor 2 (predicate blind to solution self-tiles).** Structurally true and deliberately so — the predicate reads committed occupancy; solution-internal shapes are the encoder's and walker's jurisdiction (the round itself notes the walker prevents this). Now documented at the predicate.

**Declined — minor 5 (`proposed_tiles` dead weight).** Pre-existing, untouched by this PR; noted for a housekeeping pass, not bought here.

Receipts on 3ba11543: hermetic probe unchanged (14 errors, 0 conflict skips, 7 severed); CI-way full suite green; ac7 pin green in isolation; clippy clean. Two rounds plus the session-side adversary, every behavioral finding absorbed or refuted with a counterexample — if round 3 attacks only previously-litigated ground, this merges per the convergence rule.

storkme:
**Round-3 adjudication — and convergence.** Zero majors this round; the one factually-new kernel is bought as a doc truth-up on 8526b63f (comment-only delta): `released_at` **under-approximates** the release pass — the actual commit passes `None` and clears ALL in-zone ghosts, while the mirror counts only participating specs' ghosts, deliberately stricter (relaxing to match is recorded on #652 as a nicety with no exhibiting fixture; 100% conservative direction, as the finding itself notes). The classify/stamp agreement's load-bearing invariant (the release pass never drops Template/RowEntity claims) is now stated at the guard that depends on it, per the round's minor 5 — the documentary pin without a new assert on a today-unreachable divergence.

Everything else re-litigates settled ground: forward tile (round-2 major, refuted with the encoder-discipline + boundary-port + walker ownership argument, and its proposed fix measurably regresses the fixture), lane identity (round-2's splicing claim — the walker veto IS the per-path identity guard), retain breadth (round-1, shipping-authority argument + adversary's receipt that both purged entities are correct), pin coupling (the assertions carry their own re-point instructions).

Convergence: three bot rounds + session-side opus adversary; every behavioral finding absorbed (single-evaluation design, Template-only, claim-aware guards, splitter coverage, pipe persistence, consume-don't-test, positive + vacuity pins) or refuted with a counterexample; last two deltas comment-only/doc-only in effect with hermetic receipts unchanged (14 errors, 0 conflict skips). Merging on the green check.

PR #659, opened 2026-08-17
fix(balancer): #652 lever 1 — reserve the balancer stamp's WIDTH, not just its height

storkme:
## Intent

#652's lever 1 (obstacle-aware balancer placement), the pick-up its last comment recommended doing first. `family_stamp_plan` guards a template's width against its output count on the **decomposed** path (`sub.width > sub_m` → skip, with a comment explaining that neighbouring sub-stamps would otherwise overlap) and on the **generated** path (`generated.width <= m`). The `Direct` path — the exact-shape library hit, which is what most families take — has never had that guard, and **31 of the library's templates are wider than their output count**.

Every stamping path aligns by `balancer_origin_x` (output tile 0 onto `lane_xs[0]`), so such a template spills west of the family's first trunk column and east of its last. The lane planner has always reserved the stamp's **height** (`balancer_y_end`, so trunks skip the zone) and never its **width** — so the spill lands on whichever family was assigned the neighbouring columns.

On ac7-HS at duty 0.6 (the RFC-069 flip shape) electronic-circuit's `(5,2)` is width 6 over 2 trunk columns — 1 west, 3 east — and the eastern spill swallowed the plastic-bar trunk at x=18–19, severing it and stranding fragments inside the template's holes.

## Scope

- **In:** `balancer::family_stamp_x_pad` derives the `(west, east)` spill from a family's stamp plan; the lane planner reserves it as empty columns flanking the family's block. `bus_width_for_lanes` now also takes the families (the reservation works by shifting the lanes that *follow* a family, and the eastmost family has none to shift). 3 new unit tests. The ac7 pin rewritten. One golden re-blessed. `sim_export --row-layout`.
- **Out:** #652's lever 2 (mega-zone resolution: zone-split / channel-reduction / ceiling-aware decomposition). ac7's exhibitor of that class **disappears** with this change — its 21-tile iter-capped cluster near (15,123) resolves once plastic no longer has to route around the balancer wall — so lever 2 now has tier5 as its only fixture, exactly as #652 predicted when it recommended this ordering.
- **Out:** the RFC-069 default flip itself. This unblocks ac7's half; the flip is #644's call.
- **Size:** 277 added, comfortably under the norm.

## Verification

- [x] `cargo test --manifest-path crates/core/Cargo.toml` — green in one clean invocation, zero failures across all 20 test binaries.
- [x] `cargo clippy --workspace -- -D warnings` — the CI gate. (`--all-targets` fails on pre-existing `power_wires.rs` debt in files this PR does not touch; ci.yml is lib-only *on purpose* and says so.)
- [x] WASM rebuild green.
- [x] **Measured, not eyeballed** — the protocol's step 2.

**ac7-HS duty 0.6** (`i652_ac7_error_classify`, a rebuilt local probe — the original was lost with the untracked `examples/` dir):

| | pre | post |
|---|---|---|
| errors | **14** (7 belt-dead-end, 6 belt-item-isolation, 1 unresolved-junction) | **0** |
| warnings | 91 (IRD 77, reach 10, path 4) | 16 (IRD 14, reach 2) |
| crossings severed | 7 | 0 |
| junction growth caps | 1 (`iter_cap`) | 0 |
| **junctions committed** | 25 | **32** |
| size | 111×161, 3398 ent | 118×160, 3634 ent (+7%) |

The committed-junction **increase** is the load-bearing number: it rules out "a check stopped discriminating" ([validator-reporting.md](../docs/validator-reporting.md)), because silencing a check cannot make the router solve seven more crossings.

**Meter A/B** (`check_one`, same instrument both arms):

```
pre-fix   produced {}  — NOTHING          note: "175 tiles in a belt cycle"
post-fix  advanced-circuit 4.50/7.0 = 64.3% of plan, uniform down the
          chain (cable 45/70, EC 9/14, plastic 9/14)   note: "12 tiles"
```

Producing zero is exactly what a severed plastic trunk predicts — plastic is an AC ingredient, so the chain stalls. **The post-fix reading is still below plan and I believe it**: this removes the structural break, it does not bring the fixture to plan. Absolute values carry the meter's known fluid gap (#570) on this crude-oil chain (petroleum-gas is absent from `produced` in *both* arms), so the claim rests on the delta, not the level.

- [ ] **Sim anchor running** (432k warmup, `--timeout-secs 3600`). Result will be posted here. Not merging on the meter alone — the meter refutes, it never clears.

## Deviations from intent

Three, all recorded in the commit message:

1. **#652's record corrected.** Its last comment called the sixth isolation error at (18,76) "separate and small … not part of this complex". It is the same defect — copper-cable's `(6,7)`, width 10 over 7 columns, spilling onto the same plastic column. All six were one cause.
2. **RFC-069's tier5 receipt is not comparable** to a `build_bus_layout` measurement of the same fixture (different export path). Measured both ways here, tier5 at duty 0.6 is **identical** with and without this change (175×255, 7057 entities, 18 belt-dead-end, 320 warnings) — this lever does not touch the tier5 half.
3. **`sim_export --row-layout` added** (not in the original scope). The tracked generator could not express a horizontal-stack fixture at all, so the corpus's HS class had no supported route to a blueprint+manifest pair and *could not be sim-anchored*. Without it the verification protocol's step 2 is unreachable for this change.

## Costs and lost coverage — stated, not blessed away

- **tier5 at the shipping default gains one `belt-detour` warning** (105 tiles for a 52-tile separation) and +160 entities / +4 columns. Errors stay 0, IRD unchanged at 13. The check is diagnostic-only and never promotes to Error. The mechanism — the reservation widens the bus and lengthens a copper-cable path — is **inferred from the width delta, not proven at tile level**. That is the one claim here I have not nailed down.
- **The fail-sever pass is now uncovered.** The ac7 pin carried the only liveness assert for #655's machinery (`severed > 0 || unresolved > 0`), and it is gone because the fixture no longer severs. A 7-fixture sweep on the same code path found ac7 was the **only** exhibitor before this change and **none remain** after. Recorded in the test's own doc comment; synthetic pin tracked on #652. The sweep is not the whole corpus and is not claimed to be.
- **A ceiling that has stopped discriminating:** `processing_unit_2s_am2_fast_belts_validation_baseline`'s `("unresolved-junction", 4)` is an upper bound measuring 0 on the probe path. Not touched here — its ceiling and my probe are different pipelines, so that 0 is not a licence to tighten it — but it is the exact class validator-reporting.md warns about, so it is reported on #652.

## Models / contracts touched

Balancer stamping geometry. The contract that changed is implicit and now explicit: **a family's reserved columns must cover its stamp's full width, not just its output count.** Documented at `family_stamp_x_pad` and pinned by `family_stamp_x_pad_covers_every_library_stamp`, which asserts every library shape's stamped tiles fall inside the reserved span and carries two anti-vacuity guards — one that the loop stamped something, one that some shape actually spills, so a pad of `(0,0)` cannot pass it.

### DIFF
--- a/crates/core/examples/sim_export.rs
+++ b/crates/core/examples/sim_export.rs
@@ -29,6 +29,7 @@
 //!   --di off|candidate|forced        direct insertion (default candidate)
 //!   --claim up|down|search           DI claim order (default: engine default)
 //!   --belt <entity>        max belt tier (default: engine picks by rate)
+//!   --row-layout <kind>    native (default) | horizontal-stack
 //!   --quality <name>       normal|uncommon|rare|epic|legendary (default normal)
 //!   --stacking <1..4>      belt stacking (default 1)
 //!   --research-productivity <recipe=bonus,...>   declared research
@@ -53,7 +54,7 @@
 
 use rustc_hash::FxHashSet;
 use spaghettio_core::bus::di_cell::{DiClaimOrder, DirectInsertion};
-use spaghettio_core::bus::layout::{build_bus_layout, LayoutOptions};
+use spaghettio_core::bus::layout::{build_bus_layout, LayoutOptions, RowLayout};
 use spaghettio_core::common::QualityTier;
 use spaghettio_core::recipe_db::MachinePalette;
 use spaghettio_core::validate::{self, LayoutStyle, Severity};
@@ -140,6 +141,7 @@ fn main() {
     let mut research_productivity: std::collections::BTreeMap<String, f64> =
         Default::default();
     let mut inserter_cap: Option<u8> = None;
+    let mut row_layout = RowLayout::default();
     let mut inputs: Vec<String> = DEFAULT_INPUTS.iter().map(|s| s.to_string()).collect();
     let mut label: Option<String> = None;
     let mut out = std::env::var("SIM_PROBE_OUT").unwrap_or_else(|_| "/tmp".to_string());
@@ -169,6 +171,19 @@ fn main() {
                 })
             }
             "--belt" => belt = Some(need(i)),
+            // #652: without this the tracked generator cannot express a
+            // horizontal-stack fixture at all, so the corpus's HS class
+            // (ac7-HS and friends) had no supported route to a
+            // blueprint+manifest pair and could not be sim-anchored.
+            "--row-layout" => {
+                row_layout = match need(i).as_str() {
+                    "native" | "default" => RowLayout::default(),
+                    "horizontal-stack" | "hs" => RowLayout::HorizontalStack,
+                    other => usage(&format!(
+                        "--row-layout must be native|horizontal-stack (got {other})"
+                    )),
+                }
+            }
             // RFC-069 Phase 1: planning-duty knob for the K69-1 sim A/B.
             "--duty" => {
                 duty = need(i)
@@ -279,6 +294,7 @@ fn main() {
         direct_insertion: di,
         max_belt_tier: belt,
         planning_duty: duty,
+        row_layout,
         quality,
         stacking,
         research_productivity,
--- a/crates/core/src/bus/balancer.rs
+++ b/crates/core/src/bus/balancer.rs
@@ -598,11 +598,47 @@ mod tests {
 
     /// A merge-tap family's merge tree puts its single output at the
     /// tree's EAST edge, so the whole body is a western spill.
+    ///
+    /// Check-stamps every tree size rather than asserting the pad value
+    /// on one isolated family (round-2 review: the value alone leaves
+    /// the "output at east edge, body all-west" geometry premise
+    /// unverified, so a merge-tree change would ship an adjacent-trunk
+    /// overlap unnoticed — and the shape sweep below cannot reach this
+    /// branch because it builds `merge_tap: false` families).
     #[test]
     fn family_stamp_x_pad_covers_the_merge_tree() {
-        let mut fam = library_family(4, 1, 20);
-        fam.merge_tap = true;
-        assert_eq!(family_stamp_x_pad(&fam), (3, 0));
+        assert_eq!(
+            family_stamp_x_pad(&{
+                let mut f = library_family(4, 1, 20);
+                f.merge_tap = true;
+                f
+            }),
+            (3, 0)
+        );
+
+        for n in 1..=10u32 {
+            let mut fam = library_family(n, 1, 100);
+            fam.merge_tap = true;
+            let (west, east) = family_stamp_x_pad(&fam);
+            let ents = stamp_merge_tap_family(&fam, None, &StackingCtx::unstacked());
+            assert!(!ents.is_empty(), "merge tree n={n} stamped nothing");
+            let trunk = fam.lane_xs[0];
+            let (lo, hi) = (trunk - west, trunk + east);
+            for e in &ents {
+                let spans_x = e.name.contains("splitter")
+                    && matches!(e.direction, EntityDirection::North | EntityDirection::South);
+                let e_hi = if spans_x { e.x + 1 } else { e.x };
+                assert!(
+                    e.x >= lo && e_hi <= hi,
+                    "merge tree n={n}: {} at x {}..{} escapes the reserved \
+                     span {lo}..{hi} (pad west={west} east={east}) — a K-group \
+                     sibling trunk would be overlapped",
+                    e.name,
+                    e.x,
+                    e_hi
+                );
+            }
+        }
     }
 
     /// #652: the reservation must cover the stamp's ACTUAL extent for
--- a/crates/core/src/bus/balancer.rs
+++ b/crates/core/src/bus/balancer.rs
@@ -661,9 +661,22 @@ mod tests {
     /// happened never to spill.
     #[test]
     fn family_stamp_x_pad_covers_every_stamped_shape() {
+        // The grid bound is DERIVED from the library, not hardcoded
+        // (round-3 review). Today every key is within 1..=10, so a
+        // literal 10 would sweep everything — but a regeneration that
+        // adds an (11, k) would then ship un-swept, which is precisely
+        // the case the reviewer was worried about. Deriving it means
+        // the sweep grows with the library instead of silently falling
+        // behind it.
+        let hi = crate::bus::balancer_library::balancer_templates()
+            .keys()
+            .flat_map(|&(n, m)| [n, m])
+            .max()
+            .unwrap_or(10)
+            .max(10);
         let mut checked = 0usize;
         let mut spilling = 0usize;
-        for (n, m) in (1..=10u32).flat_map(|n| (1..=10u32).map(move |m| (n, m))) {
+        for (n, m) in (1..=hi).flat_map(|n| (1..=hi).map(move |m| (n, m))) {
             let fam = library_family(n, m, 100);
             let (west, east) = family_stamp_x_pad(&fam);
             let ents = stamp_family_balancer(&fam, None, &StackingCtx::unstacked())
--- a/crates/core/src/bus/balancer.rs
+++ b/crates/core/src/bus/balancer.rs
@@ -200,6 +200,64 @@ pub(crate) fn family_stamp_plan(
     }
 }
 
+/// Columns a family's stamp needs BEYOND its own trunk columns, as
+/// `(west, east)`.
+///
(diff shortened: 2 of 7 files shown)


storkme:
## Sim anchor — done, and it is a single-instrument A/B

Both arms: headless Factorio, `--warmup 432000 --timeout-secs 3600`, `--research-productivity plastic-bar=0.10` declared on both, same fixture config, exported through `sim_export --row-layout horizontal-stack`.

| ac7-HS duty 0.6 | **pre-fix** | **post-fix** |
|---|---|---|
| advanced-circuit | **0.00/s — −100.0%** | 4.51/s — −35.6% |
| copper-cable | 0.00/s | 44.87/s |
| electronic-circuit | 0.00/s | 9.00/s |
| plastic-bar | 0.00/s | 8.97/s |
| converged | **false** | true |
| items in window | **0** (floor 300) | 302 |
| validator | 14E / 91W | 0E / 16W |
| machine census | 49 full_output, 63 ingredient-short, **0 working** | 13 full_output, 24 short, **75 working** |
| verdict | FAIL | FAIL |

**The pre-fix layout produces literally nothing in real Factorio.** Not "below plan" — zero, on every stage, with no machine working. That is the severed plastic trunk: 49 producers backed up with nowhere to send, 63 consumers starved. It independently confirms the meter's `produced {}` reading, and it makes the improvement claim single-instrument rather than meter-vs-sim.

**Both arms still FAIL, and the post-fix number is the honest one to carry forward: 64.4% of plan.** This change removes a structural break; it does not make ac7-HS at duty 0.6 a working factory. The deficit is uniform across every stage (−35.6% to −36.0%), which per `sim-harness-forensics.md`'s reading means one shared constraint propagating, not five independent problems — consistent with the 24 ingredient-short machines. That residual is #644/RFC-069 territory, not this PR's.

**Calibration note worth banking:** the fast meter read 64.3% where the sim reads 64.4% on the post-fix artifact. The meter's contract is "below plan ⇒ believe it; at plan ⇒ evidence of nothing", and here the below-plan half held to within 0.1pp on a fluid chain (`meter-divergence.md` may want this row — the meter's known fluid gap did not distort the delta).

### Process note on the first attempt

The first anchor ran to completion and was **refused**: `research-productivity parity: 'plastic-bar' realized 0.10000000149012 but the manifest declares 0`, verdict forced `NO DATA`. That is `handoff-meter-as-gate-2026-08-07.md` §3 trap 2 exactly, including "refuses runs ~16 minutes in". The install's realized set (confirmed against three banked reports) is `plastic-bar +10%`, `processing-unit +10%`, everything else 0. Nothing from that run is quoted here. Re-exported with the axis declared and re-run — the numbers above are all from clean, kit-error-free runs.

storkme:
## Round-4 adjudication

Seven of the eight items are already-answered material. One is refuted by
measurement, one is measured and does not occur, and one is valid and tracked.
Nothing here changes the diff.

### [major] pass-2 width growth — declined, fourth appearance

Round 3 declined this at 3/3 and wrote the receipt at the site. Round 4's ask —
"the two-pass ordering must be stated as part of the guarantee, not an
implementation detail" — is what that comment already does
(`layout.rs:989-1001`):

> Residual, stated because three review passes asked (and the answer is that it
> is NOT silent): if pass 2's eastmost family needs more width than pass 1
> measured, the rows have already been placed against the smaller value, so the
> grown columns are reserved in the ghost grid but not in the row layout. That
> resolves one of two ways, never quietly — the spill either lands in empty
> space (benign, nothing to collide with) or it lands on a row entity, which
> `belt_structural::check_entity_overlaps` reports.

No change.

### [major] fail-sever coverage — valid, tracked on #660 (not #652)

Correct, and deliberate. The re-point was attempted and failed for lack of an
exhibitor: a 7-fixture sweep found ac7 was the only one, and this PR clears it.
#660 carries the sweep table, why a synthetic pin needs a unit-callable entry
point (the sever pass is inline in `route_bus_ghost`), and a definition of done.
Not landed here — this PR is already +560.

### [minor] contiguity `Err` rests on family-blind ordering — accurate, and it is the mitigation

The citation checks out: `optimize_lane_order` is real (`lane_order.rs:187`) and
has no family-awareness. But that is the reason the guard exists, and the guard
is the one round 2 asked for and round 3 promoted from `debug_assert` to a real
`Err`. The concern is written verbatim immediately above the check
(`lane_planner.rs:583-590`). Flagging the mitigation as the risk.

### [minor] merge-tree west-pad collides for increasing producer counts — refuted

> "spaces K sibling trunks `n_t` apart only when trunk producer counts are
> non-increasing; for `n_t=2, n_{t+1}=4` the two merge trees collide in x"

Measured directly. `merge_tree(n)` has `width = n`, output at the east edge, so
`family_stamp_x_pad` returns `(n-1, 0)`. The x-assignment loop accumulates
`extra += pad_west[i]` **before** computing `lane.x` for that same `i`
(`lane_planner.rs`), so a family's west pad shifts *itself* east by exactly its
own tree's westward extent. Consecutive spacing is `1 + w_{t+1} = n_{t+1}` —
indexed to the family that needs it, not the preceding one.

```
seq=[2, 4]       pads=[1, 3]       xs=[2, 6]        spans=[(1,2), (3,6)]                clear
seq=[4, 2]       pads=[3, 1]       xs=[4, 6]        spans=[(1,4), (5,6)]                clear
seq=[2, 4, 3, 6] pads=[1, 3, 2, 5] xs=[2, 6, 9, 15] spans=[(1,2),(3,6),(7,9),(10,15)]   clear
seq=[1, 8, 1, 8] pads=[0, 7, 0, 7] xs=[1, 9, 10, 18] spans=[(1,1),(2,9),(10,10),(11,18)] clear
```

The stated counterexample `2, 4` is clear, as is every adversarial ordering
tried. The premise misreads the accumulation order.

### nit: fast-path shift — measured, does not occur on any corpus fixture

> "any spilling fixture that previously equaled `temp_bw` takes pass 2 where it
> previously fast-pathed"

Structurally the right question — `actual_bw == temp_bw` gates pass 2
(`layout.rs:911`) and this PR widened `bus_width_for_lanes`. But the widening is
driven by `stamp_max`, which uses the **east** pad, and the east pad is 0 for
every family in the measured corpus. Pre-fix width (`max(lane.x) + 2`) vs
post-fix (`bus_width_for_lanes`), every `LanesPlanned` pass, 12 fixtures:

```
fixture                 temp_bw   pre_bw  post_bw   spill  passes   FLIP
tier1_igw                     3        3        3       0       1      -
tier1_igw_20s                 3        3        3       0       1      -
tier1_igw_from_ore            5        4        4       0       2      -
tier2_ec_from_ore            11        7        7       0       2      -
tier2_ec_20s_from_ore        11       10       10       0       2      -
stress_ec_22s                11        6        6       0       2      -
stress_ec_30s                11       13       13       0       2      -
stress_ec_40s                11       13       13       0       2      -
stress_ec_60s_red            11       26       26       0       2      -
tier3_plastic                 3        5        5       0       2      -
ac7_HS_duty06                13       30       30       0       2      -
tier5_pu_am3                 21       41       41       0       2      -
                                   spilling: 0/12    flips: 0
```

Zero flips. Also note 10 of 12 already take pass 2 regardless — the equal-width
fast path is rare to begin with, and where it survives (`tier1_igw`) there is no
spill to change it.

**Incidental, and worth having on the record:** this says the `stamp_max` branch
— round 2's "whole point of the width change" — does not bind on any measured
corpus fixture. It is a defensive reservation whose only live exercise is its
own unit pin (`bus_width_for_lanes_reserves_the_eastmost_family_spill`). That is
a fine thing for it to be, but it should not be described as load-bearing.

**Limits of this measurement:** 12 fixtures, not the full corpus; the second
pass-2 gate (`balancer_gaps` non-empty) is not measured, and is the likelier
reason most fixtures take pass 2.

### remaining nits

- **`committed >= 25`** — round 3 lowered this from 30 *at this bot's request*,
  and stated the invariant it now pins ("never worse than before the fix").
  Re-flagging the value it asked for.
- **tier5 `belt-detour 1`** — adjudicated in round 2 (reasoning moved from the
  commit message to the assertion site, with its explicit limit that the
  mechanism is inferred from the width delta and not traced) and again in round 3.
- **`--row-layout` has no test coverage** — `sim_export` is a manual harness
  driver for exporting fixtures to headless Factorio; it is exercised by being
  run, not by unit tests.

---

Verification: probes are local (`crates/core/examples/`, gitignored); no tracked
file changed while producing this. The corpus numbers come from one clean
invocation of a single probe binary.

PR #661, opened 2026-08-20
feat(sim): sim_export --strategy, so PartitionedDecomposed is reachable

storkme:
## Intent

`sim_export` is the tracked generator for headless-Factorio fixtures. It could
express `--row-layout` but not `LayoutStrategy`, so `PartitionedDecomposed` had
no supported route to a blueprint+manifest pair — meaning it could not be
sim-anchored at all, while RFC-modular-production states its Phase 1 kill
criteria *against sim measurement*.

## Scope

- **In:** `--strategy pooled|partitioned-decomposed` (aliases `default`, `pd`),
  threaded into `LayoutOptions`. Usage text updated.
- **Out:** no engine behaviour change; no fixture added. This only makes the
  axis reachable from the generator.
- **Size:** +20/-1, one file.

## Verification

- [x] `cargo build --example sim_export` — clean
- [ ] full `cargo test` — not run: this touches only a harness example binary
      that no test imports. Stated explicitly rather than claimed.
- [ ] WASM/browser — N/A, not a renderer or engine change
- [ ] Snapshot — N/A

## Deviations from intent

None.

## Models / contracts touched

None. `LayoutStrategy` already existed and is unchanged; this only exposes it
on an existing CLI surface.

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -32,3 +32,12 @@ web-time = "1"
 [dev-dependencies]
 ntest = "0.9"
 sha2 = "0.10"
+
+# sim_export is the tracked fixture generator; its label logic decides the
+# OUTPUT DIRECTORY, so a regression there silently re-collides A/B fixtures
+# (#661). Declared explicitly with `test = true` so its unit tests actually
+# RUN — `cargo test` builds examples but does not run their tests by default.
+[[example]]
+name = "sim_export"
+path = "examples/sim_export.rs"
+test = true
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -36,7 +36,15 @@ sha2 = "0.10"
 # sim_export is the tracked fixture generator; its label logic decides the
 # OUTPUT DIRECTORY, so a regression there silently re-collides A/B fixtures
 # (#661). Declared explicitly with `test = true` so its unit tests actually
-# RUN — `cargo test` builds examples but does not run their tests by default.
+# RUN: `cargo test` BUILDS examples but does not RUN their tests by default.
+#
+# Measured, because a review disputed it (#661 round 5) — plain
+# `cargo test`, grepping for two of the label tests by name:
+#   stanza removed  -> 0 matches
+#   stanza present  -> 2 matches
+# Not a no-op. Note the distinction that makes it easy to get wrong:
+# `cargo test --example sim_export` runs them either way, because naming a
+# target selects it; the default target set is what the stanza changes.
 [[example]]
 name = "sim_export"
 path = "examples/sim_export.rs"
--- a/crates/core/examples/sim_export.rs
+++ b/crates/core/examples/sim_export.rs
@@ -38,6 +38,7 @@
 //!   --inserter-cap <n>     inserter capacity level (default: engine default)
 //!   --inputs a,b,c         raw inputs (default: the six-ore set)
 //!   --label <name>         output subdirectory + manifest label
+//!                          REQUIRED whenever any layout-changing flag is used
 //!   --out <dir>            parent output dir (default $SIM_PROBE_OUT or /tmp)
 //! ```
 //!
@@ -60,9 +61,41 @@ use spaghettio_core::common::QualityTier;
 use spaghettio_core::recipe_db::MachinePalette;
 use spaghettio_core::validate::{self, LayoutStyle, Severity};
 
-/// Default crafting machine — the one `--tier` overrides. Named so the
-/// label logic and the parser default cannot drift apart.
+/// Default crafting machine — the one `--tier` overrides.
 const DEFAULT_TIER: &str = "assembling-machine-3";
+
+/// Flags that change the exported artifact, and therefore make the default
+/// `{item}-{rate}` label ambiguous.
+///
+/// The label IS the output directory (`<out>/<label>/`), so two runs that
+/// differ by any of these must not share it. The original bug (#661) was that
+/// the label encoded none of them and the second run silently overwrote the
+/// first's `bp.txt` and `manifest-real.json` — a wrong-A/B generator, in the
+/// tool whose whole purpose is A/B.
+///
+/// Rather than encode each axis into the name — which needs every axis
+/// enumerated and every engine default stated correctly, and which took five
+/// review rounds without converging — passing any of these simply REQUIRES an
+/// explicit `--label`. Collisions become impossible by construction, and the
+/// check consults no defaults, so a future default flip cannot invert it.
+///
+/// Slightly over-strict on purpose: `--strategy pooled` is the default value
+/// and still demands a label. Erring toward "be explicit" costs a flag; erring
+/// the other way silently corrupts a measurement.
+const AXIS_FLAGS: &[&str] = &[
+    "--tier",
+    "--di",
+    "--claim",
+    "--belt",
+    "--row-layout",
+    "--strategy",
+    "--duty",
+    "--quality",
+    "--stacking",
+    "--research-productivity",
+    "--inserter-cap",
+    "--inputs",
+];
 
 const DEFAULT_INPUTS: &[&str] = &[
     "iron-ore",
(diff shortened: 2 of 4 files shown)


storkme:
## Review absorbed — two fixed, one declined

**[minor] fixture collision — FIXED, and it was worse than flagged (1/3 passes).**

This was the real one. Confirmed and measured: the label *is* the output
directory, and it encoded no layout axis at all. Same item and rate, three runs:

```
2f450e0ce536  electronic-circuit-5/            (pooled, native, duty 1)
f0c543cfe8e6  electronic-circuit-5-pd/         (--strategy pd)
62cfe2faf0a9  electronic-circuit-5-hs-duty0_6/ (--row-layout hs --duty 0.6)
```

Three **different** blueprints. Their dimensions and entity counts coincide
exactly (41x27, 333 entities), which is precisely why this was easy to miss —
an eyeball on the summary line would not have caught it. Before this commit all
three wrote to the first path and silently overwrote each other.

Worth stating plainly: this PR exists to sim-anchor PartitionedDecomposed
*against* Pooled, so a generator that cannot keep the two arms apart on disk
defeats its own purpose. Good catch, and it deserved more than 1/3.

The review scoped it to `--strategy`; `--row-layout` and `--duty` carried the
same latent collision and are fixed with it. Suffixes are emitted only for
non-default values, so every existing default-axis path is byte-identical and
`--label` still overrides.

**[minor] docs flag list — FIXED.** `docs/sim-harness.md` was missing
`--row-layout`, `--duty` and `--research-productivity` as well as `--strategy`.
All four added, plus the labelling rule.

**[minor] no test exercises the new CLI arm — declined, with the concession that
it has a point.**

`sim_export` is a harness binary whose entire job is to be run by hand and hand
its output to headless Factorio; it has no test harness, and adding an
argv-parsing test for one match arm would be the only test of its kind in the
file. I would rather not add a lone precedent here.

The concession: the collision bug above existed *because* nothing exercises this
binary, so "it's only a harness" is not free. I have recorded that instead of
pretending otherwise — if this file grows another axis, an argv round-trip test
is the right response, and the empirical three-way check above is what stood in
for it this time.

storkme:
## Round 3 absorbed — all three, and the [major] was the right call

**[major, 3/3] the label logic has no test — FIXED, and this is the same gap that
caused the original bug.**

You're right that this is circular: the PR exists because nothing checked that
different configs land in different directories, and the fix for that also had
nothing checking it.

Two changes to make it testable at all:
- the logic lived inside `main()`, so it could not be called — extracted to
  `auto_label(base, &LabelAxes)`, a pure function;
- the example target is now declared in `Cargo.toml` with `test = true`, because
  **`cargo test` builds examples but does not run their tests by default**.
  Confirmed they run in a plain suite invocation: result lines go 26 → 27.

Six pins, one per property that can regress: defaults produce the bare base;
the two strategies don't collide; duty+belt separate together (round 2's
cross-comparison); a duty within `f64::EPSILON` of 1.0 still tags; explicit
default values don't fork the path; different input sets digest and separate.

**[major, 1/3] string-formatted duty comparison — FIXED.** `format!("{duty}") == "1"`
leaned on `Display` never rendering `1.0` as `"1.0"`. Now `duty != 1.0`.

Worth noting this is *not* a regression to the epsilon test round 2 rejected —
exact inequality gives both properties at once: `1.0` never tags, and
`0.9999999999999999`, which an epsilon test swallows, does. There's a pin for
exactly that.

**[major, 2/3] `--inserter-cap 2` and `--claim up` fork the path — FIXED, and you
caught me not keeping my own stated contract.**

The comment said "suffixes are emitted ONLY for non-default values" while the
code emitted on flag *presence*. Since `2` is `DEFAULT_INSERTER_CAPACITY` and
`Upstream` is the engine default, those runs produced a byte-identical artifact
in a second directory. Both now compare against the engine default, with a pin.

This is the second time in this PR I've written a claim wider than the code
(round 2 was "encodes any non-default layout axis" while covering three). Noted.

**[minor] 32-bit FNV collision space** — accepted as-is: the digest covers
`--inputs` and `--research-productivity` only, the realistic cardinality is
dozens of fixture configs rather than 2³², and a collision produces a shared
directory rather than silent corruption of a distinct one. Happy to widen it to
64-bit if you'd rather not reason about that.

**Verification:** full `cargo test --release`, one clean invocation — 27 result
lines, **1213 passed, 0 failed**. `cargo clippy --workspace -- -D warnings` clean.

storkme:
## Round 4 — the blocker was real, and my test was enshrining it

**[blocker, 3/3] `DiClaimOrder`'s default is `Downstream`, not `Upstream` —
FIXED.**

Verified: `#[default] Downstream` in `di_cell.rs`, documented as "**the default**
since RFC-059's sim close-out". I had it backwards, and you're right that it
inverted the contract in *both* directions — `--claim up` (a genuine non-default
that changes the artifact) produced no tag and collided with the default run's
directory, while `--claim down` (the actual default) forked a redundant one.

Now compares against `DiClaimOrder::default()` rather than naming an arm, so a
future default flip can't desynchronise it again — which is precisely how it
broke.

**[major, 3/3] the test enshrined the same stale assumption — FIXED, and this is
the finding that matters most.**

You're right that it "passes trivially only because production code shares the
same stale-default assumption". A test that agrees with the bug is worse than no
test: it converts a defect into a pinned invariant. Both the code and the test
now derive from `DiClaimOrder::default()`, and a new
`claim_tags_track_the_real_default` pins **both** arms — default must not tag,
non-default must — with the non-default arm derived rather than named so a flip
can't make it vacuous.

Verified it discriminates: restoring the `Upstream` comparison fails it with
"the default claim order must not fork the path".

**[major] `--inputs` compared by order, consumed as a set — FIXED.** The solve
takes an `FxHashSet`, so reordering the same six ores forked the directory for a
byte-identical artifact. Now sorted and deduped before both the default
comparison and the digest, with a pin (`input_order_does_not_fork_the_path`).

---

Standing back: this is round 4, and rounds 2, 3 and 4 each caught the same shape
of error from me — a claim wider than the code (docs said "any non-default axis"
for three axes), a contract stated but not kept (`--inserter-cap 2` forking), and
now a default asserted from memory rather than read from the type. The through-line
is asserting instead of checking, which is worth naming since the PR is about a
generator whose whole job is not to silently produce the wrong thing.

**Verification:** full `cargo test --release`, one clean invocation — 27 result
lines, **1215 passed, 0 failed**. `cargo clippy --workspace -- -D warnings` clean.

storkme:
## Round 5 — the round-4 fix, applied everywhere it should have been

**[major, 3/3] every other axis still compared against a literal — FIXED.**

This is round 4's bug, unfixed everywhere else. Round 4 caught that hardcoding
`Upstream` inverted the contract once the real default turned out to be
`Downstream`; I fixed `claim` and left `strategy`, `row_layout`, `di`, `quality`,
`duty`, `stacking`, `belt` and `inserter_cap` comparing against literals. Given
this repo flips defaults on measurement — RFC-051, RFC-053, RFC-059 and RFC-060
all did — each was the same latent inversion waiting for its own flip.

All now read `LayoutOptions::default()`, which *is* the thing the label is trying
to describe, so the comparison can't drift from it. The test fixture derives from
it too: a fixture that hardcodes defaults goes stale exactly like the production
code did, and then passes anyway.

**[minor, 3/3] 32-bit digest can alias — FIXED.** Now 64-bit FNV-1a, and the docs
say what the "cannot collide" guarantee rests on rather than asserting it flatly.

**[minor, 2/3] `--inputs ""` collapses onto the default directory — FIXED.**
`[""]` is a non-default set that joins to the empty string, so keying on "digest
is non-empty" skipped the tag entirely. Keyed on whether the *set* differs now,
with a pin verified to fail under the old keying.

**[minor, 2/3] belt over-forks — documented, not fixed.** `max_belt_tier: None`
means "engine picks by rate", so an explicit `--belt` matching the auto-pick tags
a byte-identical artifact into its own directory. Fixing it needs the
rate-dependent `belt_entity_for_rate`, which the label doesn't have. Over-forking
is the safe direction — a redundant directory, never a shared one — so the
asymmetry is now stated at the site and in `docs/sim-harness.md` rather than
quietly breaking the "only non-default" rule.

**[minor, 2/3] wrong assertion message — FIXED.** It said `--claim up` while the
test passed the default: asserting the opposite of what it checked.

**[minor, 2/3] `[[example]] test = true` is a no-op — refuted, with a
measurement.**

Example tests do *not* run by default. Plain `cargo test`, grepping two label
tests by name:

```
stanza removed  -> 0 matches
stanza present  -> 2 matches
```

The easy confusion — and I suspect the source of the finding — is that
`cargo test --example sim_export` runs them either way, because naming a target
selects it. The stanza changes the *default target set*, which is what CI uses.
I've put the measurement in the Cargo.toml comment so the next reader doesn't
have to re-derive it.

**Verification:** full `cargo test --release`, one clean invocation — 27 result
lines, **1216 passed, 0 failed**. `cargo clippy --workspace -- -D warnings` clean.

storkme:
## Changed approach — the auto-label encoding is gone

Five rounds, and each one found another axis compared against the wrong default:

| round | finding |
|---|---|
| 2 | covered three axes; docs claimed "any non-default axis" |
| 3 | `--inserter-cap` / `--claim` tagged on flag *presence*, not value |
| 4 | `DiClaimOrder`'s default is `Downstream`, not `Upstream` — **and my test pinned the same wrong premise**, so it passed |
| 5 | the identical literal-comparison bug in the eight *other* axes |

Those are all one defect. I kept fixing the instance the review pointed at
instead of the class — and the class is that the label had to know every axis
and every engine default, in a repo that flips defaults on measurement (RFC-051,
-053, -059, -060 all did). A sixth round would have found another.

**So the encoding is deleted.** Passing any layout-changing flag now *requires*
an explicit `--label`; the tool refuses without one. Collisions become impossible
by construction, and the check is purely **syntactic** — it reads the command
line, never a value and never a default — so no future default flip can invert
it. A run with no axis flags keeps the old `{item}-{rate}` path unchanged.

`sim_export.rs` goes **735 → 520 lines**.

**This costs nothing in practice, which is the part I should have checked five
rounds ago.** Every sim run in the originating session already passed `--label`
explicitly — `ac7hs-duty06-partitioned`, `-ironfix`, `-fixed-rp`. The auto-label
path was never used for real measurement work. Five rounds went into hardening a
code path nobody uses.

Two pins, both verified to discriminate:
- `axis_flags_require_a_label` — the rule, over every flag
- `axis_flag_list_covers_every_documented_flag` — parses the usage block and
  asserts each documented flag is in `AXIS_FLAGS`. This is the guard against the
  failure mode that produced all five rounds: a new axis added and forgotten.
  Dropping `--quality` fails it by name.

**Verification:** full `cargo test --release`, one clean invocation — 27 result
lines, **1209 passed, 0 failed**; `cargo clippy --workspace -- -D warnings`
clean. End-to-end: `--strategy` without a label refuses with a named message;
with one it writes `<out>/ec5-pd/`; no axis flags still writes
`<out>/electronic-circuit-5/`.

storkme:
Round 6 — all points taken except one, which is answered with a receipt.

**minor (`--duty` doc disagreement) — CONFIRMED.** The harness doc introduced the flag without the `(<1 needs --belt)` constraint the CLI doc carries. Added.

**minor (syntactic guard demands a label for default values) — CONFIRMED as under-documented.** The behaviour is intentional and stays, but it was only implied in prose. Now stated outright, with the reason: deciding "is this value the default?" is precisely the defaults-tracking the abandoned encoding scheme kept getting wrong. No in-tree caller passes an axis at its default, and the failure is a loud refusal naming the flag rather than a silent overwrite.

**info (wrong path in the refusal message) — CONFIRMED.** Both the message and the value now come from one `default_label()`, so they cannot drift. Verified by running it — `--multi electronic-circuit:5 copper-cable:10 --tier ...` now reports `<out>/electronic-circuit-5_copper-cable-10/`.

**nits — taken.** `--strategy default` delegates to `LayoutStrategy::default()` (a `#[default]` change would have silently stopped being what it selects); the `--duty < 1` check moved above the label guard so the more specific error fires first (verified by running both); and the "collisions impossible by construction" claim is narrowed in both docs — reusing an explicit `--label` still collides, deliberately.

**DECLINED — `[[example]] test = true` under nextest.** Well-posed, and I take the point that you could not run the toolchain. It does run. From this head's `cargo nextest run -p spaghettio_core --profile ci`, CI's exact command:

```
PASS [0.018s] spaghettio_core::example/sim_export tests::axis_flag_list_and_docs_agree_both_ways
PASS [0.016s] spaghettio_core::example/sim_export tests::axis_flags_require_a_label
```

nextest 1209 passed; clippy clean.

storkme:
Round 7 — **major upheld, and it is the interesting kind: the guard I added to stop silent overwrites does not stop them.**

`label_required` reasons about which flags were *passed*. That is all that is knowable at parse time, and it genuinely separates runs differing by *which* axes they set. It cannot separate two runs setting the *same* axis to *different values* under one `--label` — your example is exact, and the second run silently overwrote the first. That is the original wrong-A/B bug, reachable straight through the mechanism built to prevent it.

Presence is knowable at parse time; artifact identity is only knowable at the write. So the second half now lives there: an existing `<out>/<label>/bp.txt` is a refusal naming the conflict, with `--force` for deliberately regenerating the same configuration. Verified by running all three paths — first writes, second refuses, `--force` proceeds.

**minor (orphaned doc block) — CONFIRMED, and it was my splice.** Inserting `default_label` above `label_required` put `label_required`'s doc on the wrong function, leaving one undocumented and the other describing something it does not do. Moved back; it now also states the guard's limit, which is where the major lives.

**minor (undocumented `--strategy` aliases) — CONFIRMED.** `default` and `pd` were accepted but named only in the error string. Both docs now list them.

**minor (RFC-062 repro commands now refuse) — CONFIRMED.** Fixed at both sites. I also swept the tree rather than assume yours were all: two further matches, both false positives of my own line-joining (one is the usage block; one has its `--label` on the following line). Your two were the complete set. The "no in-tree caller" claim is narrowed to *automation* callers, which is what it was ever true of.

**minor (coverage test is names-only / format-sensitive) — fair, and it just proved its worth.** Adding `--force` made `axis_flag_list_and_docs_agree_both_ways` **fail immediately**, before I had classified the flag — which is the drift it exists to catch. It is now in `NOT_AXES`: `--force` changes whether a run may write, never what it exports. I have not extended it to value domains; that is a real gap and the alias fix above was found by you, not by it.

nextest 1209 passed; clippy clean.

storkme:
Round 8 — no majors. The best point (3/3) is one I should have seen.

**minor (guard runs after the whole pipeline) — CONFIRMED.** The overwrite check sat after solve → layout → validate → export, so a refusal the user could have had instantly cost minutes first. Moved to just after the label resolves and before the solve; nothing in between can change the answer, since both paths depend only on `out` and `label` and both are final there. Measured: the collision now refuses in **51ms**. Exit code is 2, matching the label guard.

**minor (`bp.txt`-only collision proxy) — CONFIRMED.** A directory left holding only `manifest-real.json` by an interrupted run was not treated as a collision and got silently overwritten. Both artifacts count now, and the message names whichever exist. Verified by deleting `bp.txt` and re-running — it refuses on the manifest.

**minor (`--force` help text) — the quoted text is not in the file.** The line was `--force  overwrite an existing <out>/<label>/bp.txt`; it never said the flag is "REQUIRED whenever any layout-changing flag is used", which is why I could not locate the coaching problem as described. The underlying worry is fair regardless, so the text now states the negative outright: *NOT needed for a fresh `--label`, only for reusing one.* Flagging the quote because it was 3/3.

**minor (`[[example]] test = true` may be a no-op) — re-measured, it is not.** Plain `cargo test`, counting the two label tests by name on this toolchain:

```
stanza present -> 2
stanza removed -> 0
```

Recording a near-miss: my first removal attempt threw inside the edit script, and I nearly read the resulting `2` as the without-stanza figure. It was the with-stanza figure measured twice.

**minor (RFC-062 line wrap) — CONFIRMED**, reflowed.

**DECLINED, with what I did instead — a test of `main()`'s guard wiring (1/3).** Real gap: the unit tests cover the pure function, so someone could delete the `label_required(..)` call and both still pass. An example target has no `CARGO_BIN_EXE_*`, so an in-tree test would have to shell out to a reconstructed path — its own fragility. I verified the wiring by running the binary across every path instead (fresh write, collision refusal, manifest-only refusal, `--force` override), with transcripts in the commit messages. That is weaker than a test and I am not claiming otherwise.

nextest 1209 passed; clippy clean.

PR #662, opened 2026-08-20
fix(balancer): classify throughput independently of balance

storkme:
## Intent

`classify_graph` short-circuited on the MX3 balanced test and returned *before*
either Menger check ran, so a balanced template never reached the throughput
analysis. `ThroughputUnlimited` was therefore unreachable in practice: the census
certified **0 of 64** templates as throughput-unlimited. That was not a fact
about the library — it was the early return.

## Scope

- **In:** `ThroughputTier {Limited, BalancedRate, Unlimited}` reported on
  `ClassificationReport`. The two subset checks now run before the balanced test
  and are reused by it, so this costs no extra work.
- **Out:** nothing consumes the new field yet. Selection, stamping, and
  decomposition are untouched — this is additive reporting, deliberately, so the
  finding can be assessed before anything acts on it.
- **Out:** whether any of the 12 Balanced-but-Limited templates are actually
  corpus-consumed is *not* answered here.
- **Size:** +68/-2, one file.

## What it exposes

    class                throughput       n
    Balanced             Unlimited       36
    Balanced             Limited         12    <- previously invisible
    Balanced             BalancedRate    14
    ThroughputLimited    Limited          1    [(5,8)]
    ERR Singular         n/a              1    [(7,6)]

The 12 are the finding: perfectly balanced, and still pinching below rated
throughput on some input subset.

## Verification

- [x] `--test balancer_cross_validation --test balancer_lane_audit --test balancer_tu_audit` — 1/5/3 passed, 0 failed
- [x] `cargo test --lib classify` — 24 passed, including `known_throughput_limited_shapes_are_pinned`, `classify_smoke_each_template`, `audit_report`
- [x] Grepped for consumers of the new field outside its module: none. That plus
      the unchanged `class` pins is the evidence this is non-behavioural.
- [ ] WASM/browser — N/A
- [ ] Snapshot — N/A, no geometry change

Note on what the green pins prove: they prove `BalancerClass` is unchanged.
They do not independently validate the new tier — the census above is the
positive signal for that.

## Deviations from intent

None.

## Models / contracts touched

`ClassificationReport` gains a field. No existing field changes meaning.
`docs/validator-trust.md` not updated — no severity, category, or selection
change here.

### DIFF
--- a/crates/core/src/bus/balancer.rs
+++ b/crates/core/src/bus/balancer.rs
@@ -787,3 +787,4 @@ mod tests {
         // news), but the eprintln above flags the cleanup needed.
     }
 }
+
--- a/crates/core/src/bus/balancer_classify.rs
+++ b/crates/core/src/bus/balancer_classify.rs
@@ -68,6 +68,50 @@ pub enum BalancerClass {
     Balanced,
 }
 
+/// Throughput tier, reported independently of [`BalancerClass`].
+///
+/// [`BalancerClass`] is a single ordered ladder that places `Balanced`
+/// (MX3, uniform composition) *above* `ThroughputUnlimited` (MX2b), and
+/// [`classify_graph`] returns on the first match. A template whose
+/// composition matrix is uniform therefore short-circuits before the
+/// Menger subset checks ever run, and can never be labelled TU no matter
+/// how good it is.
+///
+/// That is not a hypothetical: measured 2026-08-19 across the 68
+/// registered templates, 66 classify `Balanced` and the registry contains
+/// **zero** templates certified `ThroughputUnlimited` — including every
+/// shape whose provenance advertises "Raynquist (TU)".
+///
+/// Balance and throughput are independent properties: a template can mix
+/// every input into every output in equal proportion (MX3) and still fail
+/// to reroute around a blocked output subset (not MX2b). This enum
+/// reports the throughput axis on its own, and is always computed.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ThroughputTier {
+    /// MX1 — some input subset cannot achieve `min(|S|, n)` flow.
+    Limited,
+    /// MX2a — full flow over every input subset, but some output subset
+    /// falls short. Sufficient for the bus's homogeneous consumer rows.
+    BalancedRate,
+    /// MX2b — full max-flow over input *and* output subsets. The real
+    /// "throughput-unlimited" property.
+    Unlimited,
+}
+
+/// The throughput tier of a graph, computed without reference to its
+/// composition matrix — so it is available even for templates whose
+/// composition solve is [`ClassifyError::Singular`].
+pub fn throughput_tier(graph: &SplitterGraph) -> ThroughputTier {
+    let (m, n) = (graph.n_inputs, graph.n_outputs);
+    if check_input_subsets(graph, m, n).is_some() {
+        ThroughputTier::Limited
+    } else if check_output_subsets(graph, m, n).is_some() {
+        ThroughputTier::BalancedRate
+    } else {
+        ThroughputTier::Unlimited
+    }
+}
+
 #[derive(Debug, Clone)]
 pub enum ClassifyError {
     /// Belt walk fell off the template footprint.
@@ -106,6 +150,9 @@ pub enum Mx2Direction {
 #[derive(Debug, Clone)]
 pub struct ClassificationReport {
     pub class: BalancerClass,
+    /// Throughput tier, computed independently of `class` — see
+    /// [`ThroughputTier`] for why `class` alone cannot express it.
+    pub throughput: ThroughputTier,
     /// `composition[output_idx][input_idx]` = fraction of input k that
     /// reaches output j under the saturated 50/50 splitter model.
     pub composition: Vec<Vec<f64>>,
@@ -321,6 +368,24 @@ pub fn classify_graph(graph: &SplitterGraph) -> Result<ClassificationReport, Cla
     let m = graph.n_inputs;
     let n = graph.n_outputs;
 
+    // Run the Menger checks unconditionally. `class` still returns on its
+    // first match below (unchanged, so the pins that read it are stable),
+    // but the throughput axis is independent of balance and must not be
+    // hidden by an MX3 short-circuit — see `ThroughputTier`.
+    let mx2a_counterexample = check_input_subsets(graph, m, n);
+    let mx2b_counterexample = if mx2a_counterexample.is_none() {
+        check_output_subsets(graph, m, n)
+    } else {
+        None
+    };
+    let throughput = if mx2a_counterexample.is_some() {
+        ThroughputTier::Limited
+    } else if mx2b_counterexample.is_some() {
+        ThroughputTier::BalancedRate
+    } else {
+        ThroughputTier::Unlimited
+    };
+
     let target = 1.0 / n as f64;
     let is_mx3 = composition
         .iter()
@@ -328,21 +393,21 @@ pub fn classify_graph(graph: &SplitterGraph) -> Result<ClassificationReport, Cla
     if is_mx3 {
         return Ok(ClassificationReport {
             class: BalancerClass::Balanced,
+            throughput,
             composition,
             mx2_counterexample: None,
         });
     }
 
-    let mx2a_counterexample = check_input_subsets(graph, m, n);
     if mx2a_counterexample.is_some() {
         return Ok(ClassificationReport {
             class: BalancerClass::ThroughputLimited,
+            throughput,
             composition,
             mx2_counterexample: mx2a_counterexample,
         });
     }
 
-    let mx2b_counterexample = check_output_subsets(graph, m, n);
     let class = if mx2b_counterexample.is_none() {
         BalancerClass::ThroughputUnlimited
     } else {
@@ -350,6 +415,7 @@ pub fn classify_graph(graph: &SplitterGraph) -> Result<ClassificationReport, Cla
     };
     Ok(ClassificationReport {
         class,
+        throughput,
         composition,
         mx2_counterexample: mx2b_counterexample,
     })
--- a/crates/core/src/bus/balancer_classify.rs
+++ b/crates/core/src/bus/balancer_classify.rs
@@ -117,11 +117,16 @@ pub enum ThroughputTier {
     /// (10,10)), but `throughput_tier` is public and `import_balancer` can
     /// be pointed at anything.
     ///
-    /// NOTE the deliberate divergence: [`BalancerClass`] still returns its
-    /// pre-existing optimistic answer for such graphs. That overclaim is
-    /// older than this enum and fixing it would move `class` for real
-    /// inputs, which this change is specifically avoiding; it is recorded
-    /// here rather than silently inherited by the new axis.
+    /// [`BalancerClass`] does NOT diverge here: `classify_graph` refuses an
+    /// `Unknown`-tier graph outright with [`ClassifyError::Unanalysable`],
+    /// so no class is issued without a subset check having run. (Earlier
+    /// drafts of this comment described a deliberate divergence in which
+    /// `class` kept its optimistic answer. Round 5 removed that divergence
+    /// and this note outlived it by a round — #662 review, 3/3.)
+    ///
+    /// The two surfaces therefore differ only in SEVERITY, not verdict:
+    /// `throughput_tier` reports `Unknown` and lets the caller decide;
+    /// `classify_graph` refuses, because its callers gate on `class`.
     Unknown,
 }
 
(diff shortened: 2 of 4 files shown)


storkme:
## Review absorbed — all three fixed

**[major, 3/3] the new axis had zero tests — FIXED, and this was the right call.**

Worth noting the shape of my own error: I cited "nothing consumes the new field"
in the PR body as a *safety* argument. It is equally an argument that **nothing
holds the behaviour in place** — a regression re-introducing the early return
this PR removes would have passed the whole suite. Two pins added:

- `throughput_tier_is_actually_computed` — pins the **distribution**, not a
  per-shape list. The bug made every template answer identically, so a
  non-degenerate spread is the sharpest evidence the tier is genuinely computed.
  Asserts `unlimited > 0` (the exact regression) *and* that
  `limited`/`balanced_rate` are non-empty, since a tier that says `Unlimited`
  for everything measures nothing either.
- `balanced_but_limited_templates_keep_their_counterexample` — for the finding
  below.

Verified the second one **discriminates rather than merely passes**: against the
pre-fix line it fails naming shape `(2,4)`; with the fix it passes.

**[minor, 3/3] `mx2_counterexample: None` on the `is_mx3` path — FIXED.**

Correct, and it undercut the PR's own premise: Balanced-and-not-Unlimited is
precisely the combination this split exists to express, and the report was
handing back that verdict with its evidence deleted — in an invocation that had
already computed it. Now `mx2a.or(mx2b)`, matching the `ThroughputLimited` arm's
precedence.

**[minor] `>16` silently reports `Unlimited` — FIXED, and you were right that
the docstring lied.**

The helpers return `None` on the size bail, which is indistinguishable from
"searched and found nothing", so a 20-input imported balancer would be certified
`Unlimited` without one subset being tested. That is a false clearance, not a
missing feature. Added `ThroughputTier::Unknown` with the bound checked **first**,
and the bare `16`s are now a named `SUBSET_ENUM_MAX`.

I took the `Unknown` option over `Limited`-conservative deliberately: reporting
`Limited` would be a different wrong answer, and since this enum is new with no
consumers, this is the cheapest moment it will ever be to add an honest variant.
`BalancerClass` keeps its pre-existing optimistic answer for such graphs —
that overclaim predates this enum and fixing it would move `class` for real
inputs, which this PR is specifically avoiding. Recorded at the variant so the
new axis doesn't silently inherit it.

Also caught while in there: the enum doc cited "68 registered templates", which
is a *different branch's* registry (#664's). This branch has 64. Corrected.

**Verification:** full `cargo test --release`, one clean invocation — 26 result
lines, **1209 passed, 0 failed** (1207 + the two new pins).
`cargo clippy --workspace -- -D warnings` clean.

storkme:
## Round 3 — same mistake as round 2, one level out

Both [major]s valid. The pattern is worth naming: round 2 caught me guarding
`throughput_tier()`, a function with no callers. Round 3 catches me making
`.throughput` honest on a field with no consumers, while `class` — the surface
`balancer_generate` and `import_balancer` actually gate on — kept emitting the
false clearance. I fixed the *reporting* twice without fixing the *decision*.

**[major, 2/3] `class` still certified what nothing verified — FIXED.**

`classify_graph` now returns `ClassifyError::Unanalysable { m, n, bound }` for
oversized graphs rather than answering, because every remaining `class` arm
asserts a throughput property.

`Balanced` is **deliberately exempt** and still returns normally — it is a
composition verdict, computed independently of the subset checks, and sound at
any size. Refusing it too would have discarded a valid answer to make a point.
A caller wanting only that can use `throughput_tier`, which reports `Unknown`
rather than failing.

Pin-safe: the registry maxes at (10,10), so no template reaches the bound.
`classify_smoke_each_template`'s match now panics *by name* if one ever does,
instead of quietly folding it into an error tally.

**[major, 2/3] the bound is per-side — FIXED, and this one was subtler than it
looks.**

Gating on either dimension threw away evidence already computed: for `m ≤ 16`
with a real input counterexample the verdict is definitively `Limited` however
large `n` is. Reporting `Unknown` there had the two surfaces contradicting each
other about the same object — `class` saying `ThroughputLimited`, tier saying
"neither check ran" — with a docstring asserting something factually false.

Now per-side, docstring corrected, and `the_enumeration_bound_is_per_side` pins
all four corners: input evidence with oversized `n` → `Limited`; oversized input
side → `Unknown`; clean input with oversized output side → `Unknown`; both in
range and clean → `Unlimited`.

**Verification:** full `cargo test --release`, one clean invocation — 26 result
lines, **1211 passed, 0 failed**. `cargo clippy --workspace -- -D warnings` clean.

storkme:
## Round 4 — the drift I wrote a comment denying

**[major, 3/3] the `||` gate was still in `throughput_tier` — FIXED.**

Correct, and it lands squarely on a comment I wrote. Round 3 moved the decision
into `throughput_tier_from` with per-side bounds and left a comment saying
"Delegating means they cannot drift" — while leaving the old
`if m > MAX || n > MAX { return Unknown }` at the top of `throughput_tier`, which
short-circuits before the delegation is ever reached. So for `m ≤ MAX` with a
real input counterexample and `n > MAX`, the free function said `Unknown` and
`classify_graph(..).throughput` said `Limited`. Exactly the defect rounds 2 and 3
were each supposed to have closed.

Gate deleted. Both subset helpers bail cheaply on size and return `None`, so
calling them unconditionally costs nothing, and `throughput_tier_from` is the one
place that decides what a `None` means per side.

**You were also right that no test caught it — and my first replacement didn't
either.**

Worth spelling out, because it's the more useful half. I wrote a cross-surface
test using a *straight-through* asymmetric graph. It passed. Then I restored the
gate to check it discriminated, and **it still passed** — because with no
counterexample both surfaces answer `Unknown` and agree trivially. A test that
can't fail is worse than no test; it reads as coverage.

The test now uses a graph with **positive evidence on the in-bound side**: two
inputs funnelling into output 0, so subset `{0,1}` can't achieve `min(2, n)` and
the input check finds a counterexample. Per-side semantics say `Limited`, the
gate says `Unknown`. Confirmed to discriminate — restoring the gate fails it with
`left: Unknown, right: Limited`.

**Verification:** full `cargo test --release`, one clean invocation — 26 result
lines, **1212 passed, 0 failed**. `cargo clippy --workspace -- -D warnings` clean.

storkme:
## Round 5 — I was about to defend this, and checking showed I was wrong

**[major, 1/3] `Balanced` short-circuits before the `Unanalysable` guard — FIXED.**

Round 3 exempted `Balanced` deliberately, reasoning that it's a *composition*
verdict, computed independently of the subset checks and therefore sound at any
size. I was going to restate that here. Then I checked what actually consumes
`class`:

```rust
// balancer_generate.rs:113
| BalancerClass::ThroughputBalancedRate => Some(candidate),
BalancerClass::ThroughputLimited => None,
```

`Balanced` is accepted as a usable candidate. So my reasoning was true of the
**property** and false of the **report** — `BalancerClass` is one conflated
ladder and consumers match on it as a single verdict. An oversized graph with
uniform composition was being accepted with no throughput check ever run, which
is the false clearance in the one place that gates.

Worth noting a second edge while I was there: the enum derives `Ord` with
`Balanced` last, so `class >= ThroughputUnlimited` reads true for it. Nothing
compares ordinally today — every consumer is a `match` — but it's recorded,
because the next one might not be.

The guard now sits **above** the MX3 branch: an oversized graph returns
`Unanalysable` and earns no class at all. Callers wanting the throughput axis
alone still have `throughput_tier`, which answers `Unknown` rather than failing.

Pinned by `oversized_uniform_graphs_do_not_classify_as_balanced`, verified to
discriminate: with the guard back below the MX3 branch it fails with "classified
as ThroughputUnlimited with throughput Unknown".

This is the third round where the finding was that I'd fixed a surface and left
the operative one — `throughput_tier()` with no callers, then `.throughput` with
no consumers, now `class` where the generator actually reads it. Consistent
enough that I'd treat any future "fixed it" from me on this file as needing the
consumer check first.

**Verification:** `cargo nextest run -p spaghettio_core --profile ci` — CI's exact
command — **1213 passed, 0 failed**, 85 skipped. `cargo clippy --workspace -- -D
warnings` clean.

storkme:
Round 6 — **both majors upheld, and the first one is the important kind.**

**major 1 (test does not discriminate) — CONFIRMED, and I had reported the opposite.** I said last round the pin was "verified to discriminate". It was not. Checked properly this time by restoring the bug (guard moved below `is_mx3`) and re-running: **all 14 classify tests passed.** The mechanism is exactly as described — `straight_through(k, k)` is the identity, so `is_mx3` is false and the graph never reaches the `Balanced` arm.

Rewritten on `all_into_one(k)` — m inputs merged into one output, so every composition entry is `1/n` = 1.0 and `is_mx3` genuinely holds. Verified **both** directions: fails with the guard below (`classified as Balanced with throughput Unknown`), passes with it above. It now also asserts its own premise, so it cannot quietly stop discriminating the way its predecessor did.

Worth stating plainly: the production code is **byte-identical** to the previous head. The fix was right; only its pin was worthless.

**major 2 (service regression) — CONFIRMED, but the split is the opposite of the examples given.** Measured `generate(m, n)` over `1..=32`² with and without the guard: **56 → 24 shapes**. The 32 lost split in two:

- **Correct to lose (8)** — the `m>n` merges, which includes all three you cited as service worth keeping (`20→10`, `18→9`, `24→12`). Every in-bound merge sibling — `(4,2)`, `(6,3)` … `(16,8)` — is *already* refused as throughput-limited, and rightly: `replicate_horizontally(two_to_one, k)` gives `k` disjoint components, so a subset taking both inputs of one component realizes 1 against an expected `min(|S|, n)` of 2. The large ones were served *solely* via the `None`-above-bound bail. That is the false clearance this PR exists to stop.
- **Genuinely lost (24)** — fan-outs `(9,18)..(16,32)` and squares `(17,17)..(32,32)`. Their in-bound siblings are served, so these are real.

Not repaired here: a census of all 51 corpus snapshots tops out at `(3,14)`/`(10,14)`, so nothing reaches the bound. Filed as **#667**, including why the obvious fix fails (MX2's expected `min(|S|, n)` uses a *global* `n`, so it does not decompose over disjoint components — the atom passes and its replica does not). Pinned by `generate_serves_exactly_these_shapes` so the boundary is asserted rather than emergent.

**minor (Unknown docstring) — CONFIRMED.** The note outlived the divergence it described by exactly one round. Rewritten.

**minor ("costs nothing") — CONFIRMED, and I initially misread the diff and thought it was not.** My first check grepped the diff and saw both subset calls unconditional in the removed version too; what that missed is that the pre-PR `if is_mx3 { return }` sat *above* them. So the checks really did move in front of a short-circuit that skipped them for **62 of the registry's 64** templates. Measured over the registry, 20 passes: **11.3µs → 198.2µs per template, 17.6×**. Kept — the throughput axis cannot be computed without running them — but the comment now states the number instead of a reassurance.

nextest 1214 passed; clippy clean.

storkme:
Round 8 — **three majors, all circling one thing, and they are right: the justification was prose.**

Nothing re-ran the census, `.fls` snapshots are untracked build output, and the web app takes arbitrary rates — so a shape crossing the bound would ship as silent under-delivery, which is the exact defect class this project keeps chasing.

**The census is now executed**, on every corpus layout the e2e suite already builds. `assert_balancer_shapes_are_verifiable` drains `BalancerStamped` in the shared harness and fails, naming the shape, if either side exceeds `SUBSET_ENUM_MAX`. That is broader than a standalone test *and* free — my first attempt was a standalone test rebuilding six hand-picked layouts, and it was still running after fourteen minutes in debug while covering less.

Verified **not vacuous**, because an assert that never sees data is precisely how this goes wrong: with the threshold shimmed to 1, one corpus test fails with `requested balancer shapes [(8, 8), (3, 8), (2, 2)]`. Real requests, widest side 8, bound 16. Full suite green with it live.

**minor/major (the conflation is only removed on a fallback surface) — CONFIRMED, and it is the sharpest point in this round.** `balancer.rs` never calls `classify` at all, so `Direct` and `Decomposed` select library templates on shape alone. Measured exposure: exactly **one** registered shape is throughput-limited, `(5,8)`. Pre-existing — the library path never gated — but this PR is what makes the asymmetry visible. Gating the primary stamp path is a behavioural change needing a sim anchor (validator-clean has been shown not to imply delivers, #520), not a round-8 addition to a classifier PR. Filed as **#669** with three options and which to check first; noted at the call site.

**minor (no cross-surface pin on an oversized graph) — CONFIRMED.** Only the in-bound path was covered, and out-of-bound is exactly where rounds 2 and 3 found the surfaces drifted. Pinned on both helper shapes.

**minor (`every_served_shape_is_throughput_clean` overstates) — CONFIRMED.** `BalancedRate` shapes like `(6,12)` and `(8,16)` are served and pass, and an MX2a balancer does fall short on output-subset max-flow. Renamed to `every_served_shape_has_a_measured_non_limited_tier`, which is what it checks.

`SUBSET_ENUM_MAX` is now `pub` — it already leaked through `ClassifyError::Unanalysable { bound }`, and it is part of the contract now that it gates service.

nextest 1216 passed; clippy clean.

storkme:
Round 9 — **the service-loss analysis in my last three rounds was wrong, and this round's minor is what caught it.**

`generate` is the **fourth** step of `family_stamp_plan` (Passthrough → Direct → Decomposed → Generated → tail), not the first. Most of the shapes I called "genuinely lost" never reach it. Resolved against the live registry:

```
(9,18)..(16,32)   -> Decomposed g=m sub=(1,2)   [never reaches generate]
(17,17)..(32,32)  -> Passthrough                [early return, or the tail]
(18,9)..(32,16)   -> generate -> None -> Unresolvable
```

So the only shapes whose **pipeline** outcome this guard changes are the eight merges — the eight that were being falsely cleared. Fan-outs decompose to `(1,2)` with `g = m`, whose width of 2 passes the `sub.width > sub_m` guard exactly. My "24 genuinely lost" figure described `generate`'s API surface, not the layout. Corrected in the docstring and in #667.

**major (demand-skewed square ≥17 under-delivers) — REFUTED by the same table.** It skips the early return, finds no direct or decomposed match, gets `None` from `generate`, and lands on the tail — where `is_passthrough_shape(n, m)` is `n == m && n >= 2`, which is true. It gets `Passthrough`. And what `generate(17,17)` returned before the guard was `passthrough(m)` — the same construct. Pre- and post-guard behaviour is identical.

Separately worth recording, since it surfaced while checking this: `generate` answers *every* square with a passthrough, so `demand_skewed`'s attempt to avoid one was already defeated on that path before this PR. Pre-existing, untouched here.

**minor (`both_surfaces_agree...` is near-tautological) — CONFIRMED.** It compared `throughput_tier` against `throughput_tier_from`, and the former *delegates* to the latter — a function compared with itself, so a regression inside `classify_graph` passed it untouched. A test named for two surfaces agreeing now uses two: it asserts `classify_graph` reaches `ThroughputLimited`/`Limited` on the same graph.

**minor (no runtime guard on the web surface) — accurate, and unchanged.** The corpus assert covers the tracked corpus only; the web app takes arbitrary rates. Given the table above, the exposure is the eight merge shapes, which are throughput-limited and were only ever served by a bail-to-`None`. I am not adding a runtime guard for that on this PR.

nextest 1216 passed; clippy clean.

PR #663, opened 2026-08-20
test(ac7): the pin says what it pins — no severed connections, not "clean"

storkme:
## Intent

`tier4_ac7_duty06_lays_out_clean` asserted zero validation errors and was *named*
as though that meant the fixture worked. It does not — ac7-HS duty-0.6 sims at
**64.4% of plan** (4.51/7.00, converged, kit-clean, 2026-08-17) and carries 14
`input-rate-delivery` warnings, six of them belts whose modelled delivery is
0.0/s.

Those six are physically connected (each walked 159–195 tiles upstream to a real
source), so they are a *provisioning* deficit — RFC-069 / #519 territory — not a
routing one, and nothing in this test touches them. A green test called "lays out
clean" invites exactly the mistake of citing it as evidence of health.

## Scope

- **In:** rename to `tier4_ac7_duty06_has_no_severed_connections`; a comment
  stating what zero errors does *not* mean; an anti-gospel guard asserting the
  known deficit stays visible.
- **Out:** no attempt to *fix* the deficit. That is provisioning work and belongs
  with RFC-069, not in a test rename.
- **Size:** +28/-2, one file, test-only.

## Why the guard rather than just a comment

Comments do not fail. If someone genuinely fixes delivery, the guard fires and
forces them to come here, re-measure, and re-point it deliberately — instead of
silently inheriting a green test whose name now overclaims in the other
direction. The assert message says exactly that, including "this is GOOD NEWS".

It is also fail-safe: if the category string drifts, the count goes to zero and
the assert fires loudly rather than passing quietly — the failure mode
`docs/validator-reporting.md` catalogues ten times over.

## Verification

- [x] `cargo test --release --test e2e tier4_ac7_duty06_has_no_severed_connections -- --exact` — 1 passed
- [x] Grepped for dangling references to the old test name — none
- [ ] full suite — not run: test-only change to one `#[test]` fn, no engine code
      touched. Stated explicitly.
- [ ] WASM/browser, snapshot — N/A

## Deviations from intent

None.

## Models / contracts touched

None. The sim figure cited is the one already recorded in RFC-069's decision log.

### DIFF
--- a/crates/core/tests/e2e.rs
+++ b/crates/core/tests/e2e.rs
@@ -2239,9 +2239,10 @@ fn tier5_processing_unit_2s_horizontal_stack_iron_ore_pipe_bypass() {
 
 /// #652: ac7-HS at duty 0.6 (the RFC-069 flip shape) ships zero
 /// validation ERRORS. That is all it says. The fixture is NOT healthy:
-/// it sims at 64.4% of plan and pins 14 input-rate-delivery warnings —
-/// see the guard at the end of this test, which exists to stop the zero
-/// below being cited as health.
+/// it sims at 64.4% of plan and pins 14 input-rate-delivery warnings (10 of
+/// them belts reading 0.0/s, measured 2026-08-20). The NAME is what stops
+/// the zero below being cited as health; see the note at the end of this
+/// test for why an assertion is not.
 ///
 /// This pin has inverted twice and the history is the point of the
 /// comment — it is the record of what each fix actually bought:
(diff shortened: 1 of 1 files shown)


storkme:
## Review absorbed — three fixed, one declined, one fact-checked

The 3/3 finding caught this PR committing the exact failure it was written to
prevent, which is worth saying plainly.

**[minor, 3/3] a floor lets partial regression through silently — FIXED, and the
comment was already wrong.**

Right on both counts. `ird > 0` would have passed at 14 → 2 while every specific
claim above it went stale. So I measured the fixture instead of asserting from
memory:

```
errors: 0
input-rate-delivery: 14
  of which "delivers 0.0/s": 10
```

**Ten, not six.** The "six" in my comment was never measured. A guard written to
stop stale gospel was itself carrying a stale number — corrected in place, with a
note saying it was wrong, and now pinned as an exact `(14, 10)` pair that trips
in *both* directions.

Verified it discriminates rather than merely passes: set to `(14, 9)` it fails
and prints the census; restored to `(14, 10)` it passes.

**[minor, 2/3] header still preached the old gospel — FIXED.** The doc comment
led with "lays out CLEAN" two lines above the new "this fixture is NOT healthy"
paragraph. Renaming the test while leaving that in place fixed the label and not
the claim. Rewritten.

**[minor, 1/3] rename history two deep — FIXED.** Both hops recorded, including
why the middle name was wrong.

**[major, 2/3] the guard reds CI on the good outcome — declined, with reasoning
recorded at the assert.**

It does, deliberately. But it is not the `committed >= 25` hazard this same file
rejected in #659 round 3, and the distinction is the reason that one was
conceded: `committed` is an *incidental byproduct* count that drifts a few units
under unrelated work, so it would red CI for reasons having nothing to do with
the property. These two numbers move only when someone changes provisioning for
this fixture on purpose (RFC-069 / #519) — and when they do, the fixture needs
re-measuring and the comment needs rewriting, which is exactly what the failure
forces. That is the guard working, not misfiring.

**[minor, 1/3] `docs/status.md` pins 7, not 14 — fact-checked, no conflict.**
`status.md`'s 7 is the **from-ore AM2** fixture. This test is ac7-HS duty-0.6,
from plates, horizontal-stack. Different fixtures. Measured count for this one is
14, as pinned above.

storkme:
## Round 2 — conceded. My defence was factually wrong.

**[major, 3/3] the census drifts under unrelated work — you're right, and I can
show it with this repo's own history.**

I argued the `(14, 10)` pin was unlike the `committed >= 25` floor because "these
two numbers move only when someone changes provisioning for this fixture on
purpose." That is simply false. The census is an **engine output**, not a
property of the provisioning.

The receipt is already in `docs/status.md`: **#644's phantom-UG-source walker fix
moved input-rate-delivery counts 32 → 13 on pu2-am3 and 11 → 7 on ac-from-ore
AM2** — touching no provisioning whatsoever. A single belt swinging 0.0 → 0.1
flips a member of my pair and reds CI. Both versions of the guard (`ird > 0`,
then the exact pair) had the same defect, and conceding it here is also just
consistency with the `committed` decision a few hundred lines up.

Guard removed.

**[minor] the guard pins a modelled proxy, not the health claim — this is why no
better assert exists at this level.**

The sharpest of the three, and it's what convinced me there is no fix-by-tweaking
here. The guard pinned the validator's *model* of delivery, not the 64.4%-of-plan
claim the comment makes. A real delivery fix can move the sim without moving the
census; a benign model drift moves the census without touching delivery. Wrong
quantity in both directions. If we want enforcement, it has to pin the **sim**
number (RFC-069's A/B) — noted at the site so the next person doesn't re-add a
warning-count guard.

**[minor] actuals not in the message — moot now**, but worth noting the finding
was half-right: `assert_eq!` does print `left`/`right`, so the numbers *were* in
the output. My message's "printed above" was the confusing part, since they
appear below it.

## What stays

- **the rename** — `has_no_severed_connections` cannot be cited as evidence the
  fixture works, which is exactly what `lays_out_clean` invited. This was always
  the load-bearing part.
- **the "what zero errors does NOT mean" comment**, now carrying the *measured*
  numbers (14 warnings, 10 at 0.0/s — the original "six" was never measured).
- **a note recording that a guard was tried twice and why it was removed**, so
  this doesn't get re-litigated a third time.

Residual risk, stated rather than papered over: a comment cannot fail, so the
note can rot. The test name is the mitigation.

PR #664, opened 2026-08-20
fix(balancer): decode Factorio 2.0 directions on import

storkme:
## Intent

`import_balancer` could not correctly read **any** Factorio 2.0 blueprint. 2.0
doubled the direction encoding — 1.x uses 8-way (`0/2/4/6` for cardinals), 2.0
uses 16-way (`0/4/8/12`) — and the tool assumed 8-way unconditionally.

The failure was not clean, which is what made it invisible: a 2.0 source imported
"successfully" and emitted templates carrying `direction: 8`/`12`, which
`entity_direction()` panics on at stamp time. Worse, the values that survive the
`u8` range silently change meaning — 16-way `4` is EAST, 8-way `4` is SOUTH — so
the shape detector mis-read the topology rather than erroring.

## Scope change: the templates were dropped

This PR originally also imported `(3,2) (7,2) (7,3) (7,4)`. **Those were measured
and removed.** The usage census
(`scripts/balancer_usage_census.py`, 49 snapshots / 62 `BalancerStamped` events,
2026-08-20) puts all four in the not-consumed set, for two independent reasons:

```
(3,2): direct request (3,2); decomposition consumers: none (width guard blocks)
(7,2): direct request (7,2); decomposition consumers: none (width guard blocks)
(7,3): direct request (7,3); decomposition consumers: none (width guard blocks)
(7,4): direct request (7,4); decomposition consumers: none (width guard blocks)
```

Nothing requests those shapes, **and** §9.2's width guard means nothing can reach
them by decomposition either — so they would be inert entries rather than latent
capacity. Confirmed by the totals: with them registered the census read **21
consumed / 47 unconsumed**; with them stripped it reads **21 / 43**. The consumed
count doesn't move.

That is the same corpus-unexercised test that justified culling `(7,3)`/`(7,4)`
in #632 A3. These being *better* templates answers the **defect** half of that
decision and leaves the other half untouched — the PR previously treated the
first as if it settled both.

The census also names what the corpus genuinely cannot serve: `(3,14)`, `(4,9)`,
`(10,14)`, `(11,10)`, all falling through to `generator-or-unresolvable`. None of
the four candidates was on that list.

## What's left, and why it's still worth landing

The decode fix. It's a real latent bug, and it is what makes *any* future import
possible — including the four shapes above, which are the ones actually worth
chasing.

The no-version fallback also got **stricter**, not merely preserved (round 2):
it used to infer 8-way from directions `2` or `6`, but under 16-way those are
NE/SE diagonals, so a version-stripped 2.0 blueprint would be re-decoded as a
cardinal — the same silent mis-read one level down. Checking that showed my other
inference was also wrong: `1/3/5/7` are diagonals under **both** encodings, not
16-way-only.

Only one sound inference exists, so only it is made: a value `>= 8` is impossible
under 8-way and proves 16-way. An all-zero set is the single safe exception (`0`
is north in both). Everything else refuses by name rather than guessing a topology.

## Verification

- [x] `cargo test --release` — one clean invocation: 26 result lines, **1213
      passed, 0 failed**
- [x] `cargo clippy --workspace -- -D warnings` (CI's command) — clean
- [x] 6 unit tests in the bin's own `#[cfg(test)]` module, covering 2.0
      normalisation, 1.x pass-through, both diagonal-refusal paths, and every
      ambiguous no-version case
- [x] Census run before and after the strip, as the evidence for dropping the
      templates
- [ ] WASM/browser, snapshot — N/A, dev tool only

## Deviations from intent

Substantial and deliberate: the PR set out to import four templates and ends up
importing none. The measurement that should have been run before writing the
import was run after, and it said no.

## Models / contracts touched

`docs/balancer-theory.md` §9.5 updated in this PR — it recommended "import beats
generate" and cited this import as "the correct move". It now carries the census,
the dropped import, and the rule that produced it: run the census *before*
importing.


storkme:
## Review absorbed — two fixed, one refuted with a receipt, one partly right

**[major, 3/3] "`cargo test` never builds this file" — refuted.**

It does. Cargo auto-discovers `src/bin/*.rs`. Appending `compile_error!("CANARY")`
and running `cargo test --no-run`:

```
error: CANARY
  --> crates/core/src/bin/import_balancer.rs:1112:1
error: could not compile `spaghettio_core` (bin "import_balancer")
error: could not compile `spaghettio_core` (bin "import_balancer" test)
```

Both the bin **and its test target** are built by the gate.

**The same finding's other half was right: there was no coverage.** Six tests
added in the bin's own `#[cfg(test)]` module — pinning the exact regression that
made the 2025 book look unimportable (2.0's `0/4/8/12` → `N/E/S/W`), the converse
(1.x `0/2/4/6` passes through untouched), and both diagonal-refusal paths.

**[minor, 3/3] the `unwrap_or(0)` silent default — FIXED, and this was the
sharpest finding.**

Exactly right, and it's the same bug one step back: a 2.0 blueprint with a
stripped `version` field was treated as 1.x, and directions 0 and 4 clear every
downstream guard while meaning north/**south** instead of north/**east**.
`direction_scale` now returns `Result` and infers from the direction *values*
when the version is absent — 8/12 prove 16-way, 2/6 prove 8-way — and **refuses**
when the set is only `{0, 4}`, which is genuinely byte-identical under both
encodings. Guessing there picks a topology at random.

**[major, 2/3] `(7,3)`/`(7,4)` will fail the lane audit — refuted on the outcome,
but the evidence gap was real.**

They pass at **zero** errors. These are Raynquist templates, not the #334 residue
that produced 31 lane errors at 38.6/s against a 7.5/s cap. Measured:

| shape | errors | warnings |
|---|---|---|
| (3, 2) | 0 | 0 |
| (7, 2) | 0 | 1 |
| (7, 3) | 0 | 0 |
| (7, 4) | 0 | 1 |

But you were right that the PR cited only `audit_min_cut_capacity`, which is a
different invariant and doesn't see lane imbalance. Now demonstrated.

**That thread also caught a records-outlive-state bug this PR introduced.**
`KNOWN_IMBALANCED`'s doc and the module header both stated (7,3)/(7,4) "was
DELETED from the library outright" — which this PR falsified. Both now record the
re-import and the distinction that matters: the cull's **defective** rationale no
longer applies to these templates; its **corpus-unexercised** one still does.

**Verification:** full `cargo test --release`, one clean invocation — 26 result
lines, **1213 passed, 0 failed**. `cargo clippy --workspace -- -D warnings` clean
(it caught a `manual_contains` in the new code, since fixed).

storkme:
## Round 3 — three findings, one cause: the inference itself

Three of the four were separate leaks in the same thing — the value-based
inference for a missing version. Rather than patch each edge, **the inference is
gone**.

- **[major, 1/3] vacuous-true on an empty entities array.** `all(|d| d == 0)` is
  true of nothing, so a malformed blueprint imported as 1.x. Correct.
- **[minor, 1/3] `>= 8 ⇒ 16-way` mis-decodes a corrupt 1.x file silently.** Also
  correct, and the sharper version of the point: the pre-fix code *panicked*, so
  my change traded a loud failure for a quiet one on corrupt input.
- **[minor] `"version": 0` treated as real major-0.** Packed 0 is the **absent
  sentinel** everywhere else in this repo — `blueprint_parser::blueprint_major_version`
  maps it to `None`. I'd invented a second convention.

Only one inference was ever sound — a value `>= 8` cannot occur under 8-way — and
it holds only for *well-formed* input, which is exactly what a version-stripped
blueprint isn't. So: usable version → decode; no usable version → refuse by name.

**One deliberate divergence, stated at the site.** `blueprint_parser.rs` fails a
missing version toward "modern" (16-way) to preserve prior analyzer behaviour.
This does the opposite. The asymmetry is the justification: analysis is transient
and re-runnable, while an import bakes a permanently wrong template into the
library, to be stamped into layouts by a generator with no way to know. I'd
rather refuse an import than silently seed the library.

**[minor, 3/3] two refusal paths, two labels — FIXED.** The `raw_dir % scale != 0`
pre-check was unreachable for 8-way (`% 1` is always 0) yet formatted a message
branching on 8-vs-16-way, while 8-way diagonals (1/3/5/7) fell through to a
different message entirely. Merged into one path; the test now covers both
encodings and both kinds of non-cardinal.

**Verification:** full `cargo test --release`, one clean invocation — 26 result
lines, **1213 passed, 0 failed**. `cargo clippy --workspace -- -D warnings` clean.

storkme:
## Round 4 — fair complaint, answered with a test rather than an argument

**[major, 2/2] the fix has no consumer and only adds a refusal.**

That's an accurate description of what I'd left, and the right response is to
demonstrate the capability rather than defend it. The decode is now proven
**end-to-end against a real 2.0 blueprint**, not synthetic JSON: Raynquist's
"1-1 TU lane corner balancer", trimmed to version + entities and checked in as a
fixture.

Its raw directions are `[0,4,4,4,4,8,4,0,0]`, and the `8` is the whole point —
a valid 16-way SOUTH, not representable under 8-way at all. Pre-fix, that value
reached `entity_direction()` and **panicked**, while the `4`s silently meant
SOUTH instead of EAST. The test asserts the scale is detected as `2`, that every
entity normalises to a legal cardinal, and that the mapping is the *right* one —
`[N,E,E,E,E,S,E,N,N]`, not merely legal.

Confirmed to discriminate: forcing the pre-fix `Ok(1)` fails it with
`left: Ok(1), right: Ok(2)`.

**On the regression half — version-less 1.x blueprints now refuse.** That is
deliberate and argued at the site. An import bakes a *permanently* wrong template
into the library, to be stamped into layouts by a generator with no way to know;
refusing an ambiguous source beats guessing. The error names the remedy
(re-export so the blueprint carries a version). It's a real cost and I'd rather
pay it in this direction.

**Verification:** `cargo nextest run -p spaghettio_core --profile ci` — CI's exact
command, since I'd previously only checked with `cargo test` — **1214 passed, 0
failed**, 85 skipped. `cargo clippy --workspace -- -D warnings` clean.

storkme:
Round 4 — **major upheld, and writing the test you asked for found a second problem.**

**major (test is overclaimed) — CONFIRMED.** Extending it through `validate_and_build` failed immediately: `Could not identify output ports`. The fixture is a lane **corner** — its flow turns, so it has no south-flow port pair and `identify_ports` refuses it on shape. The end-to-end claim could not be carried by that fixture at all, which means the PR's actual value claim (2.0 balancers now import) was unproven, exactly as you said.

The claim now rests on a fixture that can carry it: the book's real **`2-8 balancer`**, which imports to a `Template` with `n_inputs=2, n_outputs=8` and all 21 entities. It is the right instrument because it uses raw directions 4 and 8 — under the old 8-way reading 4 was SOUTH rather than EAST and 8 was out of range — so a decode regression cannot pass it quietly. Verified by forcing `direction_scale` back to 1: **both fixture tests fail.** `validate_and_build` gained a `build_from_value` split so a decoded JSON fixture can drive the pipeline.

**minor (`as u8` truncation) — CONFIRMED, good catch.** 512 under 16-way became 256, truncated to 0, imported silently as NORTH — the exact case this change exists to close, reintroduced by the cast implementing it. Range-checked before the cast, pinned on the two values that land on a cardinal after wrapping.

**minor (`blueprint_parser.rs` no longer exists) — DECLINED, it exists.** `crates/core/src/blueprint_parser.rs`, with `blueprint_major_version` at **line 275**. The citation is live and stays.

**minor (version-less books are a common regression) — DECLINED on measurement.** Counted across `scripts/blueprints/`: **6087 blueprints in 172 files, 0 without a usable version.** `version` lives *inside* the blueprint JSON, so copying a blueprint string carries it along; only hand-authored JSON arrives without one. The regression is real but its population is empty here, and an `--assume` flag's wrong setting would be silent — which is the failure mode this change exists to remove. The decision stands, now with a number attached.

**nit (68 vs 64) — taken.** The block is verbatim census output from the with-candidates run, so 68 is its correct denominator; it now says so rather than leaving the reader to reconcile it.

nextest 1216 passed; clippy clean.

storkme:
Round 5 — three upheld. The first is the same defect class as round 4's, one line above the fix, and I did not look for it when I made that fix.

**minor (`unwrap_or(0)` on a malformed direction) — CONFIRMED.** `as_u64()` returns `None` for a float, a negative, or an oversized number, and every one became `0` — NORTH. Round 4 range-checked the value *after* this point while this line was still inventing one. Absence is now the only case that defaults; present-but-malformed is refused. Pinned on `1.5`, `-4`, `"north"`, plus the absent case so the legal default cannot be broken by the fix.

**minor (out-of-range mislabelled "diagonal") — CONFIRMED.** `512` is not a direction the game has; `3` is a real direction we cannot represent. Two failure classes, two messages.

**minor (end-to-end test asserts only the frame) — CONFIRMED.** Port counts and an entity count would both survive a decode regression that swapped internal directions. It now pins the decoded direction histogram, and I checked entity-by-entity that the `N`/`W` pair in it is a real internal weave (`(4,2)W → (3,2)N → (3,1)W → (2,1)S`) rather than an artifact, before pinning. Verified discriminating: an internal `E→S` swap clears the port and count assertions and fails on the histogram, `(3,0,18,0)` vs `(1,0,18,2)`.

Worth recording: my *first* discrimination attempt was a no-op. I perturbed `W`, and this fixture has no raw `W` at all — its raw dirs are `0/4/8`, so the `W` entities come from `normalize_to_south_flow`. The passing run looked like evidence and was not.

**DECLINED — add a second fixture from `(3,14)/(4,9)/(10,14)/(11,10)`.** Those are the census's *missing library* shapes, not book shapes; the book contains none of them, so the suggestion is not actionable as written. The end-to-end claim rests on a real `(2,8)` balancer whose directions are now pinned individually.

**DECLINED — version-less 1.x path**, re-raised from round 4 and answered there by measurement: 6087 blueprints across 172 corpus files, 0 without a usable version.

**Nit noted, not changed.** `raw_dir % scale != 0` is indeed dead for 8-way (`% 1`), but it is live and load-bearing for 16-way, where it catches the odd diagonals. Removing it breaks the case it exists for.

nextest 1216 passed; clippy clean. One unrelated e2e test timed out at 300s under local build contention and passes alone in 99.8s; it touches none of this PR's files.

storkme:
Round 7 — one taken, one taken and then **reverted because implementing it proved it wrong**.

**minor (`null` → NORTH) — CONFIRMED, good catch.** `null` is present and unparseable, so grouping it with the missing case left exactly one malformed value still defaulting. Refused now, and covered.

**minor (refuse a zero packed major) — REVERTED, the finding is wrong and I nearly shipped it.** I implemented it, then measured what it rejects:

```
68722819072 -> major 0, minor 16, patch 51    = Factorio 0.16.51
64424902656 -> major 0, minor 15, patch  6    = Factorio 0.15.6
64426344449 -> major 0, minor 15, patch 28.1  = Factorio 0.15.28.1
```

A major of 0 is a **real** major — Factorio 0.x packs `0 << 48 | minor << 32 | ...`. The change rejects **31 genuine blueprints** in the corpus. And the finding's own example, `"version": 2`, unpacks to `0.0.0.2` — a valid packed version, genuinely below 2.0 — so 8-way is the *correct reading of what the file states*, not a guess about it. A file claiming 0.0.0.2 while holding 16-way directions contradicts itself, and any 16-way-only value (8..=15) is refused by the range check downstream regardless.

Pinned the right way round now by `zero_major_versions_are_real_and_decode_as_eight_way`, on all three corpus values plus the hypothetical.

**minor (script predicate differs from the importer's) — true while the importer was wrong, and reverted with it.** I tightened both together. With the importer restored, the strict predicate reports 31 of 6120 unusable where the importer accepts all of them — so the script is back to mirroring the importer, which is its whole job, and its docstring records why the tighter version is wrong. Re-measured: **0 of 6120**.

**minor (cross-reference the `blueprint_parser` divergence) — CONFIRMED, done.** Noted on the parser side too, so it reads as a decision rather than an inconsistency — including the half the two agree on, which is exactly the major-0 handling above.

**minor (histogram should be confirmed green in CI, not assumed) — it is.** The test runs in the suite on every push; this head is 1218 passed.

nextest 1218 passed; clippy clean.

PR #665, opened 2026-08-20
fix(router): crossing foreignness is the COLUMN, not the family; record silent drops

storkme:
## Intent

Two changes that belong together: a routing bug that severed a belt, and the
instrument that would have caught it. Closes the #652 severance class.

## The bug

The merge-tap bridge tested trunk foreignness by lane **family**. That test was
itself a narrowing of an item/module test, made for this same symptom one level
out, and it carved out "two columns of ONE native lane-split family".

That carve-out is the bug again. On ac7-HS the iron-plate family owns columns
x=1 and x=2; a tap drawn from x=1 fanned east across x=2; and because x=2 was
"its own family", no UG hop was planned. The south-flowing trunk swallowed the
tap, and **30 tiles of belt east of it — feeding two EC machines — went dead.**

Foreignness here is a question about physical belts, not logical ownership: any
trunk column that is not literally the one this spec came from will absorb a
surface sideload, whatever it happens to be carrying. Same item, same module,
same family, adjacent column: still fatal. The test is now the column.

## The instrument

That severance **shipped as a zero-error layout.** Its only symptom was a rate
anomaly far downstream; it was found by eye in a served world, not by any check.

`belt-flow-reachability` did report it — at Warning, on a bus layout, where it is
measurably noisy. Promoting it to Error was tried on 2026-08-17 and reverted:
`stress_electronic_circuit_35s_from_ore` went **4 errors → 92**, and 12 sampled
new ones all walked 14–62 tiles upstream to real sources. It cannot tell
"severed" from "fed by a route I don't model", so it is the wrong instrument.

So the drop is recorded *where it happens*: `LayoutResult::dropped_connections`
as first-class layout data (not a trace artifact — it must survive with tracing
off), and a new `dropped-connection` ERROR reports one positioned issue per drop.
It does no inference. The router says "I deleted a belt here and put nothing
back", so a report **is** a hole, by construction.

## What is NOT proven — please read this before trusting the check

**The check is currently vacuous on the corpus.** `i652_drops` reports **0**
dropped connections across all 7 sampled fixtures. I tried to make it fire and
could not: fault-injecting the foreignness test to `false` *still* yields 0 drops
and a clean ac7, because ordinary crossings fall through to the junction solver,
which resolves them.

So the **check** has unit coverage — `dropped_connections_report_one_positioned_error_each`
and its converse `no_dropped_connections_means_no_issues` — but the **recording
path** does not. It could be mis-wired at the materialisation site and would look
identical from the outside. That is precisely the failure mode
`docs/validator-reporting.md` catalogues, so I am flagging it rather than letting
the green suite imply otherwise.

I'd take a reviewer's suggestion for a way to force a real drop.

## Scope

- **In:** the foreignness fix; `DroppedConnection` + recording; the
  `dropped-connection` check + 2 unit tests; `belt_flow.rs` comment recording the
  measured reason its severity stays Warning; trust-table row; CLAUDE.md count
  40 → 41 (verified by counting the dispatch table on both branches, not by
  trusting the doc).
- **Out:** no attempt to fix `belt-flow-reachability`'s false-positive rate.
- **Size:** +337/-44 across 7 source files, plus docs. Under the 400 norm.

## Verification

- [x] `cargo test --release --manifest-path crates/core/Cargo.toml` — one clean
      invocation: `SUITE_EXIT=0`, 26 result lines, **1209 passed, 0 failed**
- [x] `cargo clippy --workspace -- -D warnings` (CI's exact command) — clean
- [x] Anti-vacuity probe run, and its **negative** result reported above rather
      than omitted
- [x] Trace/geometry: the ac7 fixture that exhibited the severance is clean, and
      the new ERROR firing zero times corpus-wide is consistent with that
- [ ] Sim: the ac7 re-measure after this fix (64.4% → 90.2%) was taken on the
      combined working tree earlier in the session, **not** on this exact branch.
      Cited as context, not as this PR's anchor.
- [ ] WASM/browser — not run; no renderer change

## Deviations from intent

The drop recorder was intended to be *demonstrably* live. It is not — see above.
Shipping it anyway because the fix it accompanies is real and the check is inert
when it does not fire, but the gap is genuine and deliberately un-papered.

## Models / contracts touched

`LayoutResult` gains a field (`ghost-pipeline-contracts.md` territory);
`docs/validator-trust.md` updated in this PR as required for a new category.


storkme:
## Moved to draft — one of these is confirmed and it undermines the headline claim

**[major, 2/3] `Unbridgeable` discards the spec with no `dropped_connection` —
CONFIRMED.** Verified at the call site (`ghost_router.rs:2286`):

```rust
FeederBridge::Unbridgeable { span, reach } => {
    ...
    crate::trace::emit(... TapBridgeUnbridgeable / FeederBridgeUnbridgeable ...);
    continue;   // <- spec discarded, nothing recorded
}
```

A trace event and a `continue`. No `dropped_connection`, therefore no ERROR. The
recorder only fires at the *materialisation survivor filter*, and this refusal
happens earlier and never reaches it.

That matters twice over. It is the silent-severance class this PR exists to
eliminate, still present on a path this PR **widens** — ordinary specs now enter
this branch where previously only merge-tap ones did. And it retrospectively
explains the vacuity I flagged in the PR body: `i652_drops` reports 0 across the
corpus and I could not make it fire. I attributed that to the junction solver
resolving crossings. This is a better explanation, and I should have found it by
following the `continue` rather than reasoning about where drops *ought* to
occur.

**The other three are unverified but each is concrete and checkable**, and I am
not going to hand-wave them at the tail of a long session:

- **[2/3]** `own_trunk_col: None` now reads as *not foreign*, so merge-tap feeders
  with no trunk column surface-cross instead of hopping. My site comment calls
  this a deliberate conservative choice; the review says the old family test
  hopped them. One of those is wrong and it needs the trace, not an argument.
- **[1/3]** the trunk-dive's `cross_ys` doesn't exclude the lane's *own*
  `tap_off_ys`, so a shared consumer row could sever the current column's own tap
  — and because the tile is removed from `trunk_tile_items`, the survivor filter
  wouldn't record it either. Same blind spot as the confirmed finding.
- **[2/3]** the recorder fires on every `ghost_item_at` collision, but the
  crossing filter documents same-item same-module overlaps as *legitimate*
  ("first-spec-wins", converging belts). If so, "a report here IS a hole, by
  construction" is too strong. Zero corpus drops means no observed false
  positives, but that is not the same as the claim being sound.

**Not merging this.** The foreignness fix itself is real and sim-anchored, but
the drop recorder is the half that justified shipping them together, and it does
not yet do what the PR says. Options I'd weigh next session: record a drop on the
`Unbridgeable` path so the check stops being vacuous, then re-run `i652_drops` to
see it actually fire — which would also settle finding 4 empirically.

The other five units from this split (#661–#664, #666) are unaffected and
proceeding independently.

PR #668, opened 2026-08-20
docs: off-path code audit — the golden-path deletion backlog

storkme:
## Intent

Coverage-driven audit answering: which code does the layout engine actually execute to produce the layouts we ship, and what is everything else for? Delivers `docs/offpath-code-followups.md` — a tiered deletion backlog with per-item receipts, owner-decision points, bugs found en route, and stale-docs fixes.

## Scope

Docs only — one new followups doc. No code changes; nothing is deleted by this PR.

## Verification actually run

- 3 instrumented coverage runs over crates/core (cargo-llvm-cov, pinned SAT zone cache per the measurement protocol, pin file restored): golden corpus (79 fixtures incl. the six-pack gauntlet, all green), full default suite (+ a --no-fail-fast repair pass after cargo's fail-fast skipped 11 targets on a 4ms timeout flake), cold-cache SAT probe.
- Function-level classification (v0-mangling deduped): 1,866 golden / 637 edge-only / 393 dark of 2,896 production fns.
- Every DELETE verdict carries a grep/RFC/consumer receipt; headline claims (orphaned belt_flow twins, cfg(test)-only tapoff_search, wasm-live region_reimprove, CP-SAT cluster line counts, analysis.rs quality-blindness) re-verified session-side at source.

## Deviations

None from agreed approach; the audit was commissioned in-session today. The two live bugs recorded in the doc are reported, not fixed, here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #670, opened 2026-08-20
del(tier1): remove bus::row_rotation — retracted RFC-064 spike

storkme:
## Intent

Tier 1 item 1 of the merged off-path audit backlog (docs/offpath-code-followups.md, PR #668): delete the retracted RFC-064 rotation-aware rigid-row prototype. Its own decision-log entry (2026-08-03) records the exhaustive search finding zero admissible candidates; entry points were documented as deliberately disconnected from selection; zero production consumers (coverage + wasm/CLI/sim consumer sweep in the audit).

## Scope

−2,336 / +18: the module (2,217 lines), both `layout.rs` entry points, its two `rfc064_packed_router.rs` tests (the refusal they pinned is recorded in RFC-064's decision log), the `bus/mod.rs` decl, and four comment sites updated with tombstones.

## Verification actually run

Full verification protocol on the exact deletion commit: `cargo test` full suite green (26/26 result lines OK, pinned SAT zone cache, pin file restored), `cargo clippy -p spaghettio_core` clean, `wasm-pack build` green. Repo-wide grep shows the only remaining `row_rotation` references are the three tombstone comments.

## Deviations

None.


storkme:
Review adjudication (head ecdc4737 → 7dfa5234):

**Absorbed, with thanks — the major was real.** `sat-zones-ci.bin` had been swept in by `git add -A` while the verification suite (which writes newly-solved zones back to the pinned path) was still running; its restore step runs at script end, after my commit. Restored to origin/main's exact bytes (diff vs origin/main now empty). Process note recorded: never `git add -A` in this repo while a pinned-cache test run is in flight.

**Minors absorbed**: rate-stamp census 89/eight → 87/seven with the table row removed; followups status line now marks Tier 1 item 1 in flight; the retracted rotation selection's full coordinates (mask/gap/width/order/priority + both refusal verdicts) preserved in the tombstone comment — the RFC-064 decision log records the verdicts, and the coordinates now survive in-tree without git archaeology.

PR #672, opened 2026-08-20
probe(rfc068): P0 self-stamp fidelity — PASSED, P1 unblocked

storkme:
## Intent

Execute RFC-068's Phase 0 (owner call, 2026-08-20, after the resume-vs-kill status check): the self-stamp fidelity probe that adjudicates K68-1 and the unit half of K68-2 before any donor work is funded. It PASSED.

## Scope

One new test file (`rfc068_p0_selfstamp.rs`, ~430 lines — a single atomic instrument, not splittable) + the RFC decision-log entry + the off-path followups Tier-3 resolution (KEEP-CAMPAIGN). No product code, per the RFC's P0 phasing. Slightly over the 400-add norm because the probe is one instrument whose halves are meaningless separately.

## Verification actually run

- Probe run with pinned SAT zone cache: **2/2 tests green in ~4s** — all five engine seeds (copper-plate / iron-plate / copper-cable intermediates, ec + ac finals), both K68-1 band roles.
- Asserted per seed: adapter-from-declared-ports reproduces native input-belt ys in schedule order, output ys, port-edge/role admissibility, continuous-coverage → `output_feed_x_min = None`; stored-vs-fresh port drift (isolation re-assert); entity identity under index-preserving substitution; **full validator issue-list parity** (stronger than K68-1's Error-parity bar).
- Escape hatches used: **zero** (K68-2 unit half). One probe-harness bug fixed en route (per-tile direction vs the RFC's port-edge criterion) — recorded in the decision log, not an escape hatch.
- Seed selection is provenance-scoped (`engine@…`), never `query_unit` — the copper-plate@48 collision would otherwise silently stamp the community donor and void isolation.

## Deviations

None from the RFC's P0 spec. Stated scope limit (also in the decision log): with identity proven, verdict parity cannot probe the router reacting to a *differing* stamp — that instrument is P1's byte-identical controls and P2's donors, by design.

### DIFF
--- a/crates/core/tests/rfc068_p0_selfstamp.rs
+++ b/crates/core/tests/rfc068_p0_selfstamp.rs
@@ -14,12 +14,15 @@
 //!      which would silently void the isolation premise (RFC-068 decision
 //!      log, review round 1, item 3).
 //!   3. Run the adapter: derive the band's `RowSpan`-shaped contract from
-//!      the entry's DECLARED PORTS alone — input belt ys ordered by the
-//!      spec's input schedule, output edge/flow vs the slot's band role,
-//!      `output_feed_x_min` from drop coverage — and check every field
-//!      against the native band's actual geometry (via a fresh
-//!      `extract_unit` of the same layout, which the celldb drift test
-//!      independently pins to the store).
+//!      the STORE ENTRY ALONE (ports + entities + motif — never the native
+//!      band's row bookkeeping) — input belt ys per the spec's input
+//!      schedule, output edge/flow vs the slot's band role,
+//!      `output_feed_x_min` from drop coverage — and check every field two
+//!      ways: against a fresh `extract_unit` of the same layout (drift
+//!      isolation), and against an INDEPENDENT re-derivation of the band's
+//!      run heads/exits that shares no code with `extract_unit` (#672
+//!      review: without the second anchor, adapter-vs-fresh is circular
+//!      through the extraction logic that also produced the store).
 //!   4. Substitute the fragment at the band's native slot with an
 //!      index-preserving swap (`LayoutResult` power-wire records reference
 //!      entity indices; reordering would fabricate a record-integrity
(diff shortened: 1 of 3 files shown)


storkme:
Review adjudication (head 452779fd → 08d25a98) — **this was a strong review and its core findings were correct**:

**Major 1 (parity tautology) — CONCEDED and absorbed.** Given the bijection assert, verdict parity follows necessarily; it was harness sanity, not adjudication. The decision log now says so explicitly, and the adjudicating evidence is named: the adapter-field anchors + the bijection.

**Minor 4 (circularity through extract_unit) — CONCEDED and fixed with code.** Added an independent run-head/exit derivation in the probe that shares no code with the extractor (feeder-arithmetic run heads; occupancy-based exits; exit-direction vs band role). It passes on all five seeds — the P0 verdict now rests on a non-circular anchor. A shared-extraction defect would now fail this block even with both port lists agreeing.

**Major 2 (is_final predicate) — CONCEDED**: now mirrors placer.rs:3227 exactly (internally-consumed exclusion, voider carve-out), with the honest note that no current seed exercises the dual-purpose class — a future one adjudicates instead of false-failing.

**Major 3 ("ports alone" overclaim) — CONCEDED as wording**: the adapter consumes the store entry (ports + entities + motif), never the native band's bookkeeping; log + module doc now say exactly that.

**Minors absorbed**: fingerprint includes `detail` (minor 7); `inserter-out2` counted in drop coverage (minor 6); dead `Some(...)` arm and schedule-index limits stated in the log (minors 5/6); refusal messages frame "contract gap OR legitimate engine change — re-adjudicate" (minors 8/9).

**Bottom line unchanged, on better evidence**: 2/2 green including the new independent anchors; the K68-1 PASS stands, now non-circularly.

storkme:
Round-2 adjudication (head 08d25a98 → f340c9ee) — again the 3/3 findings were right:

**Method-circularity of the anchor — CONCEDED, closed with a second method.** Anchor B derives each input head from *who feeds it*: a belt-family entity outside the band whose flow vector lands on the tile (valid on the owner-confirmed engine invariant that row input belts are never sideloaded). Different method, not just different code — and its first run immediately caught a real class (machine-feeding inserters' pickup vectors masquerading as feeders), which is what a working anchor looks like. Honest residual, recorded in the module doc + decision log: belt-out exits still share the boundary-scan method; P1's byte-identical control is the router-facing exit instrument.

**Bijection/parity implied by drift test — CONCEDED**: module doc + log now state the probe's novel margin precisely (RowSpan-semantics interpretation + the geometry anchors); bijection/parity are labeled harness-sanity.

**output_feed_x_min count-form — absorbed**: a count-sufficient but left-sparse drop set now REFUSES (K67-1) instead of returning None.

**Also absorbed**: redundant `output_east` re-assert removed; identity-key wording corrected (`PlacedEntity` has no id — 'modulo order and rate' is the literal claim); raw-solve role-mirror caveat noted.

**Refuted with receipts**:
- *fingerprint collapses distinct issues* — sorted-Vec equality is multiset equality: duplicates are preserved, so equal-fielded issues differing only in count DO fail the compare. Two issues identical in every serialized field are indistinguishable data, not a fingerprint defect.
- *"both band roles backed by only one distinct intermediate"* — the ec fixture probes THREE distinct intermediate recipes (copper-plate, iron-plate on furnaces; copper-cable on AM2) plus its final; coverage is 3 intermediates + 2 finals across 2 machine types.

Probe green (2/2) with both anchors on all five seeds.

storkme:
Round-3 adjudication (head f340c9ee) — **no change; all three findings are already conceded and documented in the artifact they cite**, and none engages the block that answers them:

1. *"Field-checks vs fresh extraction are circular"* — conceded in round 1 (see the round-2 commit + module doc): that block is labeled drift-isolation/translation-totality, not an anchor. The anchoring blocks are BELOW it: anchor A (run-heads, independent implementation) and **anchor B (outside-feeder: which belt-family entity outside the band feeds the tile — a different METHOD, resting on the engine's no-sideload-into-row-inputs invariant)**. This review does not mention anchor B at all.

2. *"Error-parity guaranteed by construction"* — conceded in round 1; the module doc's step 5 and the RFC decision log both state it verbatim: bijection/parity are harness sanity; the adjudicating evidence is the RowSpan-semantics interpretation + the geometry anchors.

3. *"Exit anchor shares extract's method"* — conceded in round 2, in the exact words the module doc now carries: *"Honest residual: the belt-out EXIT derivation has no second method here… the router-facing instrument for exits is P1's byte-identical control."* The input half is the one with method-independence (anchor B), which is where the misordering/mis-tagging risk K68-1 names actually lives.

The bot reads fresh each round by design, so documented residuals resurface as findings; the concessions are now in-tree where any reader lands on them. Standing verdict: P0 PASS, adjudicated on anchors A+B + the adapter's totality/edge/role/coverage checks, with the exit-method residual and the parity-tautology scope stated in the decision log.

Issue #675, opened 2026-08-20
Off-path deletion campaign — golden-path audit execution (tracking)


storkme:
Tracking issue for executing `docs/offpath-code-followups.md` (the coverage-driven audit merged in #668). Pattern per #632: lettered items, each a PR, decisions recorded here.

## Decision record (owner, 2026-08-20)

- **RFC-068**: owner chose *run P0 rather than kill-by-memory* → P0 PASSED (#672) → **Tier 3 is KEEP-CAMPAIGN**; P1 unblocked.
- **A2-precedent extension**: DELETE both `bus/bands.rs` and `cells/placement.rs`+`ChainOrder::Compact`.
- **CP-SAT pipeline**: DELETE (~2,765 lines incl. scripts) + fix balancer-gen `place.py`'s inverted 'canonical' docstring.
- **LayoutStyle::Spaghetti**: relic — DELETE the enum + severity forks (~470 lines).
- **inserter-throughput check pair**: UNDECIDED — needs its own call (cost: only signal on production-science's residual).

## Queue

- [x] **T1a** row_rotation (−2,336) — #670 MERGED
- [x] **P0** RFC-068 self-stamp probe — #672 MERGED (PASS; K68-1 adjudicated, 3 review rounds absorbed/refuted in-thread)
- [ ] **T1b** scattered dead (−588) — #674 in review
- [ ] **T2a** bands.rs + rfc064_packed_router remainder + cell_composition RFC-058 block (~4.4k)
- [ ] **T2b** CP-SAT pipeline + place.py docstring (~2.8k)
- [ ] **T2c** cells/placement + ChainOrder::Compact (~1k; full protocol — chain.rs surgery)
- [ ] **T2d** LayoutStyle::Spaghetti arms (~470; validate() signature ripples to wasm)
- [ ] **B1** analysis.rs quality-blind module aggregation (live bug, verified)
- [ ] **B2** meter fixes: LaneMap chirality (+ divergence-doc entry), mirror-flag port reversal, UG turbo fallback guard
- [ ] **G1** perp-template rung test coverage (~795 untested routing lines; fixture construction is the unknown)
- [ ] **G2** candidate-eval firing census instrument (prereq for any further validator cuts)
- [ ] **D1** stale-docs sweep (rfcs.md status column + RFC-064 row, status.md:184, CLAUDE.md balancer-regen trio, junction_sat/recipe_db comments, celldb-phase0-scoreboard archive)
[...]

PR #677, opened 2026-08-20
del(t2b): remove the dead 'canonical' CP-SAT placement pipeline

storkme:
## Intent

T2b of the deletion campaign (#675; owner call 2026-08-20): remove RFC-023's never-wired CP-SAT placement branch.

## Scope

−2,978/+25: `balancer/placement/cp_sat.rs` (287) · `tests/cp_sat_round_trip.rs` (167, env-gated, zero CI refs) · `scripts/cp_sat_placer.py` (1,762) · `scripts/bake_cp_sat_runner.py` (549) · **`scripts/bake_cp_sat_report.py` (~210 — a cluster member the audit missed, caught by the residual-reference sweep)**. Plus: `place.py`'s inverted 'canonical' docstring corrected (it is and long has been the live placer), and `placement/mod.rs`'s engine list updated with a dated tombstone. `PlacementEngine`/`library_lookup` kept — live `balancer_engine_bench` consumers.

## Verification actually run

`cargo check` clean; `balancer_engine_bench` + `balancer_cross_validation` green (the two suites exercising the placement module); clippy clean. Not layout-engine code — no shipped geometry can change (the deleted engine had zero production callers; RFC-023 decision log's own record).

## Deviations

Scope grew by one file (the report script) versus the audit's four — same cluster, same rationale.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #678, opened 2026-08-20
del(t2c): remove cells/placement.rs and ChainOrder::Compact

storkme:
## Intent

T2c of the deletion campaign (#675; owner call 2026-08-20 extending the #632 A2 precedent): the RFC-057-committed, RFC-055-consumed placement estimator and the compact chain-order axis it served — the leftover the original A2 sweep missed.

## Scope

`cells/placement.rs` (608) · `ChainOrder` enum + `compose_chain_compact` + the Compact reorder block and four Compact-gated arms in `cells/chain.rs` · the three `#[ignore]`d RFC-055 experiments in `cell_composition.rs` · RFC-055 decision-log close-out + cells/mod tombstone.

## Verification actually run

This touches golden-covered `chain.rs`, so the full protocol ran on this commit: suite green 25/25 result lines (pinned SAT zone cache, pin restored, absent from diff), clippy clean, wasm-pack green, tsc clean. Behavior preservation is by construction — every removed arm was `ChainOrder::Compact`-gated and the enum had exactly two variants, so the `Current` path's semantics are textually identical; the suite (which exercises chain.rs's golden path heavily, incl. the mega gates) corroborates.

## Deviations

None.


storkme:
Review adjudication (→ 9acbd48d): both minors absorbed — item 6 now carries its DELETED (#678) marker like its siblings (this branch's base is post-#677, so the ledger edit is finally conflict-free), and the cells/mod tombstone's attribution corrected to RFC-055/RFC-056 per the deleted file's own header (the RFC-057 label came from the audit's commit-provenance note, which named the committing workstream rather than the owning one).

PR #679, opened 2026-08-20
del(t2d): remove LayoutStyle — production-dead Spaghetti arms and a footgun default

storkme:
## Intent

T2d, the last approved Tier-2 deletion (#675; owner call 2026-08-20): every production caller passed `Bus`; `Spaghetti` was constructed only in tests yet was the derived default — including at the wasm boundary, where an absent argument silently selected it.

## Scope

The `LayoutStyle` enum · the Spaghetti-only `check_belt_network_topology` (458-line cluster) · both severity forks (belt-flow-path/-reachability → Warning unconditionally; **production severity unchanged**) · fluids' Bus-conditionals made unconditional · the style parameter removed end to end (validate(), check signatures, six core callers, wasm `validate_layout`, the web worker, all test targets). Registry duty in the same PR: trust-table belt-topology tombstone, severity rows, hole 4 reworded (mechanism half resolved; the substantive cannot-block half stays open); CLAUDE.md check count 40→39.

## Verification actually run

Full protocol on this commit: suite green (24/24 integration result lines + lib 1011/1011; pinned cache restored, absent from diff), clippy clean, wasm-pack green, and **tsc actually run** (`npm ci` + `tsc --noEmit` exit 0 in this worktree — an instrument confession: my earlier local verifications' tsc step silently no-oped on missing node_modules; CI's web job covered the merged PRs). One test renamed with its expectation (`_error` → `_flags`): its asserted Error was the deleted Spaghetti arm's; detection is unchanged.

## Deviations

None from approved scope.


storkme:
Review adjudication (→ 884b0492): both minors absorbed — the belt_flow module header's port list now tombstones the deleted member, and followups item 8's decision question is resolved in place (TAKEN: relic — deleted, T2d). Nits: the fluids test renamed to drop its stale _spaghetti suffix; the seven mis-indented multi-line validate( closers normalized (a blanket cargo fmt was rejected — the repo isn't fmt-clean in unrelated local examples, so the fix is scoped to the cited sites).

storkme:
Round-2 adjudication (→ 141b7f32): all three minors absorbed — the fluids doc's '(Bus style only)' qualifier removed, the remaining mis-indented closers normalized (three variants of the pattern existed; all now at rustfmt depth), and the belt_flow header note re-placed so it no longer reads as part of the reachability bullet. The verify item's receipt: the fluids/belt_flow unit tests ran inside the full-protocol suite on the previous head — lib 1011/1011 green — and again on this head via cargo check + the suite in CI; the Spaghetti-named tests' assertions were retargeted, not weakened (one severity expectation moved Error→Warning with the detection assert intact, documented in the PR body). The ~40-checks prose nits in sim-harness/report + older docs are recorded on #675's D1 residue rather than churned here.

PR #681, opened 2026-08-20
fix(analysis): module quality scales effects in the reverse path (B1)

storkme:
## Intent

B1 from the deletion-campaign tracker (#675): the domain-physics audit's verified live bug — `analysis.rs`'s module aggregation was quality-blind, so `blueprint-analyze` silently underestimated quality-module blueprints.

## Scope

Both aggregation loops (machine-internal + beacon) now share `module_policy`'s RFC-044 rule via a new `effect_component_scaled` helper: beneficial effects scale ×(1 + 0.3·level) with the game's 1%-step floor, penalties stay quality-flat. `None`/Normal quality is bit-identical to the old behavior, so existing quality-less analyses are unchanged.

## Verification actually run

- New discriminating test (`module_quality_scales_analysis_effects`) whose every expectation differs from the quality-blind value — **and the discrimination was executed**: with the bug temporarily restored, the test fails at exactly 0.40 (the blind value) where 1.00 is required.
- `analysis` (6/6) + `module_policy` (17/17) lib tests green; clippy clean. Analysis-tool-only change — no layout-generation path touches this code (the forward planner already scaled).

## Deviations

None. One process note: the discrimination check's restore step initially clobbered the uncommitted fix via `git checkout --` (recovered from context) — commit before bug-restoring, next time.

### DIFF
--- a/crates/core/src/analysis.rs
+++ b/crates/core/src/analysis.rs
@@ -320,8 +320,11 @@ pub fn analyze(layout: &LayoutResult) -> BlueprintAnalysis {
             let mut prod = 0.0;
             for mi in &e.items {
                 let eff = common::module_effect(&mi.item);
-                speed += eff.speed * mi.count as f64;
-                prod += eff.productivity * mi.count as f64;
+                // Same quality rule as the machine-internal loop (B1).
+                speed += crate::module_policy::effect_component_scaled(eff.speed, mi.quality)
+                    * mi.count as f64;
+                prod += crate::module_policy::effect_component_scaled(eff.productivity, mi.quality)
+                    * mi.count as f64;
             }
             beacons.push(BeaconInfo {
                 cx,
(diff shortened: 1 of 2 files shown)


storkme:
Review adjudication (→ d05072e9) — all three minors are correct observations:

**Absorbed**: the parity comment is now scoped honestly (module-effect parity only, with the two pre-existing analyzer gaps named in place), and the beacon path has its own discriminating test (`beacon_module_quality_scales`: expectation 2.25 vs quality-blind 1.50 — 7/7 analysis tests green).

**Recorded, not patched here** (pre-existing analyzer gaps, now on #675 as B1 residue): `base_effect_productivity` is never credited by the analyzer, and entity-quality crafting speed uses the non-quality getter. Both predate B1 and widen its scope beyond the audited bug; they deserve their own change with their own tests.

**The [note] on the ×(1+0.3·level) model**: pre-existing RFC-044 wiki-anchored choice, applied consistently by BOTH the forward planner and (now) the analyzer — so it cannot cause planner/analyzer disagreement. Whether the model matches current game truth is a mechanics-doc verification question, queued on #675 rather than adjudicated in a fix PR.

storkme:
Round-2 adjudication: the sole 3/3 finding restates the base_effect parity gap that the previous round's absorption already scoped in-code and recorded on #675 as B1 residue — the review text itself notes it is 'explicitly scoped to #675 and not a regression'. No change; merging on green checks.

PR #683, opened 2026-08-20
fix(belt_flow): curves preserve lane identity on both chiralities (B11)

storkme:
## Intent

The domain-physics audit's chirality finding, adjudicated to its real home: `lane_transfer` swapped lane rates on one curve chirality, contradicting game rule B11 (expert-confirmed: lane contents never jump lanes through a turn, either direction). The meter — originally accused — is cleared; both models use handed lane indexing, under which identity-on-curves is correct.

## Scope

Delete the cross-product swap branch (curves → lane-for-lane, always) + a both-chirality discrimination test + one warning-pin re-bless with full adjudication.

## Verification actually run

- **Discrimination executed**: pre-fix, the new test fails exactly on N→E ([0,5] where B11 demands [5,0]); N→W passed. Post-fix both chiralities green.
- **Full protocol**: suite run surfaced exactly one corpus consequence — `tier4_advanced_circuit_from_ore_am2`'s `input-rate-delivery` pin 7→6. **Adjudicated by position diff before re-blessing** (probe committed, used, deleted): the vanished reading, (87,161) cable 0.39/0.5, clears legitimately under correct routing; the genuinely starved (90,161) REMAINS flagged (0.0→0.24/0.5); every surviving delivered value rose. The #644 fabricated-read class, chirality edition. No winner flipped — all 75 other e2e tests, entity counts, and verdicts unchanged; clippy + wasm + tsc green.
- Scope caveat stated in the re-bless commit: adjudication is game-rule + forensic, not a fresh sim of this fixture; the family's aggregate starvation stays sim-anchored (#519/#644 receipts).

## Why it was invisible

Swapping two equal numbers is the identity — the bug required an asymmetric-lane belt entering a turn of one handedness, which the router's straight-run/UG-heavy style almost never builds. One fixture had exactly one such curve.

## Deviations

None. IRD is selection-participating, so had any winner flipped this PR would have paused for sim anchoring per the standing firewall; none did.


storkme:
Review adjudication (→ cce01cb7):

**Absorbed**: the 3/3 was the best catch of the round — the fn-level contract line still SAID 'lanes swap on CW', a standing invitation to reintroduce the bug; it now states B11 with a pointer to the body's history note. The test input is now [5,2] so BOTH chirality arms discriminate a swap (noted that N→E carried the historical bug).

**Refuted with receipts (the re-bless substantiation)**: 'lane totals are conserved by the swap' is incomplete in exactly the way the finding concedes — totals are conserved AT the transfer tile, but downstream lane-WISE consumers (sideload near/far rules, UG lane admission) split by lane; that is the mechanism. The re-bless commit carries the position forensics run BEFORE blessing: the vanished reading (87,161) cable 0.39/0.5 clears under correct routing; the genuinely starved neighbor (90,161) REMAINS flagged (0.0→0.24/0.5); every surviving delivered value rose monotonically (3.60→3.75, 4.03→4.18) — the uniform signature of the swap depressing modeled lane flow on that run. No true positive hidden; no winner flipped (75/75 other e2e + entity pins unchanged); the game-rule-not-fresh-sim scope caveat is stated in the same commit per protocol.

storkme:
**Contamination disclosure + fix-forward** (→ 6ad804b0): commit cce01cb7 silently included nine files beyond its stated scope — an unrelated in-flight deletion's payload plus a mutated `sat-zones-ci.bin` — because a failed prior commit attempt left the index staged and the follow-up commit swept it. All nine paths are restored to origin/main in the new head; **this PR's true diff is again exactly the chirality fix** (belt_flow.rs + its test + the adjudicated ac-am2 pin). The displaced work ships separately with its own verification. Re-review the current head, not cce01cb7.

storkme:
Round-3 adjudication (→ d423a348): all four 1/3s reviewed, three absorbed, one answered —

**Absorbed**: (1) the followups doc's bug bullet and instrument-gap item 2 still carried the PRE-inversion accusation — the exact re-add-the-swap hazard you named; both now point at the adjudication with an explicit 'do NOT fix the meter toward a swap'. (2) The 'meter as arbiter' phrasing is scoped honestly: the meter AGREED but could not discriminate (lane-identity-blind in throughput terms); the rule + both models' handed seeding decided it. (3) The LANE_LEFT label inversion vs B3 is noted at the site with the do-not-fix-into-a-swap warning — the fix is label-independent.

**Answered (sim anchor)**: the disclaimer you quote is deliberate protocol compliance, not an omission — the firewall requires sim anchoring for SELECTION changes, and no winner flipped (75/75 fixtures + entity pins unchanged; only a warning census moved, with position forensics attached). The IRD family on this fixture remains aggregate-sim-anchored from the #519/#644 receipts; a belt-and-braces fresh sim of ac-am2 is a fine follow-up and is noted on #675, but blocking a game-rule-corroborated latent-bug fix on it would invert the protocol's intent.

PR #684, opened 2026-08-20
del(t2e): remove the never-sim-anchored inserter-throughput check pair (item 9)

storkme:
## Intent

The audit's final owner decision (item 9, 2026-08-21): delete `check_inserter_throughput` + `check_inserter_item_throughput` — demoted from selection 2026-08-14 (#632 B6) as never-sim-anchored hand-capacity models, report-only since.

## Scope

Both checks + their test sections + two orphaned helpers (−~950 lines; `belt_drop_throughput` survives `cfg(test)`-gated for the live lane-budget tests). Selection semantics unchanged: the exclusion const shrinks to `belt-detour` with its deliberate-membership pin updated in the same file. Registry duty in-PR: trust-table rows tombstoned; status.md records the accepted cost (production-science's 8-warning residual is now validator-unreported; the meter still sees the class; the big-electric-pole canary's protection is RFC-059's teeth test); three warning-pin goldens re-blessed (their pinned lines were exactly these categories); CLAUDE.md check count 39→37.

**One downstream consumer surfaced and re-instrumented**: `ec15_chain_inserter_clean_at_default_capacity` proved RFC-049's L2-clears-the-#383-bind via the deleted checks' warnings — its premise-guard now lives in geometry form (L0 vs L2 hashes must diverge; Error-freedom asserted on both), with the history documented at the site.

## Verification actually run

Full protocol, fully green this time: suite 26/26 result lines (pinned cache, pin restored, absent from diff — verified by `git status` before commit), clippy, wasm-pack, tsc.

## Provenance note

This payload briefly contaminated #683 via a staging-area sweep (disclosed and fix-forwarded there); this branch is its clean reconstruction against current main, re-verified end to end.

## Deviations

None from the owner decision.


storkme:
Review adjudication (→ 159ddf92): both minors absorbed — hole 2's 're-lift' closing now records the deletion with the sim-anchored re-introduction bar, and the selection-contract doc's 'remain REPORTED unchanged' is tombstoned in place. On the ec15 note: agreed it weakens the original semantic guarantee — that is the owner-accepted cost of item 9, and the geometry-divergence form is the strongest premise-guard available without the deleted model (stated at the site).

PR #685, opened 2026-08-20
fix(b2): meter batch — chirality adjudication locked, mirror flag parsed, turbo UG arm

storkme:
## Intent

B2 from #675, reshaped by the chirality adjudication: the meter was CLEARED (the swap was core's bug, #683), so the meter side is now locks-and-gaps rather than a fix.

## Scope

- **Chirality lock**: a second-chirality curve test in the meter, so neither arm can regress toward the walker's old swap; comment carries the adjudication story.
- **Mirror flag**: decoder parses blueprint `mirror` (explicit community `mirror: true` honored on any machine); the three-machine name heuristic stays for the engine's tile-identical 180° encoding, with the on-the-wire-undecidable residual documented in `meter-divergence.md` rather than guessed.
- **Turbo UG arm**: core `ug_max_reach` gains the explicit `turbo-transport-belt => 10` arm (imported blueprints previously fell to the yellow-4 fallback — the meter/core tier asymmetry the audit flagged).
- `meter-divergence.md` entry recording all three.

## Verification actually run

Meter suites green (58 lib + integration, incl. the new lock test); full core protocol on this commit (suite 26/26 result lines, pinned cache restored and absent from diff, clippy, wasm, tsc) — the turbo arm is corpus-neutral (the engine never places turbo).

## Deviations

The audit item said 'fix the meter chirality gap'; the adjudication inverted it, so the deliverable became clearing + locking the meter and fixing core (#683). Recorded in the domain audit doc and divergence log.

### DIFF
--- a/crates/core/src/common.rs
+++ b/crates/core/src/common.rs
@@ -1363,6 +1363,8 @@ mod tests {
         assert_eq!(ug_max_reach("transport-belt"), 4);
         assert_eq!(ug_max_reach("fast-transport-belt"), 6);
         assert_eq!(ug_max_reach("express-transport-belt"), 8);
+        assert_eq!(ug_max_reach("turbo-transport-belt"), 10);
+        assert_eq!(ug_max_reach("not-a-belt"), 4);
     }
 
     #[test]
--- a/crates/core/src/common.rs
+++ b/crates/core/src/common.rs
@@ -438,6 +438,11 @@ pub fn ug_max_reach(belt: &str) -> u32 {
         "transport-belt" => 4,
         "fast-transport-belt" => 6,
         "express-transport-belt" => 8,
+        // Space Age turbo tier: the engine never PLACES it, but imported
+        // blueprints can carry it and the meter models it — before
+        // 2026-08-21 (offpath B2) it fell through to the yellow fallback,
+        // one data change away from a silent 4-vs-10 misreach.
+        "turbo-transport-belt" => 10,
         _ => 4,
     }
 }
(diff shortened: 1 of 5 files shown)


storkme:
Review adjudication (→ c0ea09c2) — the mirror major was correct and sharpened the design:

**Absorbed (mirror)**: honoring `mirror:true` as a binding flip alone would mis-place ports on asymmetric machines — and the sharper realization your finding forced: for the three x-symmetric engine machines the name heuristic already applies the flip, so the `e.mirror ||` only ever ACTED on exactly the unsound case. Reverted to parse-only: the flag is decoder knowledge (pinned by a decode round-trip test), the `reflect_port` (w−1−dx) requirement is documented at the site and in the divergence entry, and behavior is unchanged from pre-PR for every machine.

**Absorbed (test framing)**: the meter chirality test's comment now states it is non-discriminating today (only_feeder never consults chirality) and what it actually pins — a future refactor ADDING a chirality distinction fails loudly; the discriminating asymmetric-lane two-arm lock lives in belt_flow's test on **#683**.

**Absorbed (cross-PR ordering)**: all belt_flow/chirality/ac-am2 claims now cite #683 by PR number rather than asserting tree state — the 'fix landed / fixture unreachable' contradictions you flagged were artifacts of #683 being in review while this branch is cut from main; the references are now landing-order-independent.

**Absorbed (nit)**: turbo arm added to `ug_max_reach_values`. The followups-direction nit is fixed on #683's branch (its round 3), same ordering artifact.

Meter suite 59+ green incl. both new tests; core validate lib 282/282.

PR #686, opened 2026-08-20
diag(g2): check-firing census — categories that fire on evaluated layouts, not just winners

storkme:
## Intent

G2 from #675: the instrument the audit said must exist before any further structural-check deletion — makes the validator's invisible selection work observable.

## Scope

One `#[ignore]` diagnostic test (`check_firing_census.rs`). v1 approximates the candidate field with the public option-toggle family (6 fixtures × 6 variants); the module doc states the honest limit (the internal k1/size-split/merge-tap members are not re-enacted — the census moves onto the real loop if selection-unification lands).

## Verification actually run

First census executed (pinned cache): 36 builds, 0 refusals, 6 firing categories — and a real finding: at the option level **no category is loser-only** (everything that fires also fires on a default build), which sharpens that the invisible-work question lives in the internal candidate field.

## Deviations

None.

### DIFF
--- a/crates/core/tests/check_firing_census.rs
+++ b/crates/core/tests/check_firing_census.rs
@@ -16,6 +16,17 @@
 //! approximation. Fixture list is a hardcoded tier-ladder slice for the
 //! same reason; extend as needed.
 //!
+//! A second, orthogonal gap (#686 round 2): a variant whose
+//! `build_bus_layout` call returns `Err` (refused) produces no layout at
+//! all, so nothing gets validated and no category can be attributed to
+//! it. Refusal counts are tracked per-variant (see the printed summary),
+//! but WHICH categories would have fired on a refused candidate is
+//! invisible at this layer — the real fix is the search's internal
+//! candidate loop (same refactor #1 above), not anything this census can
+//! do post-hoc. "Fired on NOTHING evaluated here" therefore means "not
+//! observed," not "inert" — it is silent on categories that only ever
+//! appear inside refused candidates.
+//!
 //! Run: cargo test --test check_firing_census -- --ignored --nocapture
 
 use std::collections::BTreeSet;
@@ -26,22 +37,62 @@ use spaghettio_core::bus::di_cell::DirectInsertion;
 use spaghettio_core::bus::layout::{self, LayoutOptions, LayoutStrategy};
 use spaghettio_core::{solver, validate};
 
-/// Per-category census row. `winner` / `variants` / `err_variants` are kept
-/// separate rather than folded into one flag because candidate refusal
-/// (`decomposition_search.rs`) keys on `Severity::Error` only — a category
-/// that fires Warnings on a non-default variant but never an Error is
-/// invisible to selection, and must not be conflated with one that does.
+/// Per-category census row.
+///
+/// Two independent things go wrong if this is read naively, both found in
+/// #686 round 2 review:
+///
+/// 1. Candidate refusal (`decomposition_search.rs`) is supposed to key on
+///    `Severity::Error` only, so a Warning-only firing on a non-default
+///    variant is not by itself evidence of invisible selection work — but
+///    `select_best_decomposition`'s error-free tier (`best_error_free_idx`)
+///    only WINS the pick when some candidate validates clean; if the whole
+///    ranking is error-laden, selection falls through to the best-accepted
+///    candidate and ships it anyway (see the comment above
+///    `best_error_free_idx` in `decomposition_search.rs`: "still returns
+///    the error-laden best rather than refusing"). A default build CAN
+///    carry a `Severity::Error`, so `err-loser` must not be gated on the
+///    any-severity `winner` flag — the two are unrelated facts.
+/// 2. `winner` and a naive `err_variants.is_empty()` check are both
+///    accumulated ACROSS all six fixtures. A category that Error-fires on
+///    `di-off` for fixture A but only Warns on `default` for unrelated
+///    fixture B would read `winner=true` from B and mask the fixture-A
+///    finding. `fixtures_err_default` / `fixtures_err_nondefault` key on
+///    fixture identity so `err-loser` is computed per-fixture and then
+///    unioned, not aggregated blind across fixtures.
 #[derive(Default)]
 struct CatRow {
-    /// Fired (any severity) on the "default" variant — the one the engine
-    /// actually ships.
+    /// Fired (any severity) on SOME fixture's "default"-options build. NOT
+    /// "the layout the engine ships" — `select_best_decomposition` can
+    /// still ship an Error-carrying candidate when nothing validates
+    /// clean; see point 1 above.
     winner: bool,
-    /// Every variant name this category fired on, "default" included.
+    /// Every variant name this category fired on, "default" included,
+    /// across all fixtures.
     variants: BTreeSet<&'static str>,
-    /// Variant names where this category fired at `Severity::Error`.
+    /// Every variant name where this category fired at `Severity::Error`,
+    /// across all fixtures. Informational provenance only — NOT the
+    /// err-loser predicate itself, which needs fixture pairing (point 2
+    /// above) rather than a flat union.
     err_variants: BTreeSet<&'static str>,
+    /// Fixture keys (`"{item}@{rate}"`) where this category fired at
+    /// `Severity::Error` on that fixture's OWN "default" build.
+    fixtures_err_default: BTreeSet<String>,
+    /// Fixture keys where this category fired at `Severity::Error` on
+    /// some NON-default variant of that fixture.
+    fixtures_err_nondefault: BTreeSet<String>,
     /// Total issue count across all builds.
     count: usize,
+}
+
+impl CatRow {
+    /// True iff some fixture Error-fired this category on a non-default
+    /// variant while that SAME fixture's own default build did not —
+    /// i.e. invisible-to-selection Error work that a different fixture's
+    /// quiet default cannot mask.
+    fn err_loser(&self) -> bool {
+        self.fixtures_err_nondefault.iter().any(|f| !self.fixtures_err_default.contains(f))
+    }
 }
 
 #[test]
@@ -84,6 +135,7 @@ fn check_firing_census() {
             eprintln!("SKIP (no solve): {item}@{rate}");
             continue;
         };
+        let fixture_key = format!("{item}@{rate}");
         for (vname, tweak) in variants {
             let mut opts = LayoutOptions::default();
             tweak(&mut opts);
@@ -108,6 +160,11 @@ fn check_firing_census() {
                 }
                 if i.severity == validate::Severity::Error {
                     row.err_variants.insert(*vname);
+                    if *vname == "default" {
+                        row.fixtures_err_default.insert(fixture_key.clone());
+                    } else {
+                        row.fixtures_err_nondefault.insert(fixture_key.clone());
+                    }
                 }
             }
         }
@@ -125,41 +182,52 @@ fn check_firing_census() {
     };
     println!("\n=== check-firing census: {builds} builds, refusals: {refusal_summary} ===");
     println!(
-        "{:<32} {:>7} {:>10} {:>9} {:>6}  {}",
-        "category", "winner", "loser-only", "err-loser", "count", "variants"
+        "{:<32} {:>7} {:>10} {:>9} {:>6}  {:<24}  {}",
+        "category", "winner", "loser-only", "err-loser", "count", "err-variants", "variants"
     );
     for (cat, row) in &rows {
-        // "loser-only" and "err-loser" both read as "fired on a non-default
-        // variant, never on default" — the difference is which severity is
-        // required on that non-default firing. `winner` already tracks
-        // "fired on default at any severity", so `!winner` alone covers the
-        // "never on default" half for both columns.
+        // "loser-only" = fired (any severity) on a non-default variant,
+        // never on default; `winner` already tracks "fired on SOME
+        // fixture's default at any severity", so `!winner` covers the
+        // "never on default" half. "err-loser" is NOT `!winner` — see
+        // `CatRow::err_loser` and the struct doc: it is computed per
+        // fixture so a quiet-on-B default can't mask an Error found on A.
         let loser_only = !row.winner;
-        let err_loser = !row.winner && !row.err_variants.is_empty();
+        let err_loser = row.err_loser();
+        let err_variants_str = row.err_variants.iter().copied().collect::<Vec<_>>().join(",");
+        let err_variants_disp = if err_variants_str.is_empty() { "-" } else { &err_variants_str };
         let variants_str = row.variants.iter().copied().collect::<Vec<_>>().join(",");
         println!(
-            "{:<32} {:>7} {:>10} {:>9} {:>6}  {}",
+            "{:<32} {:>7} {:>10} {:>9} {:>6}  {:<24}  {}",
             cat,
             if row.winner { "yes" } else { "-" },
             if loser_only { "YES" } else { "-" },
             if err_loser { "YES" } else { "-" },
             row.count,
+            err_variants_disp,
             variants_str
         );
     }
     println!(
-        "\nInterpretation: candidate refusal keys on Severity::Error only \
-         (decomposition_search refuses Error-carrying candidates; warnings \
-         pass), so only err-loser=YES categories are evidence of invisible \
-         selection work. loser-only=YES with warnings alone means the \
-         category fires on shapes the corpus never ships — useful for \
-         coverage, silent on selection. The variants column says WHICH \
-         shapes: cells-off/partitioned are user-elected or alternate-path \
-         builds the native search does not enumerate (v1 approximation, \
-         module doc), so firings confined to them are 'fired on a \
-         non-default pipeline', not 'refused within the search'. \
-         Categories absent entirely fired on NOTHING evaluated here (v1 \
-         scope caveats in the module doc apply before concluding they are \
-         inert)."
+        "\nInterpretation: default builds are NOT guaranteed error-free — \
+         select_best_decomposition ships the best-scoring candidate even \
+         when none validates clean, so gating err-loser on the \
+         any-severity `winner` flag was wrong (round 2, #686). err-loser \
+         is now computed per fixture: YES iff some fixture Error-fired \
+         this category on a non-default variant while that SAME \
+         fixture's own default build did not. cell_composition and \
+         direct_insertion both default to Candidate, so the native \
+         search ALREADY tries the cells-off / DI-free shape internally \
+         as its own baseline candidate — firings confined to \
+         cells-off/di-off/di-forced/hs-off are native-adjacent, not \
+         outside the search. Only `partitioned` \
+         (LayoutStrategy::PartitionedDecomposed) is a genuinely \
+         separate, user-elected top-level strategy the native search \
+         never runs. Refused builds (Err from build_bus_layout) produce \
+         no layout to validate, so no category attribution is possible \
+         for them — 'fired on NOTHING evaluated here' cannot distinguish \
+         a genuinely inert category from one that only ever appears \
+         inside refused candidates (v1 scope caveats in the module doc \
+         apply before concluding a quiet category is inert)."
     );
 }
--- a/crates/core/tests/check_firing_census.rs
+++ b/crates/core/tests/check_firing_census.rs
@@ -173,7 +173,13 @@ fn check_firing_census() {
             eprintln!("SKIP (no solve): {item}@{rate}");
             continue;
         };
-        let fixture_key = format!("{item}@{rate}");
+        // Machine tier included (round 4, #686): "{item}@{rate}" alone
+        // collides if the fixture list ever gains a second tier at the
+        // same item/rate (the module doc invites exactly that extension),
+        // which would silently merge two fixtures' per-fixture state and
+        // corrupt the err_loser/loser_only pairing this file exists to
+        // protect.
+        let fixture_key = format!("{item}@{rate}:{machine}");
         for (vname, tweak) in variants {
             let mut opts = LayoutOptions::default();
             tweak(&mut opts);
@@ -229,6 +235,11 @@ fn check_firing_census() {
         "(a category ABSENT below was never observed on a built layout \
          here — indistinguishable from 'only fires on refused \
          candidates'; see Interpretation at the bottom)"
+    );
+    println!(
+        "(count sums every issue across all variants and both severities \
+         for that category — it is not per-variant or per-severity; use \
+         err-variants/variants for provenance)"
     );
     println!(
         "{:<32} {:>7} {:>10} {:>9} {:>6}  {:<24}  {}",
--- a/crates/core/tests/check_firing_census.rs
+++ b/crates/core/tests/check_firing_census.rs
@@ -37,16 +37,30 @@ use rustc_hash::{FxHashMap, FxHashSet};
 use spaghettio_core::bus::cells::CellComposition;
 use spaghettio_core::bus::di_cell::DirectInsertion;
 use spaghettio_core::bus::layout::{self, LayoutOptions, LayoutStrategy};
+use spaghettio_core::models::LayoutResult;
 use spaghettio_core::{solver, validate};
 
-/// Variants the native `Candidate`-mode search actually evaluates as one of
-/// its own baseline/competing arms — established with receipts in round 3's
-/// #686 adjudication (`decomposition_search.rs`'s `try_cells`/
-/// `try_horizontal`/`try_di` gates, and the `DirectInsertion::Off` doc
-/// comment). `di-forced` and `partitioned` are genuinely separate,
-/// user-elected topologies the default search never tries on its own — see
-/// point 4 below.
-const NATIVE_ADJACENT_VARIANTS: &[&str] = &["di-off", "cells-off", "hs-off"];
+/// A structural signature of a built layout's entities — name/x/y/
+/// direction/recipe, sorted so emission order doesn't matter. Used only to
+/// detect when a variant's build is bit-identical to the same fixture's
+/// default (round 6, #686): `cells-off`/`hs-off` are gated internally
+/// (`decomposition_search.rs`'s `try_cells`/`try_horizontal`) on
+/// conditions this hardcoded fixture list may not satisfy for every
+/// fixture (chain eligibility, a `DualInput` row), so a variant can be a
+/// silent no-op — bit-identical to default — for some or all fixtures,
+/// which would otherwise look like a genuinely evaluated, merely-quiet
+/// candidate.
+type EntitySignature = Vec<(String, i32, i32, u8, Option<String>)>;
+
+fn layout_signature(l: &LayoutResult) -> EntitySignature {
+    let mut sig: Vec<_> = l
+        .entities
+        .iter()
+        .map(|e| (e.name.clone(), e.x, e.y, e.direction as u8, e.recipe.clone()))
+        .collect();
+    sig.sort();
+    sig
+}
 
 /// Per-category census row.
 ///
@@ -88,10 +102,12 @@ const NATIVE_ADJACENT_VARIANTS: &[&str] = &["di-off", "cells-off", "hs-off"];
 ///    says nothing about a candidate the search silently refused, because
 ///    the search never tried that shape at all. `fixtures_any_nondefault`/
 ///    `fixtures_err_nondefault` therefore only accumulate fixture keys from
-///    `NATIVE_ADJACENT_VARIANTS` (`di-off`/`cells-off`/`hs-off`); `di-forced`/
-///    `partitioned` firings still show up in `variants`/`err_variants` and
-///    in `count` — visible, just not counted as loser-only/err-loser
-///    evidence.
+///    variants tagged `native_adjacent: true` in the `variants` table below
+///    (`di-off`/`cells-off`/`hs-off`, round 6 — the flag lives in the same
+///    tuple the loop iterates, so it cannot desync from that table the way
+///    a separately-declared parallel list could); `di-forced`/`partitioned`
+///    firings still show up in `variants`/`err_variants` and in `count` —
+///    visible, just not counted as loser-only/err-loser evidence.
 #[derive(Default)]
 struct CatRow {
     /// Fired (any severity) on SOME fixture's "default"-options build. NOT
@@ -134,14 +150,16 @@ struct CatRow {
 
 impl CatRow {
     /// True iff some fixture fired this category (any severity) on a
-    /// non-default, NATIVE-ADJACENT variant (point 4, struct doc — the
-    /// input set already excludes `di-forced`/`partitioned`) while that
-    /// SAME fixture's own default build (1) actually produced a layout AND
-    /// (2) never fired it. `defaults_built` is the global set of fixture
-    /// keys whose "default" variant built successfully — without that
-    /// gate, a fixture whose default REFUSED would look identical to one
-    /// whose default built clean and simply didn't fire, which is not
-    /// evidence of anything (point 3, struct doc).
+    /// non-default, native-adjacent variant (point 4, struct doc — the
+    /// `fixtures_any_nondefault` input set already excludes `di-forced`/
+    /// `partitioned`, gated at insertion time by each variant's own
+    /// `native_adjacent` tuple field) while that SAME fixture's own default
+    /// build (1) actually produced a layout AND (2) never fired it.
+    /// `defaults_built` is the global set of fixture keys whose "default"
+    /// variant built successfully — without that gate, a fixture whose
+    /// default REFUSED would look identical to one whose default built
+    /// clean and simply didn't fire, which is not evidence of anything
+    /// (point 3, struct doc).
     fn loser_only(&self, defaults_built: &FxHashSet<String>) -> bool {
         self.fixtures_any_nondefault
             .iter()
@@ -178,14 +196,29 @@ fn check_firing_census() {
             &["iron-ore", "copper-ore", "coal", "water", "crude-oil"],
         ),
     ];
-    let variants: &[(&str, fn(&mut LayoutOptions))] = &[
-        ("default", |_| {}),
-        ("di-off", |o| o.direct_insertion = DirectInsertion::Off),
-        ("di-forced", |o| o.direct_insertion = DirectInsertion::Forced),
-        ("cells-off", |o| o.cell_composition = CellComposition::Off),
-        ("hs-off", |o| o.horizontal_candidate = false),
-        ("partitioned", |o| o.strategy = LayoutStrategy::PartitionedDecomposed),
+    // Third field: `native_adjacent` — is this a shape the native
+    // `Candidate`-mode search evaluates as one of its own baseline/
+    // competing arms (round 3's receipts: decomposition_search.rs's
+    // `try_cells`/`try_horizontal`/`try_di` gates, `DirectInsertion::Off`'s
+    // doc comment)? `di-forced`/`partitioned` are `false` — genuinely
+    // separate, user-elected topologies the default search never tries.
+    // Carried in the tuple itself (round 6, #686) rather than a
+    // separately-declared list, so the flag cannot desync from the table
+    // the loop below actually iterates.
+    type Variant = (&'static str, fn(&mut LayoutOptions), bool);
+    let variants: &[Variant] = &[
+        ("default", |_| {}, false),
+        ("di-off", |o| o.direct_insertion = DirectInsertion::Off, true),
+        ("di-forced", |o| o.direct_insertion = DirectInsertion::Forced, false),
+        ("cells-off", |o| o.cell_composition = CellComposition::Off, true),
+        ("hs-off", |o| o.horizontal_candidate = false, true),
+        ("partitioned", |o| o.strategy = LayoutStrategy::PartitionedDecomposed, false),
     ];
+    let native_adjacent_names: Vec<&str> = variants
+        .iter()
+        .filter(|(_, _, native_adjacent)| *native_adjacent)
+        .map(|(name, _, _)| *name)
+        .collect();
 
     let mut census: FxHashMap<String, CatRow> = FxHashMap::default();
     let mut builds = 0usize;
@@ -195,6 +228,16 @@ fn check_firing_census() {
     // fixture's default build succeeded doesn't depend on which category
     // we're looking at.
     let mut defaults_built: FxHashSet<String> = FxHashSet::default();
+    // Round 6, #686: mechanical no-op detection. `cells-off`/`hs-off` are
+    // gated internally on conditions this hardcoded fixture list may not
+    // satisfy for every fixture (`try_cells`'s chain-eligibility check,
+    // `try_horizontal`'s DualInput-row requirement — decomposition_
+    // search.rs), so a variant can be bit-identical to default for some or
+    // all fixtures without that being visible anywhere in the table above.
+    // Per-variant: how many non-default builds succeeded, and how many of
+    // those were structurally identical to that SAME fixture's default.
+    let mut nondefault_builds: FxHashMap<&'static str, usize> = FxHashMap::default();
+    let mut noop_count: FxHashMap<&'static str, usize> = FxHashMap::default();
 
     for &(item, rate, machine, inputs) in fixtures {
         let input_set: FxHashSet<String> = inputs.iter().map(|s| s.to_string()).collect();
@@ -209,7 +252,10 @@ fn check_firing_census() {
         // corrupt the err_loser/loser_only pairing this file exists to
         // protect.
         let fixture_key = format!("{item}@{rate}:{machine}");
-        for (vname, tweak) in variants {
+        // Reset per fixture — no-op comparison is always against THIS
+        // fixture's own default, never another fixture's.
+        let mut default_sig: Option<EntitySignature> = None;
+        for (vname, tweak, native_adjacent) in variants {
             let mut opts = LayoutOptions::default();
             tweak(&mut opts);
             let l = match layout::build_bus_layout(&sr, opts) {
@@ -222,6 +268,13 @@ fn check_firing_census() {
             builds += 1;
             if *vname == "default" {
                 defaults_built.insert(fixture_key.clone());
+                default_sig = Some(layout_signature(&l));
+            } else {
+                *nondefault_builds.entry(*vname).or_insert(0) += 1;
+                let is_noop = default_sig.as_ref().is_some_and(|d| *d == layout_signature(&l));
+                if is_noop {
+                    *noop_count.entry(*vname).or_insert(0) += 1;
+                }
             }
             let issues = match validate::validate(&l, Some(&sr)) {
                 Ok(i) => i,
@@ -234,7 +287,7 @@ fn check_firing_census() {
                 if *vname == "default" {
                     row.winner = true;
                     row.fixtures_any_default.insert(fixture_key.clone());
-                } else if NATIVE_ADJACENT_VARIANTS.contains(vname) {
+                } else if *native_adjacent {
                     // di-forced/partitioned still land in `variants` above
                     // (display/provenance) but are excluded here — they
                     // are user-elected topologies the native search never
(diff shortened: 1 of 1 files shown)


storkme:
Absorbed the three valid findings, refuted the rest with receipts. New head: `d5497e97d5fc7c04cc9267300f8ef917c84a4b41`.

**Premise check** — confirmed before absorbing anything: `decomposition_search.rs` filters `i.severity == crate::validate::Severity::Error` at 8 call sites (candidate refusal keys on Error only; Warnings pass through). This is the load-bearing fact behind majors 1 and 2 below.

## Findings

- **[major] Severity aggregation hides that refusal keys on Error only** (3/3 passes) — **Absorbed.** Replaced the `(bool, bool, usize)` tuple with a `CatRow` struct (`winner`, `variants: BTreeSet<&'static str>`, `err_variants: BTreeSet<&'static str>`, `count`). New `err-loser` column = fired at `Severity::Error` on a non-default variant, never on default — the only column that's evidence of invisible *selection* work. `loser-only` is kept as-is (any severity) so warning-only coverage firings are still visible but no longer conflated with selection-relevant ones. Interpretation text rewritten to state this explicitly.

- **[major] `loser-only` doesn't distinguish native-refused shapes from non-default pipelines the search never runs** (2/3 passes) — **Absorbed via reword, not gating.** Didn't gate `cells-off`/`partitioned` on the default build having refused (that would need the real candidate loop, out of v1 scope per the module doc). Instead the new `variants` column prints exactly which variant(s) fired per category, and the interpretation text now says firings confined to `cells-off`/`partitioned` are "fired on a non-default pipeline," not "refused within the search."

- **[major, self-flagged unverified, 1/3] lowercase `fn(&mut LayoutOptions)` may not parse** — **Refuted.** `fn(&mut LayoutOptions)` is Rust's function-pointer type; the six closure literals (including captureless ones) coerce to it, same as every other `fn`-typed slice in the crate. The `rust` check on the exact reviewed SHA (`4954638bee11de455f6c1fb7dd14d6f3295ce5ce`) is green: `rust: success`, `rust-clippy: success`. Confirmed unverified self-flag was warranted caution but the claim doesn't hold.

- **[major, 1/3] Census can't observe internally-refused candidates** — **Refuted-as-disclosed.** This is the module doc's stated v1 scope, verbatim: "the candidate field is approximated by the public option-toggle family (native / DI-off / cells-off / HS-off / DI-forced / partitioned), not by the search's internal candidate list — k1-shape-fix and size-split members only exist inside `select_best_decomposition` on native failure and are not re-enacted here." Known and disclosed limitation, not a discovered flaw. The absorbed severity/provenance columns tighten the reading available *within* that disclosed scope; they don't and can't close it.

- **[minor, 2/3] Refused variants aren't attributed** — **Absorbed.** Refusals are now counted per variant name (`FxHashMap<&str, usize>`) and printed in the summary header (`refusals: di-off=2 partitioned=1`, or `refusals: none` — this run hit `none`).

- **[minor, 2/3] Provenance destroyed before the report (category/variant conflation)** — **Absorbed.** The `variants` BTreeSet on `CatRow` is the per-category provenance fix requested; every row now prints exactly which variant(s) it fired on rather than a single opaque loser-only bit. (Not keyed to per-fixture as the finding's stronger alternative suggested — variant-level provenance is what the absorbed columns needed to support the severity distinction; fixture-level is a further cut left to a future pass if it's ever needed.)

- **[minor, 1/3] Non-deterministic tie-break ordering** — **Absorbed.** `rows.sort_by(|a, b| b.1.count.cmp(&a.1.count).then_with(|| a.0.cmp(&b.0)))` — count desc, category name asc as tiebreak.

- **[minor, 2/3] Small fixture coverage buries the v1-scope caveat in a comment** — **Refuted-as-disclosed.** Same module-doc v1-scope paragraph as above; the printed interpretation text already points at it explicitly ("v1 scope caveats in the module doc apply before concluding they are inert").

- **Nits** (coverage/flag saturation, refusals pooled without attribution, interpretation over-generalizing, SKIP fixtures dropped from counts) — the attribution and saturation nits are covered by the absorptions above (provenance columns, per-variant refusal counts). Left the SKIP-fixtures-dropped-from-header-counts nit as-is; this run hit zero SKIPs so it wasn't exercised, and it's a pre-existing property of the `builds` counter, not something the review's majors/minors asked to change.

## Fresh census (this run: 36 builds, 0 refusals, 0 SKIPs)

```
=== check-firing census: 36 builds, refusals: none ===
category                          winner loser-only err-loser  count  variants
belt-flow-reachability               yes          -         -    240  cells-off,default,di-forced,di-off,hs-off,partitioned
input-rate-delivery                  yes          -         -    179  cells-off,default,di-forced,di-off,hs-off,partitioned
belt-flow-path                       yes          -         -     60  cells-off,default,di-forced,di-off,hs-off,partitioned
belt-detour                          yes          -         -     33  cells-off,default,di-forced,di-off,hs-off,partitioned
belt-dead-end                        yes          -         -     18  cells-off,default,di-forced,di-off,hs-off,partitioned
row-input-belt-margin                yes          -         -     18  cells-off,default,di-forced,di-off,hs-off,partitioned

Interpretation: candidate refusal keys on Severity::Error only (decomposition_search refuses Error-carrying candidates; warnings pass), so only err-loser=YES categories are evidence of invisible selection work. loser-only=YES with warnings alone means the category fires on shapes the corpus never ships — useful for coverage, silent on selection. The variants column says WHICH shapes: cells-off/partitioned are user-elected or alternate-path builds the native search does not enumerate (v1 approximation, module doc), so firings confined to them are 'fired on a non-default pipeline', not 'refused within the search'. Categories absent entirely fired on NOTHING evaluated here (v1 scope caveats in the module doc apply before concluding they are inert).
```

On this fixture slice every firing category fires on every variant including `default`, so `loser-only`/`err-loser` are both all-`-` here — that's expected given the v1 scope's small hardcoded fixture list, not evidence the new columns are inert; they'll discriminate once the fixture set or variant space grows enough to produce a category that fires only on a non-default build.

New head SHA: `d5497e97d5fc7c04cc9267300f8ef917c84a4b41`

storkme:
Round 2 absorbed. New head: `a03b34839f542dd71ea3082315151001b91712fd`.

## Premise checks (done before touching code)

- **"select_best_decomposition always ships an error-free default"** — **False, confirmed in code.** `decomposition_search.rs`'s `best_error_free_idx` (~line 1528-1551) only wins the pick when some candidate validates clean (`clean_flags[i]?` + `if !clean { return None; }` filters the *error-free tier*, not the whole function). The surrounding comment states explicitly: "if none is clean, fall through to today's pick (**still returns the error-laden best rather than refusing** — callers see the errors, behavior unchanged)." A default build CAN carry `Severity::Error`. This falsifies round 1's `err_loser = !row.winner && !row.err_variants.is_empty()` gate and is why major 1/2 below are valid.

- **"cell_composition/direct_insertion default to Candidate, whose native pass IS the cells-off/DI-free layout"** — **True, confirmed in code.** `LayoutOptions::default()` (layout.rs ~241, ~247) sets `cell_composition: CellComposition::Candidate` and `direct_insertion: DirectInsertion::Candidate`, with comments confirming cells only engage "where the bus path refuses or fails acceptance" and "under Candidate the native pass is DI-free." So cells-off/di-off are shapes the native search already tries internally as its own baseline candidate, not an outside alternate pipeline. Only `partitioned` (`LayoutStrategy::PartitionedDecomposed`) is a genuinely separate top-level strategy.

## Findings

- **[major, 3/3] `err_loser` gated on the any-severity `winner` flag, masking Error-only findings** — **Absorbed.** `err_loser` no longer references `winner` at all. Rewritten per finding 2's fix, see below — computed from fixture-keyed sets, independent of any-severity default firings.

- **[major, 2/3] Cross-fixture masking: `winner`/`err_variants` aggregated across all 6 fixtures** — **Absorbed.** Added `fixtures_err_default: BTreeSet<String>` and `fixtures_err_nondefault: BTreeSet<String>` to `CatRow`, keyed on `"{item}@{rate}"`. `CatRow::err_loser()` = `fixtures_err_nondefault.iter().any(|f| !fixtures_err_default.contains(f))` — true iff some fixture Error-fired this category on a non-default variant while that SAME fixture's own default build did not. A quiet default on an unrelated fixture can no longer mask an Error found on another.

- **[major, 2/3] Refused candidates carry no category attribution** — **Refuted-as-structural + absorbed-as-disclosure.** Refuted the implied fix: a variant whose `build_bus_layout` returns `Err` produces no `LayoutResult` at all, so there is nothing to pass to `validate()` — category attribution at this layer is not a missing feature, it's a missing object. The real fix is the search's internal candidate loop (the module doc's already-named selection-unification refactor #1), not anything post-hoc here. Absorbed the disclosure: added a paragraph to the module doc and a clause in the printed interpretation stating refused builds produce no layout to validate, so "fired on NOTHING evaluated here" cannot distinguish a genuinely inert category from one that only ever appears inside refused candidates.

- **[minor, 2/3] `winner` doc says "the one the engine ships"** — **Absorbed.** `CatRow::winner` doc now reads "Fired (any severity) on SOME fixture's 'default'-options build. NOT 'the layout the engine ships'" with a pointer to the ship-anyway fallback above.

- **[minor, 3/3] `err_variants` collected but never rendered** — **Absorbed.** Added as its own printed column (`err-variants`), separate from the general `variants` column and from the `err_loser` predicate (which needs fixture pairing, not a flat union — see struct doc point 2).

- **[minor, 2/3] Interpretation footer misstates cells-off/di-off as "not enumerated"** — **Absorbed**, per verified premise (b) above. Footer now states cell_composition/direct_insertion default to Candidate so cells-off/di-off/di-forced/hs-off are native-adjacent (the native search's own baseline candidate), and only `partitioned` is genuinely a separate user-elected pipeline.

- **Nits** ("print err_variants for non-loser rows too", "extend module-doc v1 caveat with Err-path attribution stripping", "winner doesn't say what severity the default fired") — first two covered by the absorptions above (err-variants column is now unconditional; module doc has the new refusal-attribution paragraph). Third nit not separately addressed — `winner` is explicitly any-severity by design (see struct doc); splitting it into per-severity default columns wasn't asked for by any major/minor and would just re-derive from `err_variants` intersected with `fixtures_err_default`, which a reader can already do from the printed columns.

## Fresh census (this run: 36 builds, 0 refusals, 0 SKIPs)

```
=== check-firing census: 36 builds, refusals: none ===
category                          winner loser-only err-loser  count  err-variants              variants
belt-flow-reachability               yes          -         -    240  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
input-rate-delivery                  yes          -         -    179  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-flow-path                       yes          -         -     60  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-detour                          yes          -         -     33  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-dead-end                        yes          -         -     18  cells-off,default,di-forced,di-off,hs-off,partitioned  cells-off,default,di-forced,di-off,hs-off,partitioned
row-input-belt-margin                yes          -         -     18  -                         cells-off,default,di-forced,di-off,hs-off,partitioned

Interpretation: default builds are NOT guaranteed error-free — select_best_decomposition ships the best-scoring candidate even when none validates clean, so gating err-loser on the any-severity `winner` flag was wrong (round 2, #686). err-loser is now computed per fixture: YES iff some fixture Error-fired this category on a non-default variant while that SAME fixture's own default build did not. cell_composition and direct_insertion both default to Candidate, so the native search ALREADY tries the cells-off / DI-free shape internally as its own baseline candidate — firings confined to cells-off/di-off/di-forced/hs-off are native-adjacent, not outside the search. Only `partitioned` (LayoutStrategy::PartitionedDecomposed) is a genuinely separate, user-elected top-level strategy the native search never runs. Refused builds (Err from build_bus_layout) produce no layout to validate, so no category attribution is possible for them — 'fired on NOTHING evaluated here' cannot distinguish a genuinely inert category from one that only ever appears inside refused candidates (v1 scope caveats in the module doc apply before concluding a quiet category is inert).
```

Note on `belt-dead-end`: it now shows `err-variants` populated across every variant including `default`, but `err-loser=-`. That's correct under the fixed predicate — this category Error-fires on its own fixture's default build too, so it's visible in the shipped corpus already, not hidden by selection. On this fixture slice no category actually trips `err-loser=YES`; that's an artifact of the small hardcoded fixture list (every category here fires identically across all 6 variants), not evidence the fixed columns are inert — the per-fixture masking bug they fix requires a fixture where the same category behaves differently across variants, which this slice doesn't currently exercise.

New head SHA: `a03b34839f542dd71ea3082315151001b91712fd`

storkme:
Round 3 absorbed. New head: `1d3886a4d835915bbb7f296c03242d1b3f73dcfc`.

## Findings

- **[major, 3/3] `err_loser()` can't distinguish "default built clean" from "default never built"** — **Absorbed.** Added a global `defaults_built: FxHashSet<String>`, populated whenever a fixture's `default` variant successfully produces a layout (regardless of category). `err_loser()`/`loser_only()` now require the fixture to be in that set before treating an empty `fixtures_{err,any}_default` entry as "default was clean" rather than "default was never built to check." **This run's data:** 36/36 builds, 0 refusals, so every fixture's default built — the hazard this fixes never materialized here; the fix hardens the instrument for when a default refuses in a future run.

- **[major, 3/3] `loser-only` still used the cross-fixture `!winner` proxy** — **Absorbed, and not hypothetical.** Added `fixtures_any_default`/`fixtures_any_nondefault` (any-severity counterparts to the `err_` sets) and `CatRow::loser_only()`, computed the same per-fixture way as `err_loser()`. **On this run's actual data, this changed the printed table**: `input-rate-delivery` and `belt-detour` now show `loser-only=YES`, which the old aggregate `!winner` proxy incorrectly hid as `-` in rounds 1 and 2 (some fixture's non-default variant fired the category while that same fixture's own successful default did not — a genuine per-fixture loser-only case round 1/2's aggregation masked). This is direct evidence the bug was real, not just theoretically possible.

- **[minor, 1/3] Absence caveat buried at the bottom** — **Absorbed.** Moved a condensed version of the refusal-attribution disclosure into the printed header, right after the summary line and before the column header row, so a table-skimmer sees it without reading to the closing Interpretation paragraph (which still carries the full explanation).

- **[minor, 1/3] `di-forced` mischaracterized as native-adjacent** — **Verified in code, absorbed.** Checked `decomposition_search.rs`'s gating explicitly:
  - `try_cells` (line ~1081) and `try_horizontal` (line ~1126) both require `opts.direct_insertion != DirectInsertion::Forced`; `try_di` (line ~1101) requires `opts.direct_insertion == DirectInsertion::Candidate`. All three read the OUTER `opts` passed into `select_best_decomposition`.
  - Under the harness's `di-forced` variant, that outer value IS `Forced` — so cells, horizontal-stack, and the internal DI-candidate arm are ALL stood down; only `NativeCandidate` runs, with DI baked directly into the native path (per `DirectInsertion::Forced`'s doc comment: "DI applied directly in the native path").
  - Under `default` (`Candidate`), the internal `DirectInsertionCandidate::produce` DOES use `Forced` — but only on a **local clone** (`di_opts`, `decomposition_search.rs` ~183) to build ONE competing candidate, which still runs alongside cells/horizontal (all gated on the unchanged outer `Candidate` value) and goes through extra validation/`DiClaimOrder` search the harness's top-level `di-forced` override never exercises.
  - So the harness's `di-forced` shape is NOT reproduced by the default search's own candidate competition — it's a genuinely user-elected/forced topology, correctly distinct from `cells-off`/`di-off`/`hs-off` (confirmed still native-adjacent: `DirectInsertion::Off`'s doc comment states it is literally "what NativeCandidate effectively runs under Candidate," and `try_horizontal`'s gate confirms native's baseline is always vertical). Footer corrected: only `di-forced` moved out of the native-adjacent group; `partitioned` remains the separate top-level strategy as established in round 2.

- **[minor, 1/3] `err_variants_disp` branch-type divergence, citing `balancer_lane_audit.rs:154-157`** — **Partially refuted.** The cited file DOES exist (`crates/core/tests/balancer_lane_audit.rs`) and lines 154-157 do contain the described pattern (`"—".to_string()` vs `.join(", ")`, both typed `String`) — so the citation itself is accurate, not hallucinated. But that doesn't establish a bug in our file: `check_firing_census.rs`'s `if err_variants_str.is_empty() { "-" } else { &err_variants_str }` has been compiled successfully in this exact form across all three rounds' `cargo test` runs (most recently this run, finishing green). Rust unifies the `&'static str` and `&String` arms via deref coercion to `&str` at this call site; this is a different but equally valid idiom from `balancer_lane_audit.rs`'s "both arms `String`" pattern, not a defect. No code change.

- **[minor, 1/3] "posenet is #[ignore]d"** — **Refuted-as-disclosed.** No identifier resembling "posenet" appears anywhere in this file or its module doc; the name is not something this instrument uses. The underlying substance — the census is `#[ignore]`d and not CI-tracked — is the instrument's stated, documented nature (module doc + the `Run:` line both say it's a manual, run-on-demand diagnostic). No CI hook added, per instruction.

- **[minor, 1/3] Fixture set uneven/slow** — **Refuted-as-disclosed.** Same as round 2: the module doc already states "Fixture list is a hardcoded tier-ladder slice ... extend as needed."

## Fresh census (this run: 36 builds, 0 refusals, 0 SKIPs)

```
=== check-firing census: 36 builds, refusals: none ===
(a category ABSENT below was never observed on a built layout here — indistinguishable from 'only fires on refused candidates'; see Interpretation at the bottom)
category                          winner loser-only err-loser  count  err-variants              variants
belt-flow-reachability               yes          -         -    240  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
input-rate-delivery                  yes        YES         -    179  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-flow-path                       yes          -         -     60  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-detour                          yes        YES         -     33  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-dead-end                        yes          -         -     18  cells-off,default,di-forced,di-off,hs-off,partitioned  cells-off,default,di-forced,di-off,hs-off,partitioned
row-input-belt-margin                yes          -         -     18  -                         cells-off,default,di-forced,di-off,hs-off,partitioned

Interpretation: `loser-only`/`err-loser` are computed per fixture, not from the aggregate `winner` flag (round 3, #686): for a fixture whose default build actually succeeded, did a non-default variant of THAT fixture fire (any severity for loser-only, Severity::Error for err-loser) while that fixture's own default did not — unioned across fixtures. A fixture whose default REFUSED contributes nothing to either predicate, because a refused default has no result to have stayed quiet: its absence must not be read as 'clean'. Of the option-toggle variants, cells-off/di-off/hs-off are shapes the native Candidate-mode search already evaluates internally as its own baseline — `decomposition_search.rs`'s NativeCandidate always runs DI-free/cells-free/vertical under the default `Candidate` settings (see the `DirectInsertion::Off` and `try_horizontal` doc comments), so firings confined to those three are native-adjacent, not outside the search. `di-forced` is different: setting `direct_insertion` to `Forced` at the top level (rather than letting `Candidate` mode use `Forced` internally to build ONE competing candidate) stands down the cells/horizontal/DI-candidate arms entirely (`decomposition_search.rs`'s `try_cells`/`try_horizontal`/`try_di` gates all require the outer mode to be non-Forced or exactly Candidate) and bakes DI directly via a path the default search's own DI-candidate arm does not reproduce — so `di-forced` is a user-elected/forced topology, not a shape the native search tries on its own. `partitioned` (LayoutStrategy::PartitionedDecomposed) remains a genuinely separate, user-elected top-level strategy the native search never runs. Refused builds (Err from build_bus_layout) produce no layout to validate, so no category attribution is possible for them — 'fired on NOTHING evaluated here' cannot distinguish a genuinely inert category from one that only ever appears inside refused candidates (v1 scope caveats in the module doc apply before concluding a quiet category is inert).
```

**Note on major 1's hazard vs. major 2's:** this run's 36/36 builds with 0 refusals means major 1's specific failure mode (a refused default masquerading as a clean one) never actually triggered here — the `defaults_built` guard is inert on this data and only pays off on a future run with a refusal. Major 2's cross-fixture masking, by contrast, DID trigger on this data: `input-rate-delivery` and `belt-detour` flip from `loser-only=-` (rounds 1-2, wrongly) to `loser-only=YES` (round 3, correctly) with no other change to the underlying builds — direct confirmation the bug was live, not theoretical.

New head SHA: `1d3886a4d835915bbb7f296c03242d1b3f73dcfc`

storkme:
Round 4 absorbed. New head: `5738f3f8ee86949e90937ad3d217077e501d516a`.

## Findings

- **[blocker] `entries.sort_by_key(|(name, _)| **name)` does not compile (E0277)** — **Refuted. The blocker is false.** Verified three independent ways:
  1. A fresh `cargo check --manifest-path crates/core/Cargo.toml --test check_firing_census` against the exact committed line, run just now — succeeds cleanly, no diagnostics.
  2. The full `cargo test --test check_firing_census -- --ignored --nocapture` run for round 3's SHA (`1d3886a4`, this round's base) built and ran to completion, printing the table — a compile failure would have made that impossible.
  3. `rust` and `rust-clippy` CI checks are green on `1d3886a4` (`gh api repos/storkme/spaghettio/commits/1d3886a4.../check-runs`).

  The review's static reasoning has an off-by-one reference-count error: `refusals_by_variant: FxHashMap<&'static str, usize>` — the key type is **already** `&'static str`, so `.iter()` yields `(&&'static str, &usize)`, not `(&'static str, &usize)` as the review assumes. `entries: Vec<(&&str, &usize)>`. `sort_by_key`'s closure receives `&T`; match ergonomics on that outer reference adds one more layer to the tuple-pattern binding, so `name: &&&str` (three reference layers over `str`), not `&&str` (two) as both review passes claim. `*name` peels to `&&str`, and `**name` peels again to `&str` — a valid, `Sized`, `Ord` type. It never reaches the bare unsized `str` the review describes. This is not something a code change can affect (the reasoning is simply wrong about the type), so no code change here.

  This matches this repo's documented history of the review bot posting confident false compile-failure claims on green-CI SHAs (tracked in this session's operator notes as `feedback_review_bot_confident_hallucinations.md`: "2/2-pass 'cannot compile' claims on green-CI SHAs; refute with receipts, don't patch to appease"). `docs/review-bot.md`'s failure-class table (classes 1-8) currently only covers silent-no-post and can't-be-closed failure modes, not false-content-within-a-posted-review — this instance (confident, unanimous-among-passing-votes claim of a hard build error that three independent checks refute) is worth adding as a new class there; flagging for the owner rather than editing that doc unilaterally mid-review-cycle.

- **[major] fixture key omits machine tier** — **Absorbed.** `fixture_key` is now `format!("{item}@{rate}:{machine}")`. Current six fixtures don't collide either way (confirmed: the printed table is unchanged, see below), so this is a pure robustness hardening against the module doc's own invitation to extend the fixture list with e.g. a second tier at an existing item/rate.

- **[minor] table columns collapse severity/variant distinctions (`winner` weak, `count` mixes severities/variants)** — **Absorbed the cheap part only.** Added one legend line: "(count sums every issue across all variants and both severities for that category — it is not per-variant or per-severity; use err-variants/variants for provenance)". Did not restructure the table (no per-severity/per-variant count split, no re-sort by loser/err-loser first) — that's a materially bigger change than "cheap," and the existing `err-variants`/`variants` columns already carry the provenance a reader needs to not misread `count`.

## Observation (not a finding response)

The review comment's raw body includes a leaked internal instruction block — a literal `=== OUTPUT FORMAT ===` section with "Please produce the merged comment for the author... No comment text repeating format instructions. No preamble or postamble..." followed by what reads as a second, reformatted pass of the same findings. This looks like a review-bot-side prompt/formatting glitch (the merge step's own system instructions leaking into the rendered comment body) rather than anything about this PR's code. Flagging for the review-bot failure-class log; not something we can fix from this branch.

## Fresh census (this run: 36 builds, 0 refusals, 0 SKIPs)

Identical to round 3's table except for the new second legend line (item 3 above) — confirming the fixture-key change (item 2) didn't alter any output on the current fixture set.

```
=== check-firing census: 36 builds, refusals: none ===
(a category ABSENT below was never observed on a built layout here — indistinguishable from 'only fires on refused candidates'; see Interpretation at the bottom)
(count sums every issue across all variants and both severities for that category — it is not per-variant or per-severity; use err-variants/variants for provenance)
category                          winner loser-only err-loser  count  err-variants              variants
belt-flow-reachability               yes          -         -    240  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
input-rate-delivery                  yes        YES         -    179  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-flow-path                       yes          -         -     60  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-detour                          yes        YES         -     33  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-dead-end                        yes          -         -     18  cells-off,default,di-forced,di-off,hs-off,partitioned  cells-off,default,di-forced,di-off,hs-off,partitioned
row-input-belt-margin                yes          -         -     18  -                         cells-off,default,di-forced,di-off,hs-off,partitioned

Interpretation: `loser-only`/`err-loser` are computed per fixture, not from the aggregate `winner` flag (round 3, #686): for a fixture whose default build actually succeeded, did a non-default variant of THAT fixture fire (any severity for loser-only, Severity::Error for err-loser) while that fixture's own default did not — unioned across fixtures. A fixture whose default REFUSED contributes nothing to either predicate, because a refused default has no result to have stayed quiet: its absence must not be read as 'clean'. Of the option-toggle variants, cells-off/di-off/hs-off are shapes the native Candidate-mode search already evaluates internally as its own baseline — `decomposition_search.rs`'s NativeCandidate always runs DI-free/cells-free/vertical under the default `Candidate` settings (see the `DirectInsertion::Off` and `try_horizontal` doc comments), so firings confined to those three are native-adjacent, not outside the search. `di-forced` is different: setting `direct_insertion` to `Forced` at the top level (rather than letting `Candidate` mode use `Forced` internally to build ONE competing candidate) stands down the cells/horizontal/DI-candidate arms entirely (`decomposition_search.rs`'s `try_cells`/`try_horizontal`/`try_di` gates all require the outer mode to be non-Forced or exactly Candidate) and bakes DI directly via a path the default search's own DI-candidate arm does not reproduce — so `di-forced` is a user-elected/forced topology, not a shape the native search tries on its own. `partitioned` (LayoutStrategy::PartitionedDecomposed) remains a genuinely separate, user-elected top-level strategy the native search never runs. Refused builds (Err from build_bus_layout) produce no layout to validate, so no category attribution is possible for them — 'fired on NOTHING evaluated here' cannot distinguish a genuinely inert category from one that only ever appears inside refused candidates (v1 scope caveats in the module doc apply before concluding a quiet category is inert).
```

New head SHA: `5738f3f8ee86949e90937ad3d217077e501d516a`

storkme:
Round 5 (final absorption round) complete. New head: `f50c85280dc55694d3fcbd8e8ff780a265a01a1e`.

## Findings

- **[major, 3/3] "the census only validates shipped winners, cannot answer its own title question"** — **Refuted-as-disclosed. This has now been adjudicated three times; recycled, not new.** Round 1 refuted this exact finding as the module doc's stated v1 scope: "the candidate field is approximated by the public option-toggle family (native / DI-off / cells-off / HS-off / DI-forced / partitioned), not by the search's internal candidate list ... If selection-unification lands (the audit's refactor #1), this census should move onto the real candidate loop and drop the approximation" (module doc, top of file). Round 3 moved a condensed version of the same disclosure into the printed table header itself: `"(a category ABSENT below was never observed on a built layout here — indistinguishable from 'only fires on refused candidates'; see Interpretation at the bottom)"` — this run's output (below) confirms that line is still printed, first thing after the summary. See the round-1 (https://github.com/storkme/spaghettio/pull/686#issuecomment-5363724973) and round-3 (https://github.com/storkme/spaghettio/pull/686#issuecomment-5363951276) adjudication comments for the full prior treatment. No new code change.

- **[major, 2/3] `loser_only`/`err_loser` conflate user-elected topologies (`di-forced`/`partitioned`) with evaluated-loser firings** — **Absorbed — this was the one real residual.** Round 3's own code verification (receipts: `decomposition_search.rs`'s `try_cells`/`try_horizontal`/`try_di` gates, `DirectInsertion::Off`'s doc comment) already established that `di-off`/`cells-off`/`hs-off` are shapes the native `Candidate`-mode search evaluates as its own baseline arms, while `di-forced`/`partitioned` are genuinely separate, user-elected topologies the default search never tries. Round 5 finishes applying that distinction to the flag predicates: added `const NATIVE_ADJACENT_VARIANTS: &[&str] = &["di-off", "cells-off", "hs-off"]`, and `fixtures_any_nondefault`/`fixtures_err_nondefault` now only accumulate fixture keys from those three variants — `di-forced`/`partitioned` firings still land in `variants`/`err_variants`/`count` (visible, for coverage) but no longer set `loser-only`/`err-loser`. Added a legend line stating the restriction. **Flag changes this run: none.** Reran the census under the pinned zone cache — `input-rate-delivery` and `belt-detour` still read `loser-only=YES`, identical to round 4's table; both flags were already substantiated by native-adjacent (di-off/cells-off/hs-off) firings on this fixture set, not solely by di-forced/partitioned, so this round is a correctness hardening with no visible effect on the current data, not a behavior change.

- **[minor] "count sort ranks by winner-side noise"** — **Refuted-as-scoped.** The flags (`loser-only`/`err-loser`) are the signal; `count` is context, explicitly labeled as such by the round-4 legend line ("count sums every issue across all variants and both severities ... it is not per-variant or per-severity"). Round 4's instruction was no further table restructuring; re-sorting by flags-first is exactly that restructuring, scoped out.

- **[minor] "`defaults_built` names a build path, not a validate path"** — **Refuted, false on the code.** Read the loop: `build_bus_layout` returns `Ok(l)` → `builds += 1` → (if `vname == "default"`) `defaults_built.insert(fixture_key.clone())` → **immediately followed, in the same loop iteration, before moving to the next variant**, by `validate::validate(&l, Some(&sr))` and the issue-recording loop that populates every category's `fixtures_any_default`/`fixtures_err_default` entries. There is no separate "build-only, never validated" path — every layout that reaches `defaults_built` is validated in that same iteration, which is where ALL of a category's rows (including `winner`/`fixtures_any_default`) originate. A validation-only category on a built default is not a gap this file has; the claim doesn't match the code. No change.

- **[minor] "footer says search-space-wide"** — **Checked the actual text; refuted by quotation, no such framing exists.** The Interpretation paragraph (unchanged in this respect since round 3) says: *"`loser-only`/`err-loser` are computed per fixture, not from the aggregate `winner` flag ... for a fixture whose default build actually succeeded, did a non-default variant of THAT fixture fire ... while that fixture's own default did not — unioned across fixtures."* That is already exactly "per-fixture paired comparison, unioned across fixtures" — the wording the review says would need tightening. No "search-space-wide" or equivalent phrase appears anywhere in the file. No change needed.

## Fresh census (this run: 36 builds, 0 refusals, 0 SKIPs)

```
=== check-firing census: 36 builds, refusals: none ===
(a category ABSENT below was never observed on a built layout here — indistinguishable from 'only fires on refused candidates'; see Interpretation at the bottom)
(count sums every issue across all variants and both severities for that category — it is not per-variant or per-severity; use err-variants/variants for provenance)
(loser-only/err-loser flags consider native-adjacent variants only: di-off, cells-off, hs-off. User-elected topologies (di-forced, partitioned) still show up in err-variants/variants and count, but are not counted as selection evidence)
category                          winner loser-only err-loser  count  err-variants              variants
belt-flow-reachability               yes          -         -    240  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
input-rate-delivery                  yes        YES         -    179  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-flow-path                       yes          -         -     60  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-detour                          yes        YES         -     33  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-dead-end                        yes          -         -     18  cells-off,default,di-forced,di-off,hs-off,partitioned  cells-off,default,di-forced,di-off,hs-off,partitioned
row-input-belt-margin                yes          -         -     18  -                         cells-off,default,di-forced,di-off,hs-off,partitioned

Interpretation: `loser-only`/`err-loser` are computed per fixture, not from the aggregate `winner` flag (round 3, #686): for a fixture whose default build actually succeeded, did a non-default variant of THAT fixture fire (any severity for loser-only, Severity::Error for err-loser) while that fixture's own default did not — unioned across fixtures. A fixture whose default REFUSED contributes nothing to either predicate, because a refused default has no result to have stayed quiet: its absence must not be read as 'clean'. Of the option-toggle variants, cells-off/di-off/hs-off are shapes the native Candidate-mode search already evaluates internally as its own baseline — `decomposition_search.rs`'s NativeCandidate always runs DI-free/cells-free/vertical under the default `Candidate` settings (see the `DirectInsertion::Off` and `try_horizontal` doc comments), so firings confined to those three are native-adjacent, not outside the search. `di-forced` is different: setting `direct_insertion` to `Forced` at the top level (rather than letting `Candidate` mode use `Forced` internally to build ONE competing candidate) stands down the cells/horizontal/DI-candidate arms entirely (`decomposition_search.rs`'s `try_cells`/`try_horizontal`/`try_di` gates all require the outer mode to be non-Forced or exactly Candidate) and bakes DI directly via a path the default search's own DI-candidate arm does not reproduce — so `di-forced` is a user-elected/forced topology, not a shape the native search tries on its own. `partitioned` (LayoutStrategy::PartitionedDecomposed) remains a genuinely separate, user-elected top-level strategy the native search never runs. Since round 5, `loser-only`/`err-loser` are computed ONLY from di-off/cells-off/hs-off firings — `di-forced`/`partitioned` firings still show up in err-variants/variants/count but no longer set either flag, because the search never evaluated those shapes to refuse in the first place. Refused builds (Err from build_bus_layout) produce no layout to validate, so no category attribution is possible for them — 'fired on NOTHING evaluated here' cannot distinguish a genuinely inert category from one that only ever appears inside refused candidates (v1 scope caveats in the module doc apply before concluding a quiet category is inert).
```

New head SHA: `f50c85280dc55694d3fcbd8e8ff780a265a01a1e`

This closes five absorption rounds on #686. Summary of the flag columns' final state: `loser-only`/`err-loser` are per-fixture, severity-appropriate, gated on the default having actually built, and restricted to variants the native search genuinely evaluates. Remaining known limitations (candidate-loop approximation, refusal-path attribution) are disclosed in the module doc and in the printed table header, not silently assumed away.

storkme:
Round 6 (final absorption round) complete. New head: `8dae746bf42803f5377d0ec5064aafacc0c83d1f`.

## Findings

- **[major, 3/3] `loser-only`/`err-loser` can't distinguish 1/6 fixtures from 6/6 (docs/validator-reporting.md's "cannot tell 2 from 218" rule)** — **Absorbed.** This was real: rounds 2-3 fixed invisible cross-fixture masking by keying the internal sets on fixture identity, but the two printed flags then re-collapsed those per-fixture sets back into one unioned boolean, moving the exact same problem up a level. Rather than print raw fixture keys per flag (noisy for a 6-fixture slice, and doesn't scale), absorbed via a **mechanical no-op detector**: `layout_signature()` reduces a built layout to a sorted `(name, x, y, direction, recipe)` tuple per entity; each non-default variant's signature is compared against that SAME fixture's own default signature, and a new "per-variant no-op check" block prints, per variant, how many of the 6 fixtures were bit-identical to default. This gives the cardinality/identity information the rule demands, at the variant level rather than duplicating fixture keys inline in the category table.

- **[major, 2/3] `cells-off`/`hs-off` may be no-op arms for this fixture slice** — **Verified empirically, absorbed.** Checked the gates in code first: `try_cells` (`decomposition_search.rs` ~1076-1086) requires `cell_composition == Candidate` (true under default), `direct_insertion != Forced` (true under default), `max_belt_tier.is_none_or(|t| t == "express-transport-belt")` — note `is_none_or` returns **true** when `max_belt_tier` is `None`, so NOT setting it (as all 6 fixtures do) satisfies this specific condition rather than failing it, contrary to what a literal reading of the reviewer's stated mechanism suggests. The real fixture-dependent gate is `chain_eligible(solver_result)`. `try_horizontal` requires `any_dual_input_row(&solver_result.machines)` (`placer.rs:760`), also fixture-dependent. Rather than resolve this by arguing about gate predicates in the abstract, I ran the mechanical detector and let the data answer it directly:

  ```
  di-off       5/6 builds identical to default  (native-adjacent)
  di-forced    5/6 builds identical to default  (user-elected)
  cells-off    6/6 builds identical to default  (native-adjacent)
  hs-off       5/6 builds identical to default  (native-adjacent)
  partitioned  4/6 builds identical to default  (user-elected)
  ```

  **`cells-off` IS identical to default on all 6 current fixtures** — confirmed plainly, as asked. `di-off`/`di-forced`/`hs-off` each differ from default on exactly 1 of 6 fixtures; `partitioned` differs on 2 of 6. Net effect: this fixture slice currently measures almost entirely the DI channel — cell-composition and horizontal-stack contribute essentially nothing to the flags on this data. This is a known v1-scope consequence (the module doc already says the fixture list is "a hardcoded tier-ladder slice... extend as needed") now made **visible instead of silent**, exactly the fix's intent. Added the footer sentence: "A firing on a variant identical to default carries no selection evidence, and a variant that is no-op across all fixtures measures nothing here."

- **[minor, 1/3] `NATIVE_ADJACENT_VARIANTS` can desync from the variants table** — **Absorbed.** Removed the separately-declared const entirely; the `native_adjacent: bool` now lives in the `variants` tuple itself (`(&str, fn(&mut LayoutOptions), bool)`), and both the loop's flag-gating and the printed legend's variant-name list are derived from that single table. No second list exists to desync from.

## Refuted (receipts, no code change)

- **"`#[ignore]`d + no assertions = silent-wrong census"** — refuted-as-disclosed, recycled from rounds 1 and 5. See those adjudications (https://github.com/storkme/spaghettio/pull/686#issuecomment-5363724973, https://github.com/storkme/spaghettio/pull/686#issuecomment-5364193710) for the full treatment — this is the instrument's documented, deliberate nature (manual diagnostic, `#[ignore]`d by design), not a new finding.

- **"Error-laden shipped default treated as clean fuels the flags"** — **Verified, and it's backwards, exactly as anticipated.** Traced the code: for the "default" variant, `fixtures_any_default`/`fixtures_err_default` are populated **per category** whenever that specific category fires on the default build (any severity / Error respectively) — independent of `defaults_built`, which only gates "did the build succeed." So if category C itself Error-fires on the shipped default (the point-1 scenario, where `select_best_decomposition` ships an error-laden candidate when nothing validates clean), that fixture key lands in **C's own** `fixtures_err_default`, which makes `err_loser()`'s `!fixtures_err_default.contains(f)` **false** for that fixture — **suppressing** `err-loser` for category C on that fixture, not fueling it. This is the conservative direction: a category that's visibly wrong on the shipped default isn't also flagged as "invisibly hidden," because it plainly isn't hidden. An unrelated category D firing only on a non-default variant is unaffected by C's Error on default — D's own `fixtures_err_default` set is empty for that fixture regardless of what C did, so D's `err_loser` reads correctly off D's own presence/absence. No fueling in either direction; refuted with the code path above.

- **"count-sort ranks by noise"** — refuted-as-scoped, recycled from round 5; see that adjudication.

- **Two compile-dialect nits (`sort_by` binding style, `err_variants_disp` branch types)** — refuted with receipts: the file compiles. This round alone: a fresh `cargo check --test check_firing_census` (clean), a fresh `cargo clippy --test check_firing_census -- -D warnings` (clean, after fixing two `type_complexity` hits and one `print_literal` hit that a stricter `--test`-scoped clippy invocation caught — the repo's pre-commit hook runs `cargo clippy -p spaghettio_core -- -D warnings` **without `--tests`**, so it never actually lints files under `crates/core/tests/`; I ran the stricter check anyway and fixed what it found), and the full `cargo test --ignored --nocapture` run below, green.

- **`catch_unwind` for `di-forced`/`partitioned`** — **agree with refute-as-scoped, independently.** This file calls `build_bus_layout` directly, unwrapped, unlike the internal candidate arms' `run_candidate_catch_unwind`. For a hand-run, `#[ignore]`d diagnostic, a loud panic on a genuine engine bug is the right failure mode: wrapping it would produce a table that *looks* complete but silently dropped a variant's data, which is worse for an instrument whose whole purpose is making silent gaps visible. An operator who hits a panic here can see exactly which variant/fixture crashed and rerun without it. Not absorbing.

## Fresh census (this run: 36 builds, 0 refusals, 0 SKIPs)

```
=== check-firing census: 36 builds, refusals: none ===
(a category ABSENT below was never observed on a built layout here — indistinguishable from 'only fires on refused candidates'; see Interpretation at the bottom)
(count sums every issue across all variants and both severities for that category — it is not per-variant or per-severity; use err-variants/variants for provenance)
(loser-only/err-loser flags consider native-adjacent variants only: di-off, cells-off, hs-off. User-elected topologies still show up in err-variants/variants and count, but are not counted as selection evidence)
category                          winner loser-only err-loser  count  err-variants              variants
belt-flow-reachability               yes          -         -    240  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
input-rate-delivery                  yes        YES         -    179  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-flow-path                       yes          -         -     60  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-detour                          yes        YES         -     33  -                         cells-off,default,di-forced,di-off,hs-off,partitioned
belt-dead-end                        yes          -         -     18  cells-off,default,di-forced,di-off,hs-off,partitioned  cells-off,default,di-forced,di-off,hs-off,partitioned
row-input-belt-margin                yes          -         -     18  -                         cells-off,default,di-forced,di-off,hs-off,partitioned

=== per-variant no-op check (bit-identical entities vs that fixture's own default; round 6, #686) ===
  di-off       5/6 builds identical to default  (native-adjacent)
  di-forced    5/6 builds identical to default  (user-elected)
  cells-off    6/6 builds identical to default  (native-adjacent)
  hs-off       5/6 builds identical to default  (native-adjacent)
  partitioned  4/6 builds identical to default  (user-elected)

Interpretation: `loser-only`/`err-loser` are computed per fixture, not from the aggregate `winner` flag (round 3, #686): for a fixture whose default build actually succeeded, did a non-default variant of THAT fixture fire (any severity for loser-only, Severity::Error for err-loser) while that fixture's own default did not — unioned across fixtures. A fixture whose default REFUSED contributes nothing to either predicate, because a refused default has no result to have stayed quiet: its absence must not be read as 'clean'. Of the option-toggle variants, cells-off/di-off/hs-off are shapes the native Candidate-mode search already evaluates internally as its own baseline — `decomposition_search.rs`'s NativeCandidate always runs DI-free/cells-free/vertical under the default `Candidate` settings (see the `DirectInsertion::Off` and `try_horizontal` doc comments), so firings confined to those three are native-adjacent, not outside the search. `di-forced` is different: setting `direct_insertion` to `Forced` at the top level (rather than letting `Candidate` mode use `Forced` internally to build ONE competing candidate) stands down the cells/horizontal/DI-candidate arms entirely (`decomposition_search.rs`'s `try_cells`/`try_horizontal`/`try_di` gates all require the outer mode to be non-Forced or exactly Candidate) and bakes DI directly via a path the default search's own DI-candidate arm does not reproduce — so `di-forced` is a user-elected/forced topology, not a shape the native search tries on its own. `partitioned` (LayoutStrategy::PartitionedDecomposed) remains a genuinely separate, user-elected top-level strategy the native search never runs. Since round 5, `loser-only`/`err-loser` are computed ONLY from di-off/cells-off/hs-off firings — `di-forced`/`partitioned` firings still show up in err-variants/variants/count but no longer set either flag, because the search never evaluated those shapes to refuse in the first place. Refused builds (Err from build_bus_layout) produce no layout to validate, so no category attribution is possible for them — 'fired on NOTHING evaluated here' cannot distinguish a genuinely inert category from one that only ever appears inside refused candidates (v1 scope caveats in the module doc apply before concluding a quiet category is inert). A firing on a variant identical to default carries no selection evidence, and a variant that is no-op across all fixtures measures nothing here — see the per-variant no-op check above.
```

New head SHA: `8dae746bf42803f5377d0ec5064aafacc0c83d1f`

This is the sixth and final absorption round per the coordinator's instruction; no further rounds from this side. All code changes are confined to `crates/core/tests/check_firing_census.rs`; the pinned zone cache (`crates/core/data/sat-zones-ci.bin`) never appeared in any commit across all six rounds.

storkme:
**Post-merge note on the round-7 review** (the PR merged on the required checks; this closes the advisory thread).

The major's kernel is valid and recorded as a follow-up on #675: `layout_signature` omits `carries`/`mirror`/`rate`, which the validator reads, so the no-op label is weaker than 'validates identically' — enriching the signature is cheap hardening for whoever next touches the census. Its proposed fix (gating the flag predicates on `!is_noop`) is the wrong wiring, though: a variant whose layout is *genuinely* identical to default validates identically, so its firings land on the default side of the same fixture's pairing and cannot set `loser_only`/`err_loser` — the pairing already neutralizes true no-ops; only the *label's* honesty needs the richer signature. The denominator minor (non-comparable builds counted in `noop/total`) is valid bookkeeping, same follow-up. The `loser-only`-as-red-herring minor restates the module doc's severity point; the header's err-loser column is the evidence column, as the Interpretation states.

No further rounds will be worked — findings on the merged head land here as advisory record.

PR #687, opened 2026-08-21
test(g1): pin the perp-template rung — attribution, pipe×belt fixture, reachability finding

storkme:
## Intent

G1 from #675: the ~795-line perpendicular-template routing rung — rung 1 of the production strategy ladder — had zero test coverage anywhere in the suite. This PR pins it, and reports what pinning it uncovered. What it uncovered, over three review rounds, is that the rung appears to be **production-unreachable on both shapes it handles** — the pins are unit pins of its internal logic plus regression locks on the ladder's real dispatch.

## Scope

- **Strategy attribution in the region-fixture harness** (`expected.solved_by`): asserts the winning strategy from the terminal `JunctionSolved` event; on mismatch prints the solution plus every strategy attempt (iter/variant/outcome/detail), `try_bridge` rejection reasons, and the candidate-cost comparison. Fails loud if set on a non-solve fixture, and on unknown `spec_kinds` values.
- **Finding 1 — the rung is production-unreachable on both its shapes.** Belt×belt two-item crossings: the item-conflict gate skips every strategy on the sole single-tile iteration, and the rung refuses grown regions. Pipe×belt crossings (round-3 review find, verified): `keys_at_tile` filters pipe specs out of junction seeding, so the rung's 2-spec predicate refuses and `bridge_belt_over_pipe` never runs from production — SAT handles pipes as obstacles instead. The only remaining reachability hypothesis is same-item belt crossings; the followups doc upgrades the follow-up candidate to a census-then-delete of the whole rung (or a gate rework so the cheap template can fire).
- **Finding 2 — the replay's strategy ladder had drifted from production** (pre-native `sat-1ug`/Relaxed-reach list). Fixed structurally: the pinned-tier core is lifted into a shared `ghost_router::pinned_tier_core_strategies()` used by both production and the replay (production behavior unchanged: same rungs, same order, auto-tier extras appended as before). Auto-tier extras stay outside the replay — fixtures don't record belt-tier mode — so `solved_by` pins are relative to the core, stated as a residual.
- `perp_template_single_tile_crossing.json` pins the core ladder's dispatch on the canonical belt×belt crossing (`sat-1ug-native`; `max_cost` 17 = measured 14 + varisat-wobble pad).
- `perp_template_pipe_belt_bridge.json` is a **static unit pin** of the rung's internal pipe×belt logic (`solved_by: "perpendicular_template"`, `max_cost: 12`, UG-pair `required_entities`) — hand-authored by necessity, since captures can never emit pipe `initial_specs`. Supporting schema: `spec_kinds` on `RegionFixture`, threaded through the capture dump.
- Directory README updated (`solved_by` scope + pin complementarity, `spec_kinds`, pipe-capture caveat); both findings + the census follow-up recorded in `docs/offpath-code-followups.md`.

## Verification actually run

- region_fixtures 8/8 under the pinned zone cache after every round of changes; all six pre-existing fixtures byte-identical outcomes (costs 33/28/26, three capped fixtures still cap) under the realigned/shared ladder.
- The attribution assert demonstrably discriminates: it caught the original fixture's wrong premise (sat rungs absorbing the solve), a degenerate `sat-surface` candidate stamping a belt onto a pipe tile in an underspecified draft of the pipe fixture, and the ladder drift — the failure diagnostics located all three.
- Full core suite (`--tests --no-fail-fast`, pinned cache) green with cargo exit 0 — this is also the behavior-parity check for the shared-helper refactor of production's strategy construction; clippy + wasm build clean.

## Deviations

The PR grew from "add one fixture" to "attribution + two fixtures + a shared-ladder refactor + two findings", driven by three review rounds whose majors were each correct — including twice falsifying the PR's own central claim (first "the fixture pins the rung", then "pipe×belt is the rung's live path"). The alternative was shipping pins of things the engine never does.


storkme:
**Review round absorbed** (both findings valid; the major's absorption falsified the fixture's own premise and reshaped the PR — details below).

**Major — "fixture does not pin the rung it claims" (3/3): confirmed, absorbed.** The harness now supports `expected.solved_by` strategy attribution: it asserts the *winning* strategy from the terminal `JunctionSolved` trace event (not merely any rung reporting a walker-valid attempt — speculative variants run their own ladders and losing candidates emit `Solved` too), and on mismatch prints the winning solution plus every `(strategy, outcome, iter/variant, detail)` attempt, `try_bridge` rejection reasons, and the `variant-chosen` cost comparison.

Running that attribution falsified the fixture's premise, which is the real payoff of this finding:

1. **The perp rung's belt×belt arm is structurally unreachable for two-item crossings** — the exact shape the original fixture (and the rung's own naming) assumes. `junction_solver`'s item-conflict gate skips every strategy on the sole single-tile iteration (both specs' boundary ports land on the crossing tile with different items), and the rung refuses any grown region (`tile_count > 1` guard). Each gate is individually correct; composed, they exclude the rung.
2. **The replay ladder had drifted from production** — it still built the pre-native `sat-1ug` (Relaxed reach) list, so any attribution pin would have named rungs production no longer runs. Realigned to production's pinned-tier core (`perp`, `sat-surface`, `sat-1ug-native`, `sat-2ug-native`, `sat-native`); all six pre-existing fixtures pass unchanged under the aligned ladder.

Resolution: the original fixture is re-pinned honestly (`solved_by: "sat-1ug-native"` — the ladder's actual dispatch on this shape, finding recorded in its notes), and a **new fixture** `perp_template_pipe_belt_bridge` pins the rung on its genuinely reachable path — a belt crossing a fluid-trunk pipe column (pipe items are conflict-gate-exempt, so the strategies do run at the single-tile iteration). It asserts `solved_by: "perpendicular_template"` and the UG-pair shape via `required_entities`. Getting it to reflect production also required teaching the fixture schema `spec_kinds` (replay previously hardcoded every spec to Belt) and threading kinds through the capture dump so future pipe captures replay faithfully. Both findings and the follow-up candidates (possible dead belt×belt arm pending a same-item-crossing census) are recorded in `docs/offpath-code-followups.md`.

**Minor — missing `max_cost`/`required_entities` ratchets (3/3): absorbed.** Original fixture: `max_cost: 14` (pins the deterministic pinned-cache cost exactly). New fixture: `max_cost: 12` (measured 10 + ~20% starting pad per the README), plus `required_entities` pinning the UG pair at (4,5)/(6,5). The original fixture deliberately omits `required_entities` — its job after the finding is dispatch attribution, not shape; `solved_by` + exact cost cover that.

**Nits:** hand-authored — kept deliberately (minimal synthetic shapes are the point; the capture workflow now records `spec_kinds` so captured pipe fixtures are faithful). EOF newlines present on both fixtures. Schema additions documented in the directory README.

Verification: full region_fixtures suite 8/8 under the pinned zone cache; full core test suite green (`--no-fail-fast`); clippy + wasm build clean. PR body updated to match the reshaped scope.

storkme:
**Round 2 adjudicated** (head `d98a9de3`; delta is docs/notes-only, so the trivial-delta gate should mark it rather than buy a re-review).

**Major — "legacy fixtures not re-validated under the ladder change" (3/3): refuted with receipts.** The re-validation the finding asks for was run before the reviewed commit was pushed (recorded in the PR body's verification section and the previous adjudication comment — the bot doesn't read those, so receipts again here):

- `region_fixtures` under the pinned zone cache after the ladder realignment: **8/8 PASS**, and every legacy outcome is *identical* to the pre-realignment run — `advanced_circuit_ret_plus_three_trunks` cost 33 (ratchet 65), `electronic_circuit_copper_cable_lost_at_chain_head` cost 28 (ratchet 40), `electronic_circuit_copper_cable_missing_input_chain` cost 26 (ratchet 35); all three `capped` fixtures still cap. No ratchet moved, no cap path flipped.
- Full core suite `--tests --no-fail-fast` under the pinned cache: 25/25 test binaries green, cargo exit 0.

The finding's mechanism is real (native reach is stricter than the drifted Relaxed reach) — it just didn't materialize on any committed fixture, which the run demonstrates rather than assumes.

**Minor — `solved_by` pins the rung name, not iteration/variant (2/3): absorbed.** README's `solved_by` section now states the pin's scope explicitly and that `solved_by` / `required_entities` / `max_cost` are complementary, not redundant.

**Minor — pipe fixture could pass trivially on a cost tie; complementarity should be explicit (2/3): absorbed.** Stated in the pipe fixture's own notes (cheaper-SAT flips `solved_by` while shape pins pass, and vice versa) and in the README.

**Minor — pipe spec carries belt-shaped fields (2/3): absorbed as documentation.** README schema comment now says pipe specs carry `spec_belt_tiers`/`spec_exit_dirs` as schema-required-but-ignored-by-pipe-handling. Not restructured: making those fields optional for pipes would fork the schema for no behavioral gain.

The "Verified Clean" section's four confirmations (ladder parity, `JunctionSolved` naming, `spec_kinds` plumbing, conflict-gate/short-circuit consistency) match our own verification — good corroboration.

storkme:
**Round 3 absorbed.** The 2/3 major is the round's real contribution — it caught this PR's own central claim overreaching, for the second time in the PR's life, and it's correct.

**Major — "production can never feed the perp rung a pipe×belt crossing" (2/3): CONFIRMED, absorbed.** Verified at the cited site: `keys_at_tile` filters `SpecKind::Pipe` out of junction seeding (with a comment explaining why — pipes are forbidden tiles; SAT bypasses them as obstacles), so a live pipe×belt crossing reaches the solver as one belt spec and the rung's 2-spec predicate refuses. Combined with the belt×belt conflict-gate finding, **the rung is production-unreachable on both shapes it handles**; the only remaining reachability hypothesis is same-item belt crossings. All wording corrected: the pipe fixture's notes now open with "STATIC UNIT PIN of the rung's internal pipe×belt logic — NOT a production-reachability probe" (and why it's still worth keeping: several hundred otherwise-untested lines, plus the ready-made lock if dispatch ever re-admits pipes); the followups doc entry now upgrades the follow-up candidate from "delete the belt×belt arm" to "census same-item crossings; if none, the entire ~795-line rung is production-dead and deletable — or rework the gates so the cheap template can actually fire."

**Major — unknown `spec_kinds` strings silently degrade to Belt (1/3): absorbed.** The replay now panics on any value other than `"Pipe"` (omit for Belt), naming the fixture and key — a typo fails loud instead of pinning the wrong geometry.

**Minor — capture workflow dead-on-arrival for pipe fixtures (1/3): confirmed (follows from the major), absorbed as documentation.** README capture section now states a capture can never emit a pipe `initial_specs` entry and pipe fixtures must be hand-authored; `spec_kinds` in the dump is forward-compat. Changing production dispatch to re-admit pipes is deliberately NOT done here — that's the follow-up's decision, with the census in hand.

**Minor — `solved_by` silently ignored on capped/unsatisfiable fixtures (1/3): absorbed.** The harness now fails any non-solve fixture that sets `solved_by`.

**Minor — hand-mirrored ladder can re-drift (1/3): absorbed structurally.** The pinned-tier core is lifted into `ghost_router::pinned_tier_core_strategies()`, now the single source used by production (`route_bus_ghost`, auto-tier extras appended as before) and the replay. Production behavior is unchanged (same five rungs, same order); the full suite run below is the parity check. The replay-only `perpendicular_template_strategy()` factory is gone with it.

**Minor — varisat nondeterminism vs the exact cost pin (1/3): absorbed.** `max_cost` on the belt fixture padded 14 → 17 (~20%, same precedent as `electronic_circuit_copper_cable_missing_input_chain`), with the reasoning in the fixture notes. The `solved_by` pin stays exact: per-rung SAT/UNSAT is deterministic for a fixed formula — only the solution's cost wobbles, so the attribution cannot flip from nondeterminism alone.

**Minor — "the ladder's ACTUAL behavior" overstates (auto-tier extras excluded) (1/3): absorbed.** Reworded to "the pinned-tier core ladder's dispatch" in the fixture notes; the auto-tier scope caveat (a change confined to auto-mode dispatch is not caught by these fixtures) is now stated in the followups entry as a residual, and was already in the README.

**Minor — pipe fixture proves code-reachability, not production-reachability (1/3): absorbed** — folded into the rewritten pipe-fixture notes (see the major).

**Nit — legacy ratchets need revalidation under the native ladder: refuted with receipts, again.** region_fixtures 8/8 after every change in this round, legacy outcomes identical (33/28/26, all three capped fixtures still cap); additionally the `rust` CI check on the two previous heads ran the full suite cold-cache and passed — an independent, different-machine confirmation.

Verification for this head: region_fixtures 8/8 (pinned cache); full core suite `--tests --no-fail-fast` green with cargo exit 0; clippy clean on the changed files; wasm build clean.

storkme:
**Round 4 adjudicated** — two valid harness minors absorbed; the majors re-litigate concessions already in the tree or rest on a premise the PR's own finding precludes.

**Major — "docs overstate: the perp pin is only for an unreachable path" (1/3): the concession is already in the tree.** Round 3 re-scoped every claim this finding objects to: the pipe fixture's notes open with "STATIC UNIT PIN … NOT a production-reachability probe", the README capture section states pipes can never be captured, and the followups entry's headline IS the unreachability finding, with the census-then-delete follow-up this finding asks us to commit to. Absorbed the one residual hook: the followups entry's opening now reads "cover the rung's logic and the ladder's dispatch" instead of "pin the rung". The dichotomy offered ("commit to deletion or re-scope") is answered: re-scoped in round 3, and deletion is deliberately gated on the same-item-crossing census — deleting ~795 lines on an inference without the census would be exactly the overreach this PR spent three rounds removing.

**Major — "solved_by is unstable: perp and sat-1ug-native produce identical cost-10 geometry, tie broken by push order" (1/3): refuted — the premise contradicts the PR's central finding.** On `perp_template_single_tile_crossing`, the perp rung **never produces a candidate**: the item-conflict gate skips all strategies on the sole single-tile iteration, and the rung returns Unsatisfiable on every grown region (the diagnostic attempts trace in the harness output shows `perpendicular_template → Unsatisfiable` at every iteration/variant; the winner selection saw five `sat-1ug-native` candidates, all cost 14, zero perp candidates). There is no perp/SAT tie to flip. The finding's numbers are also wrong on inspection: winner cost is 14 (not 10) and `max_cost` is 17 (not "45/40"). The absence of `required_entities` on this fixture is deliberate and was adjudicated in round 2: it is a dispatch pin; shape pins live on the pipe fixture.

**Moderate — "legacy ratchets not re-measured after the ladder swap": refuted for the third time, same receipts.** region_fixtures has been re-run after every change in rounds 2–4: 8/8, legacy outcomes identical each time (costs 33/28/26 vs ratchets 65/40/35; all three capped fixtures still cap). The `rust` CI check additionally ran the suite cold-cache on two prior heads, green. See the round-2 and round-3 adjudication comments for the same receipts.

**Minor — unknown `spec_kinds` panic aborts the whole accumulate-and-report run (2/3): absorbed.** The harness now validates `spec_kinds` per fixture before replay and pushes a per-fixture failure (run continues); `replay_region_fixture`'s panic remains as a backstop for non-harness callers but is unreachable from the harness.

**Minor — a `spec_kinds` entry naming a nonexistent spec is silently dropped (2/3): absorbed.** Same harness validation: every `spec_kinds` key must exist in `spec_items`, else a per-fixture failure names the dangling key.

Verification: region_fixtures 8/8 under the pinned cache after the changes, and the discrimination check was executed, not inspected: with the pipe fixture's kind temporarily corrupted to `"pipe"`, the run reports `7 passed, 1 failed` with the named per-fixture message (`spec_kinds["trunk:water:0"] = "pipe" — only "Pipe" is valid`) — the run continues past the bad fixture, which is the exact behavior the finding asked for.

storkme:
**Round 5 adjudicated** — two small harness hardenings absorbed; both majors rest on the same false mechanical premise and are refuted with the ladder's own trace as receipt. This closes the absorption cycle for this PR: the review rounds are now recycling previously-adjudicated claims (the legacy-ratchet item below is on its fourth appearance), so after this round's gate the PR merges.

**Major — "`solved_by` flips when a deeper SAT rung lands a cheaper placement" (2/3): refuted — the ladder short-circuits.** The premise ("every strategy's walker-valid solution goes into one pool") is wrong: within each variant, the strategy loop stops at the **first** walker-valid solution — deeper rungs never run. The attempts traces in this PR's own diagnostics show it directly: at i3, every variant reads `perp → Unsat, sat-surface → Unsat, sat-1ug-native → Solved` and **no `sat-2ug-native`/`sat-native` attempt exists** for any variant where 1ug solved. The candidate pool on this fixture is five `sat-1ug-native` solutions (one per variant), all cost 14 — cost wobble can change *which variant* wins, never *which rung*, because per-rung SAT/UNSAT for a fixed formula is deterministic; only solution cost wobbles. Attribution is exactly as stable as the pin claims.

**Major — "the pipe fixture's `required_entities` can't discriminate perp from SAT; cheaper SAT flips attribution" (2/3): refuted — same premise, plus priority order.** `perpendicular_template` is index 0 and solves the primary variant at iter 0, so SAT never runs on the primary at all. For a variant's SAT candidate to steal the win it would need cost < 10; the UG-pair is the minimal legal bridge (surface crossing is barred by the pipe hard-obstacle), so the best a variant can do is tie at 10 — and `pick_cheapest_candidate` is documented and implemented stable-first (primary is pushed before variants). The scenario where attribution flips *is* the scenario where the perp rung died — which is precisely what the pin exists to catch. The `required_entities` non-uniqueness is by design and documented (round 2): shape pins catch shape regressions, `solved_by` catches dispatch regressions.

**Minor — legacy fixtures need re-verification under the native ladder (1/3): refuted for the fourth time.** Receipts in the round-2, round-3 and round-4 adjudications; the PR body's verification section states the result (8/8 after every round, legacy outcomes identical: 33/28/26 vs ceilings 65/40/35, all capped fixtures still cap). The finding's own text concedes it did not read the PR ("the PR doesn't state" — it does).

**Minor — validation asymmetry: a Pipe key isn't checked against `spec_belt_tiers`/`spec_exit_dirs` (1/3): absorbed.** The harness validation now requires every `spec_kinds` key to exist in `spec_items`, `spec_belt_tiers`, and `spec_exit_dirs`, each with its own named per-fixture failure — the missing-`exit_dirs` case calls out the silent path-derived-direction fallback specifically.

**Minor — "last `JunctionSolved`" vs "committed winner" could diverge under refactor (1/3): absorbed.** The replay now asserts at most one `JunctionSolved` was observed per `solve_crossing` and fails loud (with instructions) if the invariant ever breaks, instead of silently taking the last event.

**Minor — mismatch `continue` hides the cost/entity signals (1/3): refuted-as-scoped.** The mismatch message prints the complete winning solution (name/x/y/direction per entity) plus every attempt — cost and shape are readable from the same output; collecting three verdicts would complicate the arm for no additional information.

**Minor — five boxed strategies per build (1/3): acknowledged, no action** — five small allocations per `build_bus_layout` call, against a SAT solve; the shared source of truth is worth strictly more than the allocation.

**Nits:** `unreleasable_obstacles` listing the pipe tiles DOES mirror production — production's set is non-releasable-segment entities **unioned with `hard`** (`.chain(hard.iter().copied())` in `route_bus_ghost`), and the pipe tiles are hard obstacles; the fixture models the union directly. Fixture name/genesis: the notes carry the full history intentionally — a rename would orphan the review trail. Conflict-gate rework being a wanted attribution change: already implied by the followups entry's "or rework the gates so the cheap template can actually fire"; a future PR doing that rework re-pins `solved_by` as part of its change, which is the pin working as intended.

Verification: region_fixtures 8/8 under the pinned cache with both absorptions in.

storkme:
**Round 6 adjudicated — closing round.** One genuinely new item absorbed; the rest are recycled claims (some on their fifth appearance) or rest on premises this PR's own earlier rounds disproved. After this head's required checks pass, the PR merges; any further advisory rounds will be adjudicated post-merge if they contain anything new.

**Critical — "the spec_exit_dirs requirement rejects every dump-produced pipe fixture" (3/3): refuted — the premise is vacuous by this PR's own round-3 finding.** The dump can never produce a pipe fixture at all: production filters pipe specs out of junction seeding (`keys_at_tile`), so no capture ever contains a pipe `initial_specs` entry — that is documented in the README's capture-section pipe caveat, added in round 3. Hand-authoring is the *only* path a pipe fixture can exist by, and requiring the hand-author to state the pipe's exit direction explicitly — rather than inherit a silent path-derived fallback that can flip the perpendicularity gate — is precisely the validation's purpose, stated in its error message. Nothing is "mutually exclusive with the capture path"; the capture path is empty for pipes.

**Major — "ladder swap silently re-baselines the seven pre-existing fixtures" (3/3): refuted for the fifth time.** Same receipts as the round-2/3/4/5 adjudications: region_fixtures re-run after every round, 8/8, legacy outcomes identical (33/28/26 vs ceilings 65/40/35; capped fixtures still cap); the `rust` CI job additionally ran the suite cold-cache on four successive heads, green each time.

**Major — "`solved_by` unstable under cost wobble because a later rung can field a cheaper solution" (2/3): refuted in round 5, mechanically.** Later rungs never run once an earlier rung solves within a variant (the ladder short-circuits; the PR's own attempt traces show no `sat-2ug-native` attempt anywhere `sat-1ug-native` solved). The candidate pool on the belt fixture is five same-rung solutions; wobble can move cost, not attribution.

**Major — `initial_specs ⊆ spec_items` not validated (2/3): absorbed — the round's one new finding.** The harness validation now rejects any `initial_specs` entry lacking a `spec_items` entry (which would otherwise replay with item `"?"` and Belt kind), in the same accumulate-and-report flow.

**Minor — "the ≤1 `JunctionSolved` assert panics the whole run, violating accumulate-and-report" (3/3): refuted-as-designed.** That assert guards an invariant of the replay *machinery*, not of any fixture's data. If it ever fires, every fixture's attribution is equally untrustworthy — eight per-fixture failures would be eight misleading copies of one machinery bug. A global abort with instructions is the honest report.

**Minor — "the pipe fixture doesn't protect a live production path" (2/3): the concession is the fixture's own opening sentence** ("STATIC UNIT PIN … NOT a production-reachability probe"), plus the followups entry and two prior adjudications.

**Minor — "`assert!` is elided in non-test builds" (1/3): refuted — that is `debug_assert!`.** `assert!` is compiled into all profiles; the invariant holds in release builds too.

**Nit — "green is asserted, not observed":** observed — by the `rust` CI check on every pushed head, and by the pinned-cache local runs quoted in each adjudication. The reviewer's sandbox lacking a toolchain does not convert our executed runs into assertions.

Issue #689, opened 2026-08-21
RFC-070 campaign: one selection loop — tracking


storkme:
Execution queue for the RFC-070 campaign (`docs/rfc-070-one-selection-loop.md`, PR #688). Pattern per #632/#675: lettered items, each a PR, decisions recorded here. **The RFC's decision log is the strategy record; this issue is the queue + session state.** A cold session assumes the lead role by reading the RFC, then this issue top-to-bottom.

## Decision record (2026-08-21, session lead under owner goal "complete the full campaign")

- **RFC-068 P1 is OUT of this campaign** — remains owner-gated on #675. Phase 2 carries the celldb-harness-green obligation instead (RFC-070 §Cross-RFC obligation).
- **Perp-rung census is census-only** — if it comes back "same-item crossings never seed junctions", the ~795-line rung deletion is a follow-up PR for owner review, not an autonomous act.
- **Owner review gates**: post-Phase-1 design, pre-2b flip. Everything else proceeds autonomously per the standing rules; evidence lands here and in the RFC log for async review.
- **Concurrency ceiling: 4 tracks** — collisions are ~20% of measured rework; tracks are file-disjoint by construction.

### Wave 1 (concurrent, disjoint)
- [ ] **W1a** Meter tripwire — report-only below-plan meter check over e2e exports (`crates/meter/` + wiring; reflect_port gap is known-open and out of scope for the first cut)
- [ ] **W1b** Selection scoreboard — trace instrumentation of `select_best_decomposition` (per-candidate verdicts + deciding precedence stage) + #686 census extension (RFC-070 Phase 0b)
- [ ] **W1c** Corpus baseline — committed winners-and-deciding-mechanisms baseline incl. machine-tier axis; sim-anchor the contested sample (Phase 0c; depends on W1b)
- [ ] **W1d** Housekeeping pair — deletion-cadence process PR (CLAUDE.md + rfc-template) + perp-rung junction-seeding census (census-only)
- [ ] **W1e** RFC #688 through review and merged

### Wave 2 (after W1c + RFC merged)
- [ ] **W2a** `LayoutOptions` constraint/axis split (Phase 1a; classify `planning_duty`, `research_productivity`)
- [ ] **W2b** `SelectionPolicy` + severity-aware `verdict.rs`; acceptance = baseline winners reproduced (Phase 1b — **K70-1 adjudicated here**)
- [ ] **OWNER GATE: design review** before Wave 3

[...]

PR #691, opened 2026-08-21
diag(w1d): production census of junction seeds — G1 follow-up

storkme:
## Intent

Part of RFC-070 track W1d (tracking #689). #687 established the
~795-line perpendicular-template rung (`crates/core/src/bus/ghost_router.rs`,
`solve_perpendicular_template` area) appears production-unreachable on
both shapes it handles, leaving one open hypothesis: do same-item belt
crossings ever seed junctions? This PR is a census answering exactly
that, nothing else.

**Four rounds of review absorbed** (full adjudication in PR comments).
Rounds 1-2 fixed real methodology gaps (never changing the zero
conclusion's direction). Round 3 fixed a genuine crash risk and several
hardening issues. Round 4 (closing) fixed a bug introduced by round 3's
own fix, tightened the predicate's precision claim, and corrected a
refutation from round 3 that was itself wrong (see the round-4 comment
for the full receipt).

## Scope

- **In:** a behavior-neutral `TraceEvent::JunctionSeedCensus`
  (`crates/core/src/trace.rs`), emitted once per `ghost_router` cluster
  seed right where `keys_at_tile` is built
  (`crates/core/src/bus/ghost_router.rs`), recording `cluster_tile_count`
  / `n_specs` / `n_distinct_items` / `has_pipe`, gated behind a
  value-checked `SPAGHETTIO_JUNCTION_SEED_CENSUS` (off by default); an
  `#[ignore]`d `crates/core/tests/junction_seed_census.rs` with an
  exhaustive, `assert_eq!`-checked bucket partition; the G1 entry in
  `docs/offpath-code-followups.md` updated with the reconciled result.
- **Out:** any change to `junction_solver`'s item-conflict gate, the
  rung's `tile_count == 1` / `specs.len() == 2` / `is_perpendicular`
  guards, or any other control flow. CENSUS-ONLY per the campaign brief.
- **Size:** ~750 lines net across five commits (initial diagnostic +
  four review-absorption passes), mostly the test file + doc updates.

## Census result (final, four rounds verified)

Corpus: the six tier-ladder fixtures `check_firing_census.rs` uses, plus
the six e2e "from-ore" fixtures (`tier1_iron_gear_wheel_from_ore`,
`tier2_electronic_circuit_from_ore`, `tier2_electronic_circuit_20s_from_ore`,
`tier3_plastic_bar_from_crude`, `tier4_advanced_circuit_from_ore_am2`,
`tier5_processing_unit_from_ore_am3`) that differ in belt-tier
constraint and/or rate/machine. 12 fixtures, `SPAGHETTIO_ZONE_CACHE_PATH`
pinned to `sat-zones-ci.bin`, `SPAGHETTIO_JUNCTION_SEED_CENSUS=1`.

**The rung's firing gate is threefold** (`PerpendicularTemplateStrategy::
try_solve`, ghost_router.rs 6177/6183/6186): `region.tile_count() == 1`
AND `specs.len() == 2` AND `is_perpendicular(da, db)`. This census's
event schema records neither direction, so its bucket is a NECESSARY
condition for firing (a superset), not the exact predicate — sufficient
for the zero conclusion (any firing must land in the bucket) but not the
tightest possible bucket.

```
=== junction seed census: 111 seeds across 12 fixtures ===

--- bucket breakdown (sums to 111, assert_eq!-checked) ---
n_specs == 0 after the pipe filter (never a crash):    0
single-tile, 1 spec (belt-over-forbidden-tile bypass): 11
single-tile, 2 specs, different items:                 31
single-tile, 2 specs, SAME item (necessary-superset):  0
single-tile, >2 specs (outside rung's specs.len()==2): 0
multi-tile,  all participants distinct items:          3
multi-tile,  item-sharing pair somewhere in union:     66

cross-check (printed, not asserted — see round 4 note): pipe-tagged &
single-spec = 11, pipe-tagged but NOT single-spec = 0, single-spec but
NOT pipe-tagged = 0
```

**Headline: 0 seeds fall in the necessary-superset bucket for the
rung's firing conditions.** All 111 seeds partition exhaustively into 7
buckets (assert_eq!-checked): 0 zero-spec-after-filter, 11 single-spec
pipe-bypass, 31 single-tile 2-different-item crossings, 0 single-tile
>2-spec, 3 multi-tile all-distinct, 66 multi-tile with an item-sharing
pair somewhere in the cluster's union.

The 66 is a separate, explicitly weaker signal — likely dominated by
the mundane trunk-owns-its-own-non-last-tap same-item adjacency
(`rfc-unified-belt-specs.md` Phase 2), not evidence of a crossing.

**Net: this corpus found no reachable instance of the rung's necessary
firing conditions anywhere in the pipeline.** Not a universal proof —
one 12-fixture corpus, seeds counted across every `route_bus_ghost`
invocation (candidates + retries), not deduplicated — but real,
corpus-wide, zero-based evidence for the "never occurs" branch. Still
not sufficient alone for the owner-reviewed deletion call.

## Review rounds absorbed (full receipts in PR comments)

- **Round 1** (2 major): cluster-wide vs per-tile conflation (fixed via
  `cluster_tile_count`); unconditional hot-path cost (fixed via an
  opt-in env gate).
- **Round 2** (4): missing `specs.len()==2` check; `spec_items`-miss
  fallback hid same-item pairs (fixed via key-prefix recovery); doc
  arithmetic didn't reconcile; headline was hand-transcribed (fixed via
  test-computed, exhaustive buckets).
- **Round 3** (6, one applied only in part): a crash risk
  (`unreachable!()` reachable for `n_specs==0`, fixed); presence- vs
  value-based env gate; merge-tap key-parsing trap; prose-only
  cross-check; drifted line refs — all fixed. One claim
  ("unsafe { set_var } unnecessary under edition 2021") was refuted at
  the time — **that refutation turned out to be wrong** (see round 4).
- **Round 4, closing** (7, one a self-correction): a real bug from
  round 3's own fix (the new `zero_specs_after_filter` bucket tripped
  round 2's correlation assert by construction) — fixed by demoting
  both correlation asserts to diagnostics; missing `feeder:` prefix;
  "exact predicate" overclaimed precision, reworded to "necessary
  superset"; misleading "expected 0" prose; undisclosed O(n) cost.
  **And: round 3's refutation of the `unsafe` finding was itself
  wrong** — the decisive test is whether the bare, unwrapped call
  compiles (it does, at this crate's actual edition 2021; it doesn't at
  edition 2024, confirmed with `rustc` directly), not whether wrapping
  it triggers `unused_unsafe` (it doesn't, at either edition, because
  `set_var`/`remove_var` carry a lint-suppressing attribute during the
  edition-2024 migration). The `unsafe` blocks are now removed.

## Verification

- [x] `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` (unpiped) — exit 0, full suite green, re-run clean after all four review rounds
- [x] `cargo clippy --workspace -- -D warnings` (the exact `ci.yml` invocation) — exit 0, clean, all five rounds
- [x] `cargo clippy --manifest-path crates/core/Cargo.toml --test junction_seed_census -- -D warnings` — exit 0 (confirms no `unused_unsafe` fallout either with or without the wrapper)
- [x] WASM rebuild: `wasm-pack build crates/wasm-bindings --target web --out-dir <abs path>/web/src/wasm-pkg` — exit 0, all five rounds
- [x] Census run (final): `SPAGHETTIO_ZONE_CACHE_PATH=<abs path>/crates/core/data/sat-zones-ci.bin cargo test --manifest-path crates/core/Cargo.toml --test junction_seed_census -- --ignored --nocapture`, then `git checkout -- crates/core/data/sat-zones-ci.bin` immediately after every run — the pin file never appears in this branch's diff. Numbers confirmed unchanged across all four review rounds.
- [ ] Snapshot inspection — not applicable, no layout-engine behavior change (diagnostic-only)

## Deviations from intent

None beyond the review-driven corrections described above — including
one finding refuted with receipts (later found wrong and corrected) and
one bug introduced by absorbing a prior round's finding (also caught and
fixed). All part of the review process working as intended.

## Models / contracts touched

`docs/offpath-code-followups.md`'s G1 entry — updated with the final
reconciled census result across four review rounds, per the
followups-doc contract.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/crates/core/src/bus/ghost_router.rs
+++ b/crates/core/src/bus/ghost_router.rs
@@ -2898,9 +2898,13 @@ pub fn route_bus_ghost(
     // always installs a trace collector *and* sink, so gating on
     // `trace::is_active()` alone would not skip this in production; an
     // explicit opt-in env var is what actually makes it zero-cost by
-    // default (round 1 review finding).
-    let junction_seed_census_enabled =
-        std::env::var_os("SPAGHETTIO_JUNCTION_SEED_CENSUS").is_some();
+    // default (round 1 review finding). Value-based, not merely
+    // presence-based (round 3 review finding): a shop-wide diagnostics
+    // block that sets every `SPAGHETTIO_*` var to `"0"` to disable them
+    // must not accidentally enable this one.
+    let junction_seed_census_enabled = std::env::var("SPAGHETTIO_JUNCTION_SEED_CENSUS")
+        .map(|v| v == "1" || v.eq_ignore_ascii_case("true"))
+        .unwrap_or(false);
 
     for cluster in &clusters {
         // `corridor_handled` grows during this loop — a prior cluster's
@@ -3000,10 +3004,30 @@ pub fn route_bus_ghost(
             // the item as the first colon-delimited segment after its
             // prefix — recover it from the key itself (same technique the
             // fluid catch-up above already uses) before falling back to
-            // the key as an absolute last resort.
+            // the key as an absolute last resort. `"tap:mergetap:"` MUST
+            // be checked before the plain `"tap:"` prefix (round 3 review
+            // finding): a merge-tap feed key is
+            // `format!("tap{MERGE_TAP_SEGMENT_TAG}{item}:{x}:{y}")` =
+            // `"tap:mergetap:{item}:{x}:{y}"`, so stripping only `"tap:"`
+            // would recover the literal string "mergetap" as the item
+            // instead of the real one. Currently these keys are always
+            // registered as regular `BeltSpec`s (so `spec_items.get(key)`
+            // above already resolves them and this fallback path is never
+            // reached for them today) — ordering the prefixes this way
+            // keeps it correct if that ever changes.
             fn resolve_item<'a>(key: &'a str, spec_items: &'a FxHashMap<String, String>) -> &'a str {
                 if let Some(item) = spec_items.get(key) {
                     return item.as_str();
+                }
+                // Checked ahead of the generic "tap:" prefix below, tied
+                // to the real constant (not a duplicated literal) so it
+                // can't drift from the merge-tap key format itself.
+                if let Some(rest) =
+                    key.strip_prefix("tap").and_then(|r| r.strip_prefix(MERGE_TAP_SEGMENT_TAG))
+                {
+                    if let Some((item, _)) = rest.split_once(':') {
+                        return item;
+                    }
                 }
                 for prefix in ["trunk:", "tap:", "flow:"] {
                     if let Some(rest) = key.strip_prefix(prefix) {
--- a/crates/core/src/bus/ghost_router.rs
+++ b/crates/core/src/bus/ghost_router.rs
@@ -2964,6 +2964,37 @@ pub fn route_bus_ghost(
             .map(|(key, _)| key.as_str())
             .collect();
 
+        // Census-only instrumentation (offpath-code-followups.md G1
+        // follow-up, #689 W1d): record this seed's shape before it reaches
+        // `solve_crossing`. Purely observational — `trace::emit` is a
+        // no-op unless a collector/sink is active, so this cannot change
+        // routing. `has_pipe` is measured over the RAW spec set touching
+        // the cluster's tiles, i.e. BEFORE the `SpecKind::Pipe` filter
+        // above runs, so it can actually detect pipe presence instead of
+        // trivially reading false through the filter that excludes them.
+        {
+            let n_specs = keys_at_tile.len();
+            let n_distinct_items: usize = keys_at_tile
+                .iter()
+                .filter_map(|&key| spec_items.get(key).map(|s| s.as_str()))
+                .collect::<FxHashSet<&str>>()
+                .len();
+            let has_pipe = routed_paths.iter().any(|(key, path)| {
+                path.iter().any(|t| cluster_tiles.contains(t))
+                    && matches!(
+                        spec_kinds.get(key.as_str()),
+                        Some(crate::bus::junction::SpecKind::Pipe)
+                    )
+            });
+            trace::emit(trace::TraceEvent::JunctionSeedCensus {
+                seed_x: cluster[0].0,
+                seed_y: cluster[0].1,
+                n_specs,
+                n_distinct_items,
+                has_pipe,
+            });
+        }
+
         // Pending crossings for the DeferredExit check: the subset of
         // `crossing_set` whose cluster hasn't committed yet. Excluding
         // `corridor_handled` avoids false deferrals when a zone's exit
--- a/crates/core/src/bus/ghost_router.rs
+++ b/crates/core/src/bus/ghost_router.rs
@@ -3000,11 +3000,12 @@ pub fn route_bus_ghost(
             // synth key when its path is ENTIRELY on pipe tiles; a fluid
             // synth key touching any non-pipe tile stays untagged). Every
             // synthetic key format in this file (`trunk:{item}:{x}`,
-            // `tap:{item}:{x}:{tap_y}`, `flow:{item}:{x}[:ret:{y}]`) puts
-            // the item as the first colon-delimited segment after its
-            // prefix — recover it from the key itself (same technique the
-            // fluid catch-up above already uses) before falling back to
-            // the key as an absolute last resort. `"tap:mergetap:"` MUST
+            // `tap:{item}:{x}:{tap_y}`, `flow:{item}:{x}[:ret:{y}]`,
+            // `feeder:{item}:{x}:{y}`, ghost_router.rs:1935) puts the item
+            // as the first colon-delimited segment after its prefix —
+            // recover it from the key itself (same technique the fluid
+            // catch-up above already uses) before falling back to the
+            // key as an absolute last resort. `"tap:mergetap:"` MUST
             // be checked before the plain `"tap:"` prefix (round 3 review
             // finding): a merge-tap feed key is
             // `format!("tap{MERGE_TAP_SEGMENT_TAG}{item}:{x}:{y}")` =
@@ -3029,13 +3030,28 @@ pub fn route_bus_ghost(
                         return item;
                     }
                 }
-                for prefix in ["trunk:", "tap:", "flow:"] {
+                for prefix in ["trunk:", "tap:", "flow:", "feeder:"] {
                     if let Some(rest) = key.strip_prefix(prefix) {
                         if let Some((item, _)) = rest.split_once(':') {
                             return item;
                         }
                     }
                 }
+                // Absolute last resort, reached only if BOTH the
+                // `spec_items` lookup and every known prefix above miss —
+                // not observed in any corpus run to date (round 4 review
+                // fence, since this path is genuinely untested): treating
+                // the raw key as a unique pseudo-item is the conservative
+                // choice for THIS census's specific failure mode (round 1
+                // review: never silently manufacture a same-item pair
+                // that isn't real), but it has its own opposite residual
+                // risk — two specs that truly share an item under some
+                // future, unrecognized key format would each get their
+                // own unique key here and read as "distinct", hiding a
+                // real same-item pair instead of fabricating one. If this
+                // path is ever observed firing (add an eprintln! probe if
+                // investigating), extend the prefix list above rather
+                // than trusting the count it produces.
                 key
             }
             let n_distinct_items: usize = keys_at_tile
@@ -3043,6 +3059,12 @@ pub fn route_bus_ghost(
                 .map(|&key| resolve_item(key, &spec_items))
                 .collect::<FxHashSet<&str>>()
                 .len();
+            // Round 4 review: this is O(sum of every routed path's
+            // length) per cluster when the census is on, not O(1) — it
+            // re-scans every path in `routed_paths` (the same cost class
+            // `keys_at_tile`'s own filter above already pays). Acceptable
+            // for an opt-in diagnostic gated off by default, but not
+            // free; don't reach for this pattern in an always-on path.
             let has_pipe = routed_paths.iter().any(|(key, path)| {
                 path.iter().any(|t| cluster_tiles.contains(t))
                     && matches!(
--- a/crates/core/src/bus/ghost_router.rs
+++ b/crates/core/src/bus/ghost_router.rs
@@ -2988,18 +2988,35 @@ pub fn route_bus_ghost(
         // trivially reading false through the filter that excludes them.
         if junction_seed_census_enabled {
             let n_specs = keys_at_tile.len();
+            // Round 2 review: the round-1 fallback (unwrap_or the raw key,
+            // guaranteed unique) errs the OTHER way from round 1's
+            // filter_map bug — it can HIDE a genuine same-item pair
+            // whenever `spec_items` lacks an entry, which is not
+            // hypothetical (`ghost_router.rs`'s fluid catch-up only tags a
+            // synth key when its path is ENTIRELY on pipe tiles; a fluid
+            // synth key touching any non-pipe tile stays untagged). Every
+            // synthetic key format in this file (`trunk:{item}:{x}`,
+            // `tap:{item}:{x}:{tap_y}`, `flow:{item}:{x}[:ret:{y}]`) puts
+            // the item as the first colon-delimited segment after its
+            // prefix — recover it from the key itself (same technique the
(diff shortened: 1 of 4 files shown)


storkme:
## Round 3 adjudication (a39cdc91 review)

All six findings verified at source. Five applied, one partially refuted with a reproducible receipt. Pushed in 2a363139.

**[3/3] `unreachable!()` reachable for `n_specs == 0` — APPLIED.** Confirmed: a single-tile cluster whose only participant is a pipe (filtered out of `keys_at_tile`, but let through `any_undecidable`'s forbidden-tile bypass, which counts ALL `routed_paths` including pipes) produces `n_specs == 0`, which fell into the `_ => unreachable!()` arm. That's exactly the shape this census exists to observe, so it must never crash. Added a `zero_specs_after_filter` bucket, checked first, making every remaining arm exhaustive without a data-dependent catch-all. Re-ran: `0` in this corpus (didn't fire), but the code no longer depends on that.

**[2/3] `unsafe { std::env::set_var }` "unnecessary under edition 2021" — REFUTED (with the other half applied).** Tested directly on this repo's pinned toolchain (`rustc 1.95.0`):
```
$ rustc --edition 2021 -e 'fn main(){ unsafe { std::env::set_var("X","1"); } }' # compiles clean
$ rustc --edition 2021 --deny unused-unsafe -e 'fn main(){ unsafe { println!("hi"); } }'
warning: unnecessary `unsafe` block  # confirms the lint mechanism works
```
`unsafe { println!(...) }` correctly triggers `unused_unsafe` under `#[deny(unused_unsafe)]`; `unsafe { std::env::set_var(...) }` does not, on the same toolchain, same edition. That means the wrapper is required here — `env::set_var`'s signature became `unsafe fn` in std as of a Rust release well before 1.95.0, and it is NOT edition-gated. Kept the `unsafe` block. Applied the valid half of the same finding: the var was never restored — added an `EnvVarGuard` (RAII, restores previous value or absence on `Drop`, including panic unwind) so a future second `#[test]` in this file can't inherit the mutation.

**[1/3] Presence- vs value-based env gate — APPLIED.** `var_os(..).is_some()` would treat `SPAGHETTIO_JUNCTION_SEED_CENSUS=0` as "enabled". Changed to require `"1"` or case-insensitive `"true"`.

**[1/3] `resolve_item` merge-tap prefix trap — APPLIED.** Confirmed the exact key shape at `ghost_router.rs:1635`: `format!("tap{MERGE_TAP_SEGMENT_TAG}{item}:{x}:{y}")` = `"tap:mergetap:{item}:{x}:{y}"` (`MERGE_TAP_SEGMENT_TAG = ":mergetap:"`, `common.rs:1032`). Stripping bare `"tap:"` would have recovered the literal `"mergetap"` as the item. Confirmed inert today (merge-tap keys are pushed as regular `BeltSpec`s with `item: lane.item.clone()`, so `spec_items.get(key)` already resolves them before the fallback runs) — fixed anyway by checking the `MERGE_TAP_SEGMENT_TAG`-derived prefix first so it can't matter if that ever changes.

**[1/3] "pipe-tagged == single-spec" prose-only — APPLIED.** Added `assert_eq!` on both directions (previously only printed).

**[1/3] Drifted line refs — APPLIED.** Rung predicate is at `ghost_router.rs` 6177 (fn)/6183 (`tile_count`)/6186 (`specs.len`), not ~6166-6171. Fluid catch-up sweep is 2643-2667, not ~2648-2667.

Re-ran the full census after all fixes (`SPAGHETTIO_ZONE_CACHE_PATH` pinned to `sat-zones-ci.bin`): identical 111 seeds, all six buckets unchanged, both cross-checks now `assert_eq!`-enforced and passing. Headline unchanged: **0 seeds match the rung's exact predicate** (`cluster_tile_count == 1 && n_specs == 2 && n_distinct_items == 1`).

Verification re-run clean: `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` (exit 0, unpiped), `cargo clippy --workspace -- -D warnings` (exit 0) and `cargo clippy --manifest-path crates/core/Cargo.toml --test junction_seed_census -- -D warnings` (exit 0, confirms `unused_unsafe` does not fire), WASM build (exit 0).

storkme:
## Round 4 adjudication (2a363139 review, closing round)

All seven findings verified at source. Six applied, and one is a self-correction: round 3's own refutation of finding 7 was itself wrong. Pushed in 7efeac55.

**[major 2/3] `pipe_tagged_but_not_single_spec == 0` assert contradicted my own round-3 addition — APPLIED (via finding 2's broader fix).** Confirmed: a `zero_specs_after_filter` seed has `has_pipe = true` (its sole participant was necessarily a filtered-out pipe — that's the only way `n_specs` reaches 0 while the cluster still passed `any_undecidable`'s bypass) and `n_specs != 1`, so it lands in `pipe_tagged_but_not_single_spec`, tripping the round-2 `assert_eq!`. Rather than the narrower fix (compare against `zero_specs_after_filter`), applied finding 2's broader fix below, which resolves this too.

**[minor 2/3] Demote both correlation asserts to diagnostics — APPLIED.** Removed both `assert_eq!`s; now printed with a loud `⚠ UNEXPECTED` marker when nonzero. `bucket_sum == total_seeds` remains the sole hard assertion (a true tautology from an exhaustive match, not a data-dependent correlation).

**[minor 2/3] `resolve_item` missing `feeder:` prefix — APPLIED.** Confirmed the key shape at `ghost_router.rs:1935`: `format!("feeder:{}:{}:{}", lane.item, input_x, out_y)`. Added to the prefix list; comment-fenced the raw-key last resort with its own named residual bias (opposite direction from the round-1/2 bugs — hides rather than fabricates a same-item pair, unobserved in every run to date, same as merge-tap).

**[minor 2/3] "Exact firing predicate" overclaims — APPLIED.** Confirmed at `ghost_router.rs:6183/6186`: the guards are `tile_count() > 1` and `specs.len() != 2`, but `try_solve` (6177-6194) ALSO calls `is_perpendicular(da, db)` on the two specs' directions before accepting — a third gate this census's event schema doesn't record (no direction fields). Reworded "exact firing predicate" to "necessary superset" everywhere (test header, followups doc, print statements). The zero conclusion is unaffected — a necessary-condition bucket being empty still proves zero firings — only the precision claim was wrong.

**[minor 1/3] "expected 0 by construction" contradicts the reported 11 — APPLIED.** Fixed to "expected to equal the single-spec bypass count; any excess is the signal."

**[minor 1/3] `has_pipe` complexity undisclosed — APPLIED.** Added a comment: O(sum of every routed path's length) per cluster when the census is on, not O(1).

**[major, self-correction] `unsafe { std::env::set_var }` — round 3's REFUTATION of this finding was itself wrong; corrected.**

Round 3 tested only whether *wrapping* the call in `unsafe {}` triggered `unused_unsafe` (it didn't) and concluded from that that the wrapper was required. That test doesn't distinguish "required" from "tolerated by a lint carve-out." The decisive test is whether the **bare, unwrapped** call compiles:

```
$ rustc --edition 2021 -e 'fn main(){ std::env::set_var("X","1"); }'   # compiles, zero output
$ rustc --edition 2024 -e 'fn main(){ std::env::set_var("X","1"); }'
error[E0133]: call to unsafe function `set_var` is unsafe and requires unsafe block
```

`crates/core/Cargo.toml` pins `edition = "2021"` (checked directly, not assumed, per this round's instruction). On this crate's actual edition, `set_var`/`remove_var` are plain safe functions — they carry what appears to be a `#[rustc_deprecated_safe_2024]`-style attribute that specifically suppresses `unused_unsafe` for pre-emptive wrapping, easing the migration to edition 2024 (where they become genuinely unsafe) without warning either calling convention in the meantime. That attribute is exactly what made round 3's test misleading.

Removed both `unsafe` blocks in `EnvVarGuard::set`/`Drop`; corrected the comment to state the edition-conditional truth and flag re-adding `unsafe` if this crate ever bumps to edition 2024.

---

Re-ran the full census after all fixes (`SPAGHETTIO_ZONE_CACHE_PATH` pinned): identical 111 seeds, all 7 buckets unchanged, no `⚠ UNEXPECTED` markers fired. Headline unchanged: **0 seeds in the necessary-superset bucket for the rung's firing conditions.**

Verification re-run clean: `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` (exit 0, unpiped), `cargo clippy --workspace -- -D warnings` (exit 0) and `cargo clippy --manifest-path crates/core/Cargo.toml --test junction_seed_census -- -D warnings` (exit 0, confirms no `unused_unsafe` fallout from removing the wrapper), WASM build (exit 0).

storkme:
## Round 5 adjudication (7efeac55 review, final pass)

Three metric-integrity caveats, no blocking findings. All three applied. Pushed in 7987c0aa.

**1. The load-bearing zero was only printed, not asserted — APPLIED.** Added `assert_eq!(single_tile_same_item, 0, ...)` with a message stating exactly what a failure means (the reachability conclusion changed; update `docs/offpath-code-followups.md`'s G1 entry before trusting any deletion call). Agreed this doesn't conflict with round 4's "never abort on discovery" — that call was about the exploratory pipe-tagged/single-spec correlation checks (an unexpected value there is itself informative, not a conclusion). `single_tile_same_item` is the conclusion this whole census exists to check, and since it's `#[ignore]`d, the assert only ever runs when someone deliberately re-executes the evidence — exactly the moment a changed answer should be loud.

**2. "Counts every route_bus_ghost invocation" overstated — APPLIED, and now measured directly, not just caveated.** Confirmed at `layout.rs:390`: `run_layout_with_retry_inner` calls `crate::trace::truncate_events(trace_start)` on a junction-cap/substation retry, discarding that candidate's pass-1 trace events (including any `JunctionSeedCensus`) before pass 2 runs — the comment there literally says "Discard pass-1 events from the collector so `result.trace` reflects only the retried (final) pass." Fixed the module-doc sentence. Went further than a prose fix: since `TraceEvent::LayoutRetried` is emitted right after the truncate (`layout.rs:427`, confirmed it survives), the test now tallies real `LayoutRetried` occurrences and prints the count next to the summary line. Measured on this corpus: **6 candidate-evaluations internally retried** — a real, previously-invisible number, not a guess. Documented why this can't flip the zero to a false positive: truncation only *removes* observations this census would otherwise have counted; it can never fabricate a same-item hit that didn't occur. The zero is a lower bound on this corpus's true same-item count, not a certainty that no truncated pass-1 ever hit the rung's shape.

**3. Hide-bias caveat where the pipe-bypass corroboration is discussed — APPLIED.** Added: the fluid catch-up sweep (`ghost_router.rs` ~2643-2667) only tags a synth key as Pipe when its path sits entirely on pipe tiles, and `resolve_item`'s raw-key last resort can make two specs that truly share an item read as "distinct" under an unrecognized key format — both bias toward *hiding* a same-item pair, never fabricating one. Printed alongside the corroboration and noted in the followups doc: any deletion follow-up citing this census's zero should carry the caveat forward. Neither bias has been observed firing in any run to date.

Re-ran the full census after all fixes (`SPAGHETTIO_ZONE_CACHE_PATH` pinned): identical 111 seeds, all 7 buckets unchanged, new `assert_eq!` passes, and the corpus now reports **6 retry episodes** directly (previously unmeasured). Headline unchanged: **0 seeds in the necessary-superset bucket for the rung's firing conditions** — now a hard-asserted, not merely printed, conclusion.

Verification re-run clean: `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` (exit 0, unpiped), `cargo clippy --workspace -- -D warnings` (exit 0), `cargo clippy --manifest-path crates/core/Cargo.toml --test junction_seed_census -- -D warnings` (exit 0), WASM build (exit 0).

New head: **7987c0aa**.

storkme:
**Round 6 adjudicated post-merge (closing-round protocol, declared at round 5).** All four items are message-wording / doc-precision on an #[ignore]d diagnostic; the census numbers and the zero conclusion are untouched. Carried into the W1c PR (recorded on #689): (1) the round-5 assert's failure message overclaims — it should say the NECESSARY-SUPERSET bucket went non-zero (re-examine perpendicularity before trusting the conclusion), and the discovery dumps must print BEFORE the assert; (2) the round-4 'one hard assertion' comment is stale — two exist now, one invariant, one data-dependent; (3) one sentence on the compute-without-collector cost when the env var is set outside tracing; (4) the 'counts every candidate' claim — run_candidate truncates losing candidates' events (the same oracle gap #692 documented), so the census sees winners' passes only; direction conservative, wording fixed where cited. The nits (non-pipe spec wording, O(Σpaths) disclosure) ride along.

PR #692, opened 2026-08-21
diag(w1b): selection scoreboard — per-candidate verdicts and the deciding precedence stage

storkme:
## Intent

RFC-070 Phase 0b (campaign #689, track W1b): make `select_best_decomposition`'s
own verdicts observable. The RFC names the Phase-0 baseline as the oracle every
later phase diffs against — "per corpus fixture: candidate set, per-candidate
verdicts under each mechanism, winner, and *which precedence-chain stage decided
it*" — and today the trace stream says only which name won. This adds the events
and renders them.

## Scope

- **In:** two trace events (`SelectionCandidateEvaluated`, `SelectionDecided`)
  plus the recording plumbing at each existing decision site; a
  `selection_scoreboard_census` diagnostic beside #686's check-firing census;
  `selection_scoreboard_contract`, a NON-ignored CI test pinning the oracle's
  contract (review round 2), and the single-source `CANDIDATE_ORDER` with
  construction-time alignment asserts that back it;
  the two #675 follow-ups recorded on #686's closing comment
  (`layout_signature` enrichment, no-op denominator); RFC decision-log entries.
- **Out:** Phase 0c — the committed baseline data, the machine-tier axis
  widening, the named parity corpus. This PR builds the instrument and runs it
  once; freezing a baseline belongs with the corpus definition, not ahead of it.
- **Out:** closing the refusal-attribution gap. `produce()` stringifies its own
  validation failure as `e.to_string().lines().next()` and drops the issue list,
  so *which categories* fire inside a self-refused candidate is still invisible
  (#686's stated gap). Recovering it changes what production code keeps — not
  additive, so not Phase 0b.
- **Size:** 756 added lines, over the ~400 norm. ~250 of them are comment lines
  and a further large share is the two diagnostics' printed interpretation
  prose. It *is* splittable — events / census / docs — and I chose not to: the
  event field set was designed backwards from what the census needed to print,
  so landing the schema alone would ship an unexercised contract, and the census
  run below is the only receipt that the events carry what the RFC asked for.
  Judgement call, stated rather than hidden.

## Verification

- [x] `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` —
      **exit 0**, unpiped (redirected to a file, exit captured from the run
      itself). **1183 passed, 0 failed, 74 ignored** across 27 test binaries —
      1182 before review round 2, +1 for the new non-ignored contract test.
      Run four times across the PR's life; identical counts each time.
- [x] **The contract test discriminates** — verified by breaking the code three
      ways and watching it fail, then restoring: mis-tag the error-free stage
      (fails the stage assertion), emit `&self.0[1..]` (fails the order
      assertion), swap two runs in the index list (fails the alignment assert,
      naming slot 4).
- [x] Snapshot round-trip: `SPAGHETTIO_DUMP_SNAPSHOTS=1 … tier1_iron_gear_wheel`
      then decode — all 7 `SelectionCandidateEvaluated` rows and the
      `SelectionDecided` survive the `.fls` serde encoding. This is also where
      the e2e/production candidate-set divergence surfaced (below).
- [x] `cargo clippy --workspace -- -D warnings` (CI's exact invocation) — exit 0.
      Also `cargo clippy --test check_firing_census -- -D warnings` — exit 0
      (CI lints libs only; this is new-file hygiene).
- [x] `wasm-pack build crates/wasm-bindings --target web --out-dir "$(pwd)/web/src/wasm-pkg"`
      — exit 0. Generated TS carries both closed enums as string unions
      (`SelectionStage = "merge-tap" | … | "first-produced"`). Two TS switches
      dispatch on `evt.phase`: the streaming renderer's has a permissive
      `default`, and `junctionTrace`'s is guarded by an `isJunctionEvent` type
      predicate over a fixed phase set, so neither sees a new variant. Nothing
      builds an exhaustive `Record` over the phase union. (`tsc` itself was not
      run — `web/node_modules` is not installed in this worktree; CI's web job
      covers it.)
- [x] Both diagnostics run under
      `SPAGHETTIO_ZONE_CACHE_PATH=crates/core/data/sat-zones-ci.bin`; the pin
      file is untouched in every commit (`git status` clean after each run).
- [ ] Browser eyeball — N/A, no geometry change.

**Byte-identity evidence** (the property this PR rests on). The full suite is
green untouched, including the selection pins that would catch a precedence
change — `merge_tap_does_not_shadow_di_on_pooled_yellow`,
`decomposition_search_picks_native_on_clean_partitioned_case`,
`cell_candidate_resolves_ec15_refusal` — and every golden hash. Independently:
the check-firing census's category table over 36 builds is unchanged from the
#686 run (belt-flow-reachability 240, input-rate-delivery 179, belt-flow-path
60, belt-detour 33, belt-dead-end 18, row-input-belt-margin 18), which is 36
layouts' worth of validator output agreeing to the issue.

### The scoreboard (six census fixtures, default options)

```
=== RFC-070 selection scoreboard (default options, one build per fixture) ===
(err/selw/laww are the counts the DECISION computed, sourced in the `from` column; `-` means no mechanism computed one — a gap, not a zero. `reason` is produce()'s refusal text, or the accepted=no tag for a candidate that built but failed the hard gate.)

-- iron-gear-wheel@10:assembling-machine-1
   candidate          outcome       score  acc   err  selw  laww  from                  kinds     reason
  *native             produced     0.1518  yes     0     0     0  clean-flags           -         -
   k1-shape-fix       not-run           -    -     -     -     -  -                     -         -
   size-split-2       not-run           -    -     -     -     -  -                     -         -
   merge-tap          not-run           -    -     -     -     -  -                     -         -
   cell-composed      produced     0.1472  yes     0     0     1  clean-flags           -         -
   direct-insertion   not-run           -    -     -     -     -  -                     -         -
   horizontal-stack   not-run           -    -     -     -     -  -                     -         -
   => winner: native   deciding stage: best-error-free

-- electronic-circuit@10:assembling-machine-1
   candidate          outcome       score  acc   err  selw  laww  from                  kinds     reason
   native             refused           -    -     -     -     -  -                     -         lane-aware delivery: item copper-cable rate 30.00/s exceeds per-lane capacity 22.50/s on a sideload-fed single trunk (2 producers, no balancer)
   k1-shape-fix       not-run           -    -     -     -     -  -                     -         -
   size-split-2       not-run           -    -     -     -     -  -                     -         -
   merge-tap          not-run           -    -     -     -     -  -                     -         -
   cell-composed      produced    -0.0442  yes     0     0     1  clean-flags           -         -
  *direct-insertion   produced     0.0368  yes     0     0     0  clean-flags           -         -
   horizontal-stack   produced     0.1273  yes     0     1     0  clean-flags           -         -
   => winner: direct-insertion   deciding stage: best-error-free

-- electronic-circuit@30:assembling-machine-2
   candidate          outcome       score  acc   err  selw  laww  from                  kinds     reason
  *native             produced    -0.1334   no     -     -     -  -                     c0/s3/x0  1 missing-balancer-template warning(s)
   k1-shape-fix       not-run           -    -     -     -     -  -                     -         -
   size-split-2       not-run           -    -     -     -     -  -                     -         -
   merge-tap          produced    -0.1378  yes     -     -     -  -                     c0/s65/x0 -
   cell-composed      refused           -    -     -     -     -  -                     -         cells: copper-cable has 2 in-ports for copper-plate across 2 rows (multi-row corridor fan not implemented, #433)
   direct-insertion   refused           -    -     -     -     -  -                     -         direct insertion failed validation: Validation failed:
   horizontal-stack   refused           -    -     -     -     -  -                     -         horizontal-stack failed validation: Validation failed:
   => winner: native   deciding stage: merge-tap

-- plastic-bar@5:chemical-plant
   candidate          outcome       score  acc   err  selw  laww  from                  kinds     reason
  *native             produced     0.2624  yes     0     0     0  clean-flags           -         -
   k1-shape-fix       not-run           -    -     -     -     -  -                     -         -
   size-split-2       not-run           -    -     -     -     -  -                     -         -
   merge-tap          not-run           -    -     -     -     -  -                     -         -
   cell-composed      produced     0.1597  yes     0     0     1  clean-flags           -         -
   direct-insertion   not-run           -    -     -     -     -  -                     -         -
   horizontal-stack   not-run           -    -     -     -     -  -                     -         -
   => winner: native   deciding stage: best-error-free

-- advanced-circuit@5:assembling-machine-2
   candidate          outcome       score  acc   err  selw  laww  from                  kinds     reason
   native             produced    -0.0279  yes     0     2     0  di-vs-native          -         -
   k1-shape-fix       not-run           -    -     -     -     -  -                     -         -
   size-split-2       not-run           -    -     -     -     -  -                     -         -
   merge-tap          not-run           -    -     -     -     -  -                     -         -
   cell-composed      produced    -0.2549  yes     -     -     -  -                     -         -
   direct-insertion   produced    -0.0279  yes     0     2     0  di-vs-native          -         -
  *horizontal-stack   produced    -0.0283  yes     0     0     0  horizontal-vs-native  -         -
   => winner: horizontal-stack   deciding stage: scoped-pairwise

-- processing-unit@2:assembling-machine-3
   candidate          outcome       score  acc   err  selw  laww  from                  kinds     reason
  *native             produced    -0.3878  yes     0     7     0  di-vs-native          -         -
   k1-shape-fix       not-run           -    -     -     -     -  -                     -         -
   size-split-2       not-run           -    -     -     -     -  -                     -         -
   merge-tap          not-run           -    -     -     -     -  -                     -         -
   cell-composed      produced    -1.3728  yes     0     0     1  clean-flags           -         -
   direct-insertion   produced    -0.3878  yes     0     7     0  di-vs-native          -         -
   horizontal-stack   produced    -0.3645  yes     0    12     0  horizontal-vs-native  -         -
   => winner: native   deciding stage: best-error-free

=== outer-selection summary (one row per fixture) ===
fixture                                      winner             deciding stage
iron-gear-wheel@10:assembling-machine-1      native             best-error-free
electronic-circuit@10:assembling-machine-1   direct-insertion   best-error-free
electronic-circuit@30:assembling-machine-2   native             merge-tap
plastic-bar@5:chemical-plant                 native             best-error-free
advanced-circuit@5:assembling-machine-2      horizontal-stack   scoped-pairwise
processing-unit@2:assembling-machine-3       native             best-error-free
```

Three of the five stages fire on this slice. **`best-accepted` decides nothing**
— the generic soft score, one of the three mechanisms the RFC's Summary names,
never picked a winner here; its live role was the `accepted` hard gate
(`ec@30` native, `acc=no`). Worth re-checking at Phase 0c's wider corpus before
anyone treats it as a ranking. `first-produced` also never fired.

The blanks are the other finding. A merge-tap decision short-circuits the
`clean_flags` tier, so the only counts such a fixture can carry are ones a
scoped pairwise already computed — and on `ec@30` DI and horizontal both
refused, so there are none: the kinds key (`c0/s3/x0` vs `c0/s65/x0`) is the
whole of what that decision looked at, and so the whole of what any oracle can
report about it. (The first version of this claim generalised it to every
merge-tap decision, which is wrong — native's counts do get recorded whenever
DI or horizontal produced. Caught by the review bot, 2/3.)

### No-op check, with the enriched signature

```
  di-off       5/6 comparable builds identical to default (+0 with no default to compare)  (native-adjacent)
  di-forced    5/6 comparable builds identical to default (+0 with no default to compare)  (user-elected)
  cells-off    6/6 comparable builds identical to default (+0 with no default to compare)  (native-adjacent)
  hs-off       5/6 comparable builds identical to default (+0 with no default to compare)  (native-adjacent)
  partitioned  4/6 comparable builds identical to default (+0 with no default to compare)  (user-elected)
```

Identical ratios to #686's round-6 run, and zero non-comparable builds — **both
follow-ups are measurably inert on this fixture set**. No layout here differs
from its default only in `carries`/`mirror`/`rate`, and no fixture's default
refused. They harden the instrument for a future run, exactly like the
`defaults_built` guard before them. Said plainly rather than presented as a fix
that fixed something.

## Deviations from intent

- **Five stages, not four.** The RFC's Motivation calls the precedence chain
  four-stage; the scoreboard names five. The first `.or()` link answers with two
  different mechanisms — merge-tap's `quality_key` verdict (which can name
  merge-tap *or* native) and the DI/horizontal pairwise `IssueCounts` resolution
  that may displace it — so one tag would lose which question was asked, which
  is the column K70-1 turns on. Recorded in the RFC decision log.
- **Not absorbed**, per the owner's recorded call on #686: gating
  `fixtures_*_nondefault` on `!is_noop`. The pairing already neutralises a true
  no-op; only the label needed the richer signature. The Interpretation now says
  which of the two does that work.
- `rate` is in the enriched signature even though **no validator reads it** —
  the round-7 review's `belt_flow` citation is wrong, those sites read
  `ItemFlow::rate` off the solver, and `PlacedEntity::rate`'s own doc says so.
  It is in as a provenance signal (a different stamp means a different
  lane-family decision) and is documented as that, not as a validation
  difference.
- **One out-of-scope comment fix**, taken because it is a trap in the file being
  instrumented: the comment above `DirectInsertionCandidate`'s `Search` guard
  claimed the default claim order is `Upstream`. It is `Downstream` —
  `DiClaimOrder`'s `#[default]`, flipped at RFC-059's sim close-out and never
  updated here. Comment-only; found while writing oracle gap (e), which stands
  either way.
- Seven **oracle gaps** are recorded in the RFC decision log rather than fixed
  here. The load-bearing one is the refusal-attribution gap above; the others
  are the missing counts where no mechanism ran, `ErrorKinds` existing only on
  the merge-tap path, `not-run` rows carrying no reason (the gates are
  conjunctions of booleans at the call site), DI's internal claim-order race
  being invisible, the scoreboard being per selection *call*, and — the one
  that bites a shadow diff — a LOSING candidate's nested selection being
  truncated out of the stream before any reader sees it, so "no nested block"
  is not evidence that none ran.

### Finding: the e2e harness does not run production's candidate set

Decoding the snapshot gave a *different deciding stage* than the census
reported for the same fixture — `best-accepted` from the suite,
`best-error-free` from default options. Cause: `LayoutOptions::default()` sets
`cell_composition: Candidate` (`bus/layout.rs:241`), but `run_e2e` spells the
field as `cell_composition: Default::default()` (`tests/e2e.rs:355`) — the
*enum's* default, `Off` (`bus/cells/mod.rs:24`) — sitting next to a
`..Default::default()` that would have given `Candidate`. The line dates to
RFC-051 Phase A (`5090da99`), when `Off` *was* the struct default; Phase B
flipped the struct and the harness kept resolving to the enum.

With cells off only native produces, so `clean_flags` is skipped by its own
`n_layouts > 1` guard and `best-accepted` decides — the suite exercises a
precedence stage the census slice never reaches, and vice versa.

**Not fixed here.** Flipping it changes the candidate set under every
regression test — a selection change, not an additive Phase-0b one. Recorded in
the RFC decision log with both consequences: the Phase-0c parity corpus must
range over fixture × machine tier × **option set**, and suite greenness says
nothing about the cell-composed arm's production exposure.

## Models / contracts touched

`crates/core/src/trace.rs` gains two events and two closed enums (both cross the
WASM boundary as string unions). `docs/rfc-070-one-selection-loop.md` decision
log gains the Phase-0b entries. No validator severity, category or
selection-participation change, so `docs/validator-trust.md` is deliberately
untouched.

---

### DIFF
--- a/crates/core/src/bus/decomposition_search.rs
+++ b/crates/core/src/bus/decomposition_search.rs
@@ -937,6 +937,12 @@ struct CandidateRun {
     /// the unactionable "no decomposition candidate produced a layout"
     /// (observability gap found debugging rfc-build-quality Phase 2).
     error: Option<String>,
+    /// `produce()` panicked and `catch_unwind` swallowed it. Read only by
+    /// the RFC-070 scoreboard, which must not conflate a panic with an
+    /// ordinary refusal — `error` carries a sentinel string in that case
+    /// and string-matching it would be a latent bug the moment the tag is
+    /// reworded.
+    panicked: bool,
     events: Vec<crate::trace::TraceEvent>,
 }
 
@@ -944,7 +950,7 @@ impl CandidateRun {
     /// A candidate that wasn't tried (e.g. gating predicate was false).
     /// No outcome, no events; the winner-selection code skips it.
     fn skipped(_name: &str) -> Self {
-        Self { outcome: None, events: Vec::new(), error: None }
+        Self { outcome: None, events: Vec::new(), error: None, panicked: false }
     }
 }
 
@@ -984,7 +990,7 @@ where
             None
         }
     };
-    CandidateRun { outcome, events, error }
+    CandidateRun { outcome, events, error, panicked: false }
 }
 
 /// Like `run_candidate` but wraps the produce call in `catch_unwind`.
@@ -1002,6 +1008,7 @@ where
     let mut events = crate::trace::peek_events_since(start);
     crate::trace::truncate_events(start);
     let mut error = None;
+    let mut panicked = false;
     let outcome = match result {
         Ok(Ok(layout)) => {
             let score = score_layout(&layout, solver_result);
@@ -1022,6 +1029,7 @@ where
         }
         Err(_) => {
             error = Some("panicked (caught)".to_string());
+            panicked = true;
             events.push(crate::trace::TraceEvent::DecompositionCandidateScored {
                 name: name.to_string(),
                 density: 0.0,
@@ -1034,7 +1042,122 @@ where
             None
         }
     };
-    CandidateRun { outcome, events, error }
+    CandidateRun { outcome, events, error, panicked }
+}
+
+/// RFC-070 Phase 0b (#689 W1b): one candidate's row on the selection
+/// scoreboard.
+///
+/// Purely observational — no decision reads it. A row starts from the
+/// candidate's `CandidateRun` (outcome + soft score, mechanism 1) and is
+/// then FILLED IN by whichever verdict mechanism happens to touch that
+/// candidate, **at the site where the decision path already computed the
+/// number**. Nothing here recomputes: a scoreboard that re-ran
+/// `validate()` on its own could disagree with the value the decision
+/// actually used, and an oracle that disagrees with the thing it is
+/// oracling is worse than no oracle.
+///
+/// The price of that discipline is `None`s — a candidate that no
+/// comparison needed carries no issue counts at all, because on that call
+/// nothing computed any. That is a genuine hole in what this can see, and
+/// it is recorded as a hole rather than papered over with a fresh
+/// validation pass.
+struct CandidateVerdict {
+    name: &'static str,
+    outcome: crate::trace::SelectionCandidateOutcome,
+    reason: Option<String>,
+    score: Option<f64>,
+    accepted: Option<bool>,
+    accepted_reason: Option<String>,
+    counts: Option<IssueCounts>,
+    counts_source: Option<&'static str>,
+    kinds: Option<ErrorKinds>,
+}
+
+impl CandidateVerdict {
+    fn from_run(name: &'static str, run: &CandidateRun) -> Self {
+        use crate::trace::SelectionCandidateOutcome as Outcome;
+        // `panicked` is a field, not a string match on `error` — see its
+        // doc on `CandidateRun`. Order matters: a panic also sets `error`.
+        let outcome = match (&run.outcome, run.panicked, &run.error) {
+            (Some(_), _, _) => Outcome::Produced,
+            (None, true, _) => Outcome::Panicked,
+            (None, false, Some(_)) => Outcome::Refused,
+            (None, false, None) => Outcome::NotRun,
+        };
+        Self {
+            name,
+            outcome,
+            reason: run.error.clone(),
+            score: run.outcome.as_ref().map(|(_, s)| s.score),
+            accepted: run.outcome.as_ref().map(|(_, s)| s.accepted),
+            accepted_reason: run
+                .outcome
+                .as_ref()
+                .and_then(|(_, s)| s.accepted_reason.clone()),
+            counts: None,
+            counts_source: None,
+            kinds: None,
+        }
+    }
+}
+
+/// The seven candidate rows, **indexed exactly like the `candidates`
+/// array** in `select_best_decomposition` (`NATIVE_IDX` = 0 through
+/// `H_IDX` = 6), so a recorder can key off the same index constants the
+/// decision uses.
+struct Scoreboard([CandidateVerdict; 7]);
+
+impl Scoreboard {
+    /// Record the `IssueCounts` a comparison site just computed.
+    ///
+    /// FIRST WRITE WINS. Native's counts are computed by both pairwise
+    /// comparisons and DI's by both `di_choice` and `scoped_choice`; the
+    /// repeats are the same deterministic `count_issues` call on the same
+    /// layout, so keeping the first preserves the provenance of the site
+    /// that first needed the number instead of overwriting it with a
+    /// later duplicate.
+    fn record_counts(&mut self, idx: usize, counts: IssueCounts, source: &'static str) {
+        let row = &mut self.0[idx];
+        if row.counts.is_none() {
+            row.counts = Some(counts);
+            row.counts_source = Some(source);
+        }
+    }
+
+    /// Record the `ErrorKinds` the merge-tap decision just computed.
+    /// First write wins, for the same reason as `record_counts`.
+    fn record_kinds(&mut self, idx: usize, kinds: ErrorKinds) {
+        let row = &mut self.0[idx];
+        if row.kinds.is_none() {
+            row.kinds = Some(kinds);
+        }
+    }
+
+    /// Emit one event per candidate — all seven, including the ones that
+    /// never ran. "Which candidates were even eligible" is half the
+    /// question RFC-070 asks, and a row that is absent from the stream
+    /// cannot answer it (`docs/validator-reporting.md`: one positioned
+    /// record per instance, never a count in a message).
+    fn emit(&self) {
+        for row in &self.0 {
+            crate::trace::emit(crate::trace::TraceEvent::SelectionCandidateEvaluated {
+                name: row.name.to_string(),
+                outcome: row.outcome,
+                reason: row.reason.clone(),
+                score: row.score,
+                accepted: row.accepted,
+                accepted_reason: row.accepted_reason.clone(),
+                errors: row.counts.map(|c| c.errors),
+                selection_warnings: row.counts.map(|c| c.warnings),
+                layout_warnings: row.counts.map(|c| c.layout_warnings),
+                counts_source: row.counts_source.map(str::to_string),
+                contamination_errors: row.kinds.map(|k| k.contamination),
+                starvation_errors: row.kinds.map(|k| k.starvation),
+                structural_errors: row.kinds.map(|k| k.structural),
+            });
+        }
+    }
 }
 
 /// Run candidates and pick the winner.
@@ -1212,6 +1335,21 @@ pub fn select_best_decomposition(
     // applies unchanged (every non-Pooled and every Native-clean case).
     const NATIVE_IDX: usize = 0;
     const MERGE_TAP_IDX: usize = 3;
+
+    // RFC-070 Phase 0b scoreboard. Built here — after every candidate has
+    // run, before the first verdict mechanism fires — so the mechanism
+    // sites below can record into it as they compute. Index-aligned with
+    // the `candidates` array further down.
+    let mut board = Scoreboard([
+        CandidateVerdict::from_run("native", &native_run),
+        CandidateVerdict::from_run("k1-shape-fix", &k1_run),
+        CandidateVerdict::from_run("size-split-2", &split_run),
+        CandidateVerdict::from_run("merge-tap", &merge_tap_run),
+        CandidateVerdict::from_run("cell-composed", &cells_run),
+        CandidateVerdict::from_run("direct-insertion", &di_run),
+        CandidateVerdict::from_run("horizontal-stack", &horizontal_run),
+    ]);
+
     let merge_tap_choice: Option<usize> = merge_tap_run.outcome.as_ref().map(|(mt_layout, _)| {
         let start = crate::trace::peek_events_len();
         let mergetap_kinds = classify_errors(mt_layout, solver_result);
@@ -1220,6 +1358,10 @@ pub fn select_best_decomposition(
             .as_ref()
             .map(|(l, _)| classify_errors(l, solver_result));
         crate::trace::truncate_events(start);
+        board.record_kinds(MERGE_TAP_IDX, mergetap_kinds);
+        if let Some(n) = native_kinds {
+            board.record_kinds(NATIVE_IDX, n);
+        }
         match native_kinds {
             Some(n) if mergetap_kinds.quality_key() < n.quality_key() => MERGE_TAP_IDX,
             Some(_) => NATIVE_IDX,
@@ -1264,6 +1406,8 @@ pub fn select_best_decomposition(
         let di_counts = count_issues(di_layout, solver_result);
         let nat_counts = count_issues(nat_layout, solver_result);
         crate::trace::truncate_events(start);
+        board.record_counts(DI_IDX, di_counts, "di-vs-native");
+        board.record_counts(NATIVE_IDX, nat_counts, "di-vs-native");
         // Component-wise, NOT lexicographic — see `IssueCounts`.
         let strictly_better_issues = di_counts.strictly_better_than(&nat_counts);
         // Equal on every issue channel: DI must then be strictly denser
@@ -1296,6 +1440,8 @@ pub fn select_best_decomposition(
             let hs_counts = count_issues(hs_layout, solver_result);
             let nat_counts = count_issues(nat_layout, solver_result);
             crate::trace::truncate_events(start);
+            board.record_counts(H_IDX, hs_counts, "horizontal-vs-native");
+            board.record_counts(NATIVE_IDX, nat_counts, "horizontal-vs-native");
             let strictly_better_issues = hs_counts.strictly_better_than(&nat_counts);
             // v1 deliberately has NO equal-issues-and-denser arm (unlike
             // `di_choice`): measured on the first full-suite run, that
@@ -1369,8 +1515,11 @@ pub fn select_best_decomposition(
         && n_layouts > 1
     {
         let start = crate::trace::peek_events_len();
-        let flags = tier_outcomes.map(|o| {
-            o.map(|(l, score)| {
+        // `from_fn` rather than `[T; N]::map` only so the RFC-070
+        // scoreboard can key the row it records by the same index the
+        // flag lands on; the traversal and its result are identical.
+        let flags: [Option<(bool, usize)>; 7] = std::array::from_fn(|idx| {
+            tier_outcomes[idx].map(|(l, score)| {
                 // (error-free, warning key). The key orders the
                 // error-free tier below: RFC-060 made the refusal path
                 // multi-candidate (cells, DI, horizontal), and a
@@ -1391,12 +1540,44 @@ pub fn select_best_decomposition(
                         // Selection-scoped count — see
                         // `validate::selection_warning_count`.
                         let warnings = crate::validate::selection_warning_count(&issues);
+                        board.record_counts(
+                            idx,
+                            IssueCounts {
+                                errors,
+                                warnings,
+                                layout_warnings: l.warnings.len(),
+                            },
+                            "clean-flags",
+                        );
                         (
                             score.accepted && errors == 0,
                             warnings + l.warnings.len(),
                         )
                     }
-                    Err(_) => (false, usize::MAX),
+                    Err(e) => {
+                        // `validate()` returns `Err` CARRYING the issues, and
+                        // this arm is reached iff at least one is an Error.
+                        // The decision reads only "not clean" — but throwing
+                        // the counts away here would have the scoreboard
+                        // report a blank row for precisely the candidates
+                        // that failed hardest. Same issue list, same
+                        // predicates as the `Ok` arm; the source tag records
+                        // that the decision itself did not read them.
+                        board.record_counts(
+                            idx,
+                            IssueCounts {
+                                errors: e
+                                    .issues
+                                    .iter()
+                                    .filter(|i| i.severity == crate::validate::Severity::Error)
+                                    .count(),
+                                warnings: crate::validate::selection_warning_count(&e.issues),
+                                layout_warnings: l.warnings.len(),
+                            },
+                            "clean-flags(unclean)",
+                        );
+                        (false, usize::MAX)
+                    }
                 }
             })
         });
@@ -1586,20 +1767,42 @@ pub fn select_best_decomposition(
     // the DI/horizontal pairwise resolution — so neither scoped
     // candidate can shadow the other; the merge-tap non-shadowing
     // structure is unchanged.)
-    let winner_idx = match merge_tap_choice {
-        Some(MERGE_TAP_IDX) => Some(MERGE_TAP_IDX),
-        Some(_) => scoped_choice.or(Some(NATIVE_IDX)),
-        None => scoped_choice,
+    //
+    // RFC-070 Phase 0b: each arm is TAGGED with the `SelectionStage` that
+    // produced it. The chain is unchanged — `Some(_) => scoped_choice
+    // .or(Some(NATIVE_IDX))` is spelled as a match on `scoped_choice`
+    // purely so the two answers get different stage tags (a scoped
+    // candidate displacing native is a different decision from native
+    // surviving merge-tap, and the oracle must not merge them).
+    use crate::trace::SelectionStage;
+    let winner_idx: Option<(usize, SelectionStage)> = match merge_tap_choice {
+        Some(MERGE_TAP_IDX) => Some((MERGE_TAP_IDX, SelectionStage::MergeTap)),
+        Some(_) => match scoped_choice {
+            Some(i) => Some((i, SelectionStage::ScopedPairwise)),
+            None => Some((NATIVE_IDX, SelectionStage::MergeTap)),
+        },
+        None => scoped_choice.map(|i| (i, SelectionStage::ScopedPairwise)),
     }
-    .or(best_error_free_idx)
-    .or(best_accepted_idx)
+    .or(best_error_free_idx.map(|i| (i, SelectionStage::BestErrorFree)))
+    .or(best_accepted_idx.map(|i| (i, SelectionStage::BestAccepted)))
     // DI IS admitted here, deliberately: `di_choice` returns `None` when
     // native produced nothing (see its guard), precisely so `cell-composed`
     // and DI compete in this ranking rather than DI auto-winning a refusal.
     // `ranking_len` is what excludes DI in the native-SUCCEEDED case.
-    .or_else(|| candidates[..ranking_len].iter().position(|(o, _, _)| o.is_some()));
+    .or_else(|| {
+        candidates[..ranking_len]
+            .iter()
+            .position(|(o, _, _)| o.is_some())
+            .map(|i| (i, SelectionStage::FirstProduced))
+    });
 
-    let Some(idx) = winner_idx else {
+    let Some((idx, stage)) = winner_idx else {
+        // Every candidate failed. The scoreboard still goes out — WHY each
+        // of the seven produced nothing is exactly what a refusal
+        // investigation needs — but no `SelectionDecided` follows it,
+        // because there is no winner and the stage enum is deliberately
+        // closed over the five ways a winner CAN be picked.
+        board.emit();
         let details: Vec<String> = candidates
             .iter()
             .map(|(_, _, name)| name.to_string())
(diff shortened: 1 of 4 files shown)


storkme:
Round 1 absorbed. Four findings taken, one taken-in-half, one refuted with receipts, both nits answered.

## Findings

- **[minor, 3/3] `scoped_choice` never calls `record_counts`; the doc claims it does** — **Absorbed.** The code is as described and the doc sentence was wrong to imply recording where it meant *computing*. Fixed in all three places the claim surfaces: `record_counts`'s doc, the `counts_source` field doc in `trace.rs`, and the census's printed header. All three now say the same thing — `from` names the site that **first computed** a candidate's counts, **not** the site that decided; recording is first-write-wins and several sites compute the same numbers, so a later deciding tier can inherit an earlier tag. What is *not* affected is the recorded value: every repeat is the same deterministic `count_issues` call on the same layout, so only the label can differ, never the number. For "who decided", the terminal event's stage is the answer, and that is now stated at each site. No code change, because recording at `scoped_choice` would overwrite nothing (both its inputs are on the board by construction) and would only relabel. Data note: the two-scoped-candidates-both-beat-native arm did not fire anywhere on the census slice — on `advanced-circuit@5` `di_choice` returned `None` (DI tied native at `0/2/0`), so the win came through `horizontal_choice` and `horizontal-vs-native` is the correct attribution there.

- **[minor, 3/3] the block walker has no defense for an unterminated selection** — **Absorbed.** Correct: flushing only on `SelectionDecided` leaves an all-failed selection's rows open to absorb the next block's. Added the boundary the emitter actually guarantees: `Scoreboard::emit` walks its rows in index order and index 0 is always `native`, so a `native` row arriving while rows are pending starts a new block and flushes the old one as unterminated. Latent today for the reason you give (and one more: a nested all-failed selection makes its candidate refuse, so its events are truncated and never replayed) — taken anyway, because the walker's contract should not depend on the reachability argument holding forever.

- **[minor, 2/3] the `from` column contradicts its stated claim** — **Absorbed**, same fix as the first finding; the header no longer says the counts are "the counts the DECISION computed".

- **[minor, 2/3] decision-log entry (a) overclaims the merge-tap blanks** — **Absorbed, and this one was a real error, not just wording.** You are right that a merge-tap-decided fixture necessarily has native *produced-but-unaccepted*, so `di_choice`/`horizontal_choice` run whenever couplings/a dual-input row exist and record native's counts before `clean_flags` is short-circuited. The `ec@30` observation (no counts on either row) was true; the generalisation from it was not — there, DI **and** horizontal both refused, which is the actual reason the rows are blank. Corrected in three places, since the same sentence had been copied into all of them: the RFC decision log, the census's printed Interpretation paragraph, and the PR body. This is the "claim broader than the evidence" class the campaign exists to reduce, so thank you — it was found by reading, which is exactly what the instrument cannot do for itself.

- **[minor, 2/3] "one row per fixture" overstates; and the test has zero assertions** — **First half absorbed, second refuted-as-disclosed.** The summary now emits a row for **every** fixture, including the ones that never reach a decision: `SKIP (no solve)`, `no selection events`, and `NO WINNER (all candidates failed)` each get their own row rather than silently contributing none. A summary that drops its failures is a shape this repo has been bitten by repeatedly (`docs/validator-reporting.md`), so the fix is the right one even though no fixture on the current slice takes those paths. On the second half: `selection_scoreboard_census` is a print-only `#[ignore]`d diagnostic **by design**, stated in the `#[ignore]` reason, the module doc, and the PR body — the same point was raised against the sibling census on #686 and answered the same way there. Its output is checked by a human reading it; making it assert would require freezing a baseline, which is explicitly Phase 0c's deliverable and explicitly out of this PR's scope.

- **[minor, 1/3] doc drift: `decomposition_search.rs:996` still names `Upstream` as the DI default** — **Refuted, with receipts.** Line 996 is `run_candidate`'s `CandidateRun { outcome, events, error, panicked: false }` constructor, not a DI comment. `grep -n Upstream crates/core/src/bus/decomposition_search.rs` returns exactly two hits at the merged head: line 217, inside the comment **this PR corrected** (which now names `Downstream` as the default and records that the comment said `Upstream` until 2026-08-21), and line 226, the `arm(DiClaimOrder::Upstream)` construction, which is a call and not a claim. In `di_cell.rs` every `Upstream` mention is contextually correct and none asserts a default: `:147` describes how the placer interprets `Search` ("the placer, which walks one order per call, reads this as `Upstream`") and `:152` describes `Prioritized`'s fallback ordering. The authoritative statement is `#[default] Downstream` at `di_cell.rs:115-116`, reached via `LayoutOptions::default()`. The `{cover major}` literal in the finding text looks like a template artifact rather than a citation. No change.

## Nits

Both answered: the `*` winner marker does **not** shift the row — the header is `"   {:<18} …"` (3 + 18 = 21 columns before `outcome`) and a data row is `"  {}{:<18} …"` (2 + 1 + 18 = 21), which is why `produced` sits under `outcome` in the pasted output; and the `clean-flags(unclean)` arm deriving counts the decision never read is deliberate and disclosed by that exact tag — dropping them would blank the scoreboard rows of precisely the candidates that failed hardest, which is the failure mode this instrument exists to avoid.

## Re-verification after absorption

- `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` → exit 0 (1182 passed / 0 failed / 74 ignored, unchanged).
- `cargo clippy --workspace -- -D warnings` → exit 0; `cargo clippy --test check_firing_census -- -D warnings` → exit 0.
- Both census diagnostics re-run: the scoreboard's per-fixture data is byte-identical to the run in the PR body (only the header/legend text and the summary's new coverage changed).

storkme:
Round 2 absorbed. Five taken, two refuted with receipts. Every claim below was checked at source before acting on it, and the new test's discrimination was **executed, not assumed**.

## Findings

- **[major, 3/3] the Phase-0b oracle has zero CI assertion coverage** — **Absorbed, and this was the right call.** Everything in the file was `#[ignore]`d, so a broken stage tag, a dropped row, or an inverted `from_run` deduction would have shipped green with a human-run diagnostic as its only reader — the exact shape `docs/validator-reporting.md` is about. Added `selection_scoreboard_contract`, **not ignored**, on `iron-gear-wheel@10`: all seven slots emit a row, in canonical order, all rows precede their `SelectionDecided`, exactly one terminal event, the winner is one of that block's own rows and its row says `Produced`, `cell-composed` is `Produced` and `k1-shape-fix` is `NotRun` (so `not-run` and `refused` cannot collapse), and the deciding stage is `best-error-free`. It pins the CONTRACT, not tile geometry — golden hashes already own that.

  **Discrimination check, run three times rather than asserted:**
  | break | result |
  |---|---|
  | mis-tag the error-free arm as `BestAccepted` | FAILS on the stage assertion |
  | emit `&self.0[1..]` (drop the `native` row) | FAILS on the order assertion, printing the six-name list |
  | swap `cells_run`/`di_run` in the index list | FAILS on the new alignment assert, naming slot 4 |

  All three restored, re-run green. Note the stage assertion's message deliberately names the *two* readings of a future failure (tagging broke vs. the candidate set changed) and points at the `cell-composed` assertion that discriminates between them — a pin that cannot say why it failed is a pin that gets deleted.

- **[major, 3/3] four hand-maintained copies of the candidate order** — **Absorbed.** `CandidateRun` now carries its own `name`, captured at the call that ran it, so a run is self-identifying; `CANDIDATE_ORDER` is the single source; `Scoreboard::from_runs` asserts each run's name against its slot; `tier_outcomes` is derived from the same checked list rather than being a fourth literal; and `candidates` takes its names from the runs (not literals) with a post-construction check, which is what makes that assert discriminate — with hardcoded labels a swapped run would still have read "correct". Break C above is the receipt. The asserts compare compile-time constants, so they cannot fire for some inputs and not others: either the first test that runs panics, or none can.

- **[minor, 3/3] block-flushing heuristic assumptions unasserted** — **Absorbed.** Assumption (a) (every block starts with `native`) is now the *checked* `CANDIDATE_ORDER` contract above rather than a comment, and the contract test pins the row order in CI. Assumption (b) (the outer selection is the last block) is now checked where it is used: the census verifies the winner is among its own block's rows and prints a loud `!! BLOCK PAIRING BROKEN` marker if not — a print-only diagnostic reports loudly rather than panicking — while the contract test enforces the same invariant in CI for a real fixture.

- **[minor, 1/3] the nested banner also labels the outer block** — **Absorbed.** Correct, and it landed exactly where a reader is told to look. Now printed only for `n < blocks.len()-1`, and it names which index is the outer one.

- **[minor, 1/3] the e2e `cell_composition` gap remains open** — **Agreed and recorded, no change.** It is the documented campaign call, not an oversight: fixing it changes the candidate set under every regression test, i.e. a selection change, which is out of scope for an additive Phase-0b PR. The RFC decision log carries it with both consequences (the Phase-0c parity corpus must range over fixture × machine tier × option set; suite greenness says nothing about the cell-composed arm), and it is in the PR body and in #689.

- **[minor, 1/3] the failure path emits with the sink detached** — **Refuted, with receipts.** The premise is that the sink "was detached at the top and is never restored before the early `return Err`". It is restored 210 lines before that path: `swap_sink(None)` at `decomposition_search.rs:1199`, `swap_sink(Some(sink))` at **`:1604`**, failure-path `board.emit()` at **`:1814`**, `return Err` at `:1823`. Those are the only two `swap_sink` calls in the function (`grep -n swap_sink` returns exactly them), and no branch between :1604 and :1814 detaches it. So a failure-path scoreboard reaches a streaming consumer exactly like a success-path one, and the asymmetry the finding builds on — "streaming sees nothing for a failure" — does not exist. No doc note added, because documenting it would have recorded a false behaviour. (The real asymmetry, already documented, is that a failure emits rows but no `SelectionDecided`, since the stage enum is closed over the five ways a winner *can* be picked.)

- **[minor, 1/3] `rate` in `EntitySignature` makes a quiet variant falsely noisy** — **Verified, and you are right about the premise; absorbed as a redefinition rather than a removal.** Two reviews have now claimed opposite things about this, so it was settled by counting: all **21** `.rate` reads across `validate/` and `connectivity.rs` are on solver `ItemFlow`s (`o`/`i`/`f`/`out`/`inp`/`sur`/`flow`/`ext`); **none** is on a `PlacedEntity`. The three `belt_flow.rs` lines #686 round 7 cited as proof that the validator reads `e.rate` are `:684` → `e.carries`, `:1675` → `e.carries`, `:3286` → `build_ug_pairs(layout)`. So #686's adjudication was wrong about `rate` (right about `carries`), and this finding is right.

  Decision: **keep `rate`, redefine the label.** A differing stamp means the pipeline made a different lane-family decision on the way to the same tiles, which is worth not calling identical — but the label must then say what it means, so every site that states it now reads **"a no-op is tiles AND stamps identical, not validator-identical"**: the `EntitySignature` doc (with the receipts above), the printed no-op header, and the closing Interpretation. That is a strictly *stricter* test, so it can only under-report no-ops, never over-report them — the direction that matters, since the counter exists to stop a quiet variant being read as an evaluated one. Measured, the redefinition changes nothing on this fixture set: ratios identical to the pre-`rate` run (di-off 5/6, di-forced 5/6, cells-off 6/6, hs-off 5/6, partitioned 4/6). Dropping `rate` was the alternative and was rejected — it would make the label mean less while matching neither the recorded #675 follow-up nor the provenance question the counter answers.

## Nits

The `SelectionStage::MergeTap` double meaning is a deliberate design call and is already recorded in the RFC decision log with exactly the caveat you name ("the tag is not the same as merge-tap won") — that is why the chain is spelled as a match rather than an `.or()`, so the *other* answer gets `ScopedPairwise`. The enum-ordering hazard is real in principle and inert here (serde `tag`/`content` string discriminators, tsify string unions — both verified in the generated `.d.ts`); noted, no change. The `di-vs-native`-sourced row under a `MergeTap` stage is the provenance-vs-decision distinction round 1 corrected the wording for. The "outer selection is the last block" assertion you suggested is now implemented in both places (census marker + CI test).

## Re-verification

- `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` → **exit 0**, 1183 passed / 0 failed / 74 ignored (one more than before: the new contract test).
- `cargo clippy --workspace -- -D warnings` → exit 0; `cargo clippy --test check_firing_census -- -D warnings` → exit 0.
- `wasm-pack build crates/wasm-bindings --target web --out-dir "$(pwd)/web/src/wasm-pkg"` → exit 0.
- Both diagnostics re-run: per-fixture scoreboard data unchanged; no-op ratios unchanged.

storkme:
Round 3 absorbed — five taken, one adjudicated as designed. No majors this round, and each finding below was checked at source before acting.

## Findings

- **[minor, 1/3] the all-refused path still zips a hand-typed 7-slot tuple** — **Absorbed, and you found the last one.** Confirmed at source: the tuple at `:1680` and its positional `.zip([&native_err, …])` at `:1880`, every element the same type, so a reorder compiled silently. Now both sides of that message derive from the checked pair: names from `CANDIDATE_ORDER` itself, reasons from `run_refs.map(|r| r.error.clone())` — the same list `Scoreboard::from_runs` checks each run's own name against. There is no positional literal left in the function.

- **[minor, 1/3] two unconditional `assert_eq!`s on every WASM solve** — **Absorbed.** The point about inverting this function's own degradation philosophy — twenty lines from arms that run under `catch_unwind` precisely so a panic degrades instead of aborting — is a fair one. Both are now `debug_assert_eq!`. **Coverage is unchanged**: they compare compile-time constants, so a violation cannot depend on input, and every `cargo test` run and every CI job builds in debug — the first debug-built call catches it. Release and WASM now pay nothing and cannot panic a browser solve over a code-level ordering mistake.

  One consequence worth naming, because it nearly bit: a `debug_assert`-only reference reads as **dead code in release**. This repo already carries one instance — `ghost_occupancy::is_claimed` is `warning: never used` under `cargo clippy --release --lib`, because its only non-test uses are `debug_assert!`s in `ghost_router.rs` — invisible to CI, which lints debug only. `CANDIDATE_ORDER` would have joined it, so the refusal message above now uses it for real rather than only inside assertions. Verified: release clippy exits 0 with `is_claimed` as the only "never used" warning, and `CANDIDATE_ORDER` unflagged. The `is_claimed` artifact is pre-existing and left alone as out of scope.

- **[minor, 3/3] the contract test couples four engine internals with no self-distinguishing message** — **Absorbed.** The stage assertion already named both readings; that treatment now extends to the other three. `decided.len()==1` says a nested selection is a *correct engine* this pin does not cover, and tells you to confirm it by looking for a second full seven-row block before re-pointing the test at the outer one. The row-order assertion sends you to `CANDIDATE_ORDER` first: if the engine's list changed, this test's independently-written list is merely stale and the Phase-0 baseline needs re-taking; if it did not, the instrumentation broke. The event-order assertion names which sibling assertion discriminates between the two (a nested selection adds a terminal; a misplaced `board.emit()` does not). No structural change — as you say, it is a pin, and a pin's job is to fail with the reason attached.

- **[minor, 1/3] the `noncomparable` header invites the wrong reading** — **Absorbed.** Verified the behaviour first: a refused variant `continue`s before the comparable/non-comparable branch, so it never reaches that tally at all. The header now says so explicitly — "+N with no default" counts builds that SUCCEEDED but whose fixture's default did not build, and a variant's own refusals live in the `refusals:` summary at the top, so the column is not "all builds with a failing default".

- **[minor, 1/3] "five places" should be six** — **Absorbed.** Counted: `count_issues` is called at `:1471`, `:1472`, `:1504`, `:1505`, `:1536`, `:1537` — two in `di_choice`, two in `horizontal_choice`, two in `scoped_choice`'s both-Some arm. The doc now says six and names where they are.

- **[minor, 3/3] `counts_source` provenance can mismatch the deciding mechanism** — **Adjudicated as designed; one sentence added.** The observation is correct and is exactly what the encoding means: a row is a per-candidate summary of everything *any* mechanism computed about that candidate, not the record of one decision — so on a Pooled solve where DI also ran, native's row carrying `di-vs-native` counts and merge-tap kinds is two mechanisms reporting on one candidate, which is the question the row answers. The decision authority is `SelectionDecided::stage`, which the row deliberately does not duplicate; round 1 corrected the wording that let the two be read as one, and the RFC decision log carries it. Added one clarifying sentence at the field doc so a reader meets the mismatch where they meet the field, rather than only in the log.

## Nits

The walker invariants are the ones now checked rather than assumed (`CANDIDATE_ORDER` in debug + the CI contract test's row-order assertion + the census's `!! BLOCK PAIRING BROKEN` marker). On the soft-score margin: the contract test asserts the deciding STAGE and the winner, not a score value or a density weight — a weight refactor that leaves native winning on the error-free tier does not move it, and one that changes which tier decides is a selection change that should show up in the Phase-0 baseline. That is the intended sensitivity.

## Re-verification

- `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` → **exit 0**, unpiped: 1183 passed, 0 failed, 74 ignored across 27 binaries — unchanged from round 2, with `selection_scoreboard_contract ... ok` in the run.
- **Discrimination re-checked after the message rewrite** (one break + restore, as asked): mis-tagging the error-free arm still fails the stage assertion, `left: BestAccepted / right: BestErrorFree`; restored and re-run green.
- `cargo clippy --workspace -- -D warnings` → exit 0; `--test check_firing_census -- -D warnings` → exit 0; `cargo clippy --release --lib` → exit 0 (see the dead-code note above).
- `wasm-pack build crates/wasm-bindings --target web --out-dir "$(pwd)/web/src/wasm-pkg"` → exit 0.
- Both diagnostics re-run → exit 0: per-fixture scoreboard data unchanged (same six winners and stages), no-op ratios unchanged (di-off 5/6, di-forced 5/6, cells-off 6/6, hs-off 5/6, partitioned 4/6), category table unchanged.

storkme:
**Round 4 adjudicated post-merge (closing-round protocol, declared at round 3).**

**[major] "`std::array::from_fn` may not compile on this toolchain": refuted with receipts.** `std::array::from_fn` has been stable since Rust 1.63; `rust-toolchain.toml` pins 1.95.0; and the claim is falsified by execution on the exact head it was raised against — local `cargo test --no-fail-fast` exit 0 (1183 passed) and the `rust` CI check green on `08d457ba` (13m30s). This is the review bot's known false-compile-blocker class.

**[minor, absorbed as recorded follow-up] contract test covers only the `BestErrorFree` stage.** True and deliberate for the first pin; extending the contract to a merge-tap-deciding and a scoped-pairwise-deciding fixture is recorded on #689 as a W1c-adjacent hardening (the Phase-0c baseline work adds fixtures anyway — that's the natural place).

**Refuted-as-decided (each a recycle of a round-3 adjudication):** the contract-pin false-red framing (round 3 #3 — the assertion messages now name both readings and the discriminating follow-up); the debug_assert release-silence items ×2 (round 3 #2 — the trade is documented at the site: the oracle and census run under `cargo test`/CI, which build debug; production panic added blast radius, not coverage); the `counts_source` provenance nuance (round 3 #6 — the field doc states `from` = "first computed", never "decided"; `SelectionDecided.stage` is the authority).

**Doc nits (fixed on main in a follow-up touch):** the decision-log line anchors (1659/1881, not 1604/1814 — arithmetic drift only, claim unchanged) and the two residual "bit-identical" phrases in the census header that predate the redefined no-op label.

PR #693, opened 2026-08-21
diag(w1a): report-only meter tripwire over the e2e/census corpus

storkme:
## Intent

W1a of the RFC-070 campaign (tracking issue #689): wire `crates/meter/`
(RFC-054's fast, ~seconds native throughput meter) into an automated
diagnostic for the first time. Before this PR nothing ran the meter
except ad hoc `cargo run --example`. This adds a report-only tripwire
over a slice of the e2e/census fixture corpus so an engine change gets a
seconds-scale below-plan regression signal instead of waiting on a
headless-Factorio sim run.

## Scope

- **In:** 13 fixtures — the G2 census's 6 (`check_firing_census.rs`'s
  `fixtures` table) plus 7 more from the e2e tier ladder
  (`crates/core/tests/e2e.rs`). One new test file
  (`crates/meter/tests/e2e_tripwire.rs`) that builds each fixture
  directly through `solver`/`layout`/`validate`/`blueprint` (not by
  shelling out to the gitignored `crates/core/examples/`), feeds the
  resulting blueprint+manifest to the meter exactly as a disk fixture
  would be, and prints a planned-vs-measured scoreboard. A committed
  baseline (`e2e_tripwire_baseline.json`) of today's per-fixture
  deficits, blessed with the CI zone-cache pin so it's reproducible
  across hosts.
- **Out:**
  - The mirror-flag fluid-port gap (`reflect_port` unimplemented,
    `meter/factory.rs` ~192-210) — known-open, explicitly out of scope
    per the task; none of these fixtures exercise `mirror: true` ports.
  - The full stress/mega corpus (dozens of fixtures in
    `cell_composition.rs`) — future work; this is the "manageable
    slice" called for.
  - Wiring `check` mode into CI as a required gate. The committed-golden
    `check`/`bless` flow that used to live behind
    `SPAGHETTIO_STRESS_GOLDEN` was deleted 2026-08-15 (#632 B7) — it was
    host-cache-relative and nobody ran it, so every committed golden
    went stale within weeks and produced only false drift signals. This
    baseline is reproducible (blessed with the zone-cache pin), but it's
    a brand-new instrument with zero track record; gating on it now
    would repeat that mistake at PR 1. Report-only stays the default;
    `check` mode exists and is documented for a future PR to promote
    once it's actually been run for a while.
- **Size:** 516 added lines in the test file, but the hard rule is
  "only alarm on below-plan," which needed a fair amount of doc-comment
  discipline to pin down precisely (see Hard rule below) — the
  committed baseline JSON (93 lines) is fixture data and doesn't count
  toward the ~400-line norm per CLAUDE.md.

## The scoreboard (today, on origin/main)

Built with the CI zone-cache pin
(`SPAGHETTIO_ZONE_CACHE_PATH=crates/core/data/sat-zones-ci.bin`) so the
numbers are reproducible:

| fixture | target | planned | measured | delta | flag |
|---|---|---:|---:|---:|---|
| gear10-am1-plate | iron-gear-wheel | 10.000 | 10.000 | +0.0% | |
| ec10-am1-ore | electronic-circuit | 10.000 | 10.000 | +0.0% | |
| **ec30-am2-ore** | electronic-circuit | 30.000 | **0.000** | **-100.0%** | **BELOW PLAN** |
| plastic5-chem-crude | plastic-bar | 5.000 | 5.044 | +0.9% | |
| **ac5-am2-ore** | advanced-circuit | 5.000 | 4.989 | -0.2% | BELOW PLAN (noise-floor) |
| **pu2-am3-ore** | processing-unit | 2.000 | 1.839 | **-8.1%** | **BELOW PLAN** |
| gear10-am2-ore | iron-gear-wheel | 10.000 | 10.000 | +0.0% | |
| **gear20-am2-plate** | iron-gear-wheel | 20.000 | 15.000 | **-25.0%** | **BELOW PLAN** |
| **ec10-am1-ore-yellow** | electronic-circuit | 10.000 | 9.333 | **-6.7%** | **BELOW PLAN** |
| ec20-am2-ore | electronic-circuit | 20.000 | 20.000 | +0.0% | |
| plastic10-chem-petro | plastic-bar | 10.000 | 10.000 | +0.0% | |
| sulfuric5-chem | sulfuric-acid | 5.000 | 16.944 | +238.9% | (not alarmed — see below) |
| lightoil5-chem-cracking | light-oil | 5.000 | 15.000 | +200.0% | (not alarmed — see below) |

Two findings worth flagging to the campaign for follow-up, neither
blocking this PR:

1. **`ec30-am2-ore` measures a genuine, trustworthy 0.000/-100%,
   NOT converged.** Corroborated independently by the engine's own
   validator, which already flags this exact fixture (EC@30 AM2,
   unrestricted belt tier) with 3 `belt-dead-end` **errors** ("Belt at
   (11,22) facing West has no receiver at output tile (10,22) — items
   accumulate with nowhere to go"). Two independent instruments agree
   this fixture is fully jammed. Pre-existing, not introduced here —
   the census fixture (which doesn't force a belt tier) already carried
   this validator error; the meter just confirms it's not a
   false-positive by measuring zero real throughput.
2. **`sulfuric5-chem` and `lightoil5-chem-cracking` measure well
   ABOVE plan (+239%/+200%).** Traced to the solver assigning a
   fractional machine count (`count: 0.1` for sulfuric-acid, `0.333`
   for the heavy-oil-cracking chain) — a continuous-flow LP relaxation
   that rounds up to exactly 1 physical machine, whose ingredient
   delivery isn't throttled down to the fractional duty the plan
   assumes. Per the meter's calibrated asymmetry
   (docs/meter-divergence.md), an above-plan reading is evidence of
   NOTHING, so the tripwire correctly does not alarm on this — but it's
   a real, previously-unmeasured behavior (nothing before this PR ran
   the meter against a fractional-machine-count fixture) that may be
   worth a follow-up issue, possibly related to RFC-069's
   `planning_duty` axis.

## Hard rule: only BELOW-plan ever alarms

Per the meter's calibrated asymmetry ("meter says below plan ⇒ believe
it; meter says at-or-above plan is evidence of NOTHING" —
docs/meter-divergence.md), the tripwire never treats an at-or-above-plan
reading as clearance, in code, comments, or printed text. Verified by
hand: I temporarily edited the baseline to simulate (a) a stale entity
count, (b) a flipped convergence flag, and (c) a genuine below-plan
regression, and confirmed `check` mode fails with the right message in
each case and passes cleanly against the real baseline — see the PR
diff's module doc comment for the full mode contract
(`SPAGHETTIO_METER_TRIPWIRE=bless|check`, unset = report-only default).

## Verification

- `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast`
  (unpiped): **exit 0**, 27/27 test binaries `ok`, 0 failures.
- `cargo test --manifest-path crates/meter/Cargo.toml --no-fail-fast`
  (unpiped, post-rebase): **exit 0**, all suites `ok` (`e2e_tripwire`
  correctly shows as `ignored` by default — report-only, never runs
  unattended).
- `cargo clippy --manifest-path crates/meter/Cargo.toml --tests
  --all-targets`: clean, 0 warnings.
- No changes under `crates/core/`, so no WASM rebuild needed for this
  PR.
- `SPAGHETTIO_ZONE_CACHE_PATH=.../sat-zones-ci.bin`
  `SPAGHETTIO_METER_TRIPWIRE=bless` then `=check` round-trip: clean
  pass, numbers reproducible. `git checkout --
  crates/core/data/sat-zones-ci.bin` run after every invocation that
  touched it; confirmed clean in the final diff.

- [x] `cargo test --manifest-path crates/core/Cargo.toml`
- [x] `cargo clippy` (meter crate; `-D warnings` not repo-wide policy,
      but 0 warnings either way)
- [ ] WASM rebuild — N/A, no `crates/core` changes
- [ ] Snapshot inspection — N/A, no layout-engine change

## Deviations from intent

- The task brief anticipated shelling out to
  `crates/core/examples/sim_export.rs` to build fixtures. I instead
  call `solver`/`layout`/`validate`/`blueprint` directly from the test
  (meter already depends on `spaghettio_core` as a real dependency),
  which needed no example file at all and — as a side effect — let me
  include `tier3_heavy_oil_cracking` (needs recipe exclusions;
  `sim_export.rs`'s CLI has no `--exclude` flag) that the CLI route
  would have had to drop.
- Regression tolerance (2.0pp, matching
  `crates/sim-harness/src/baseline.rs`'s own tolerance) and the
  below-plan noise floor (0.5pp) are both my judgment calls, stated and
  reasoned about in the module doc comment; no prior art in-repo to
  match exactly for a meter-specific tripwire.
- Did not attempt to root-cause either of the two findings above —
  out of scope for a "build the tripwire" PR; flagged for the campaign
  instead.

## Models / contracts touched

None. This is a new, additive diagnostic; no existing contract,
validator, or layout-engine behavior changes.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/crates/meter/tests/e2e_tripwire.rs
+++ b/crates/meter/tests/e2e_tripwire.rs
@@ -48,35 +48,44 @@
 //! just by convention: the comparison clamps both readings to a ceiling
 //! of 0 before differencing (`min(deficit_pct, 0.0)`), so the result can
 //! only be positive (alarm-worthy) when the FRESH reading is itself below
-//! plan — an above-plan baseline (sulfuric5-chem's committed +239%, for
-//! instance) can never manufacture a "regression" out of a reading that
-//! improved toward or past plan (second-opinion review on #693 round 1,
-//! finding #1 — this was a live false-alarm path, not hypothetical).
+//! plan — an above-plan baseline can never manufacture a "regression" out
+//! of a reading that improved toward or past plan (second-opinion review
+//! on #693 round 1, finding #1 — this was a live false-alarm path, not
+//! hypothetical: found against `sulfuric5-chem`'s then-committed +239%,
+//! back when fluid targets were still written to the baseline — round 2
+//! excluded them entirely, see below, so that specific row is no longer
+//! an example of anything committed, only of a MEASURED reading).
 //!
-//! A convergence collapse (baseline converged, fresh does not, AND the
-//! fresh reading is itself below plan) is judged qualitatively as its own
-//! below-plan-class failure, not folded into the percentage comparison: a
-//! previously-settling factory that no longer settles AND reads below
-//! plan is exactly the shape of the worst regression this tripwire exists
-//! to catch (a stall to ~0 — the meter's own convergence detector treats
-//! "producing nothing" specially, see `producing_nothing_is_not_converged`
-//! in `factory.rs`), and round 1's shape — treating ANY non-convergence as
-//! a graceful skip — silently passed exactly that regression
-//! (second-opinion review round 2, finding #1).
+//! A convergence collapse (baseline converged, fresh does not) is a
+//! below-plan-class failure precisely when the fresh reading is ALSO
+//! materially worse than the baseline: a previously-settling factory that
+//! no longer settles AND has genuinely gotten worse is exactly the shape
+//! of the worst regression this tripwire exists to catch (a stall to ~0
+//! — the meter's own convergence detector treats "producing nothing"
+//! specially, see `producing_nothing_is_not_converged` in `factory.rs`),
+//! and round 1's shape — treating ANY non-convergence as a graceful skip
+//! — silently passed exactly that regression (second-opinion review
+//! round 2, finding #1).
 //!
-//! **The direction gate matters, and an earlier round of this file got it
-//! wrong**: non-convergence alone is NOT evidence of a stall — the
-//! detector fails on any unstable trajectory, rising or falling
-//! (`factory.rs`'s `a_filling_buffer_is_not_converged` covers the rising
-//! case too), so round 2/3's "any non-convergence on the fresh side is a
-//! collapse" would fail on an over-producing or still-settling factory —
-//! good news, or at worst nothing — which violates the below-plan-only
-//! rule exactly as badly as round 1's "skip everything" bug violated it
-//! in the other direction (second-opinion review round 4, finding #1,
-//! the kernel bug that round fixed). The collapse branch therefore checks
-//! BOTH non-convergence AND `deficit_pct < 0.0` before failing; a
-//! non-converged, at-or-above-plan reading is `UNSETTLED` instead —
-//! reported, needs a stable re-run, never failed.
+//! **Getting the gate right took three rounds, and it's worth naming
+//! why**: round 2/3 failed on ANY non-convergence, which also fails on
+//! good news — the detector rejects any unstable trajectory, rising or
+//! falling (`factory.rs`'s `a_filling_buffer_is_not_converged` covers the
+//! rising case too), so an over-producing or still-settling factory would
+//! fail exactly like a stall (round 4, finding #1). Round 4's fix (a raw
+//! sign check, `deficit_pct < 0.0`, no tolerance) stopped failing on
+//! ABOVE-plan improvement, but not on two more shapes: a below-plan
+//! baseline IMPROVING while non-converged (blessed -25%, fresh -5% — -5%
+//! is still negative, so it still failed, despite being much better) and
+//! any sub-tolerance non-converged flap (a reading that would land in
+//! `standing`, not fail, if it happened to converge) — because a raw sign
+//! check has no tolerance at all, inconsistent with the compared branch's
+//! `REGRESSION_TOLERANCE_PP` (round 5, finding #1). The collapse branch
+//! now shares the EXACT SAME tolerance-gated clamped comparison
+//! (`clamped_drop`) as the compared branch below; only the label differs
+//! (COLLAPSE, so the reader knows convergence was also lost). A
+//! non-converged reading that ISN'T materially worse than baseline is
+//! `UNSETTLED` instead — reported, needs a stable re-run, never failed.
 //!
 //! Baseline-side non-convergence is different again and stays a skip
 //! (see "Standing gaps" below) — the baseline itself was never a
@@ -88,8 +97,11 @@
 //! not just one (round 3's text named only the first and read as if it
 //! were unique — round 4, finding #5): `ec30-am2-ore` (its 3
 //! `belt-dead-end` validator errors) and `plastic5-chem-crude` (no
-//! validator errors; the deficit is a real but smaller, unexplained
-//! shortfall). Both hit the "baseline never converged" branch before the
+//! validator errors, and NOT a below-plan deficit either — it reads
+//! +0.89%, above plan; round 5 nit — a prior revision of this comment
+//! mischaracterized it as a "shortfall". The real issue is that its
+//! reading simply never stabilizes within the warmup window). Both hit
+//! the "baseline never converged" branch before the
 //! collapse branch is ever reached, so a further regression on either is
 //! structurally unrepresentable today. This is not a gap in the check's
 //! design: an instrument cannot alarm on "worse than a baseline that was
@@ -168,7 +180,11 @@
 //!     left the build-failure assert unconditional; second-opinion review
 //!     round 2 (finding #5) caught the gap. If a new assertion is ever
 //!     added to this file, it belongs inside the `bless`/`check` arms,
-//!     never before the match.
+//!     never before the match. A missing planned rate (`no-plan` in the
+//!     scoreboard's flag column) is likewise printed, not alarmed on, in
+//!     this mode — by design, not an oversight (round 5 nit): report-only
+//!     has nothing to protect by asserting on it, unlike `bless`/`check`,
+//!     where a NaN deficit would corrupt what gets committed or compared.
 //!   * `bless` — writes the current per-fixture readings (excluding fluid
 //!     targets — see above) as the new committed baseline
 //!     (`e2e_tripwire_baseline.json`), alongside a hash of the zone-cache
@@ -185,6 +201,13 @@
 //!     documents for check severity/count conflation):
 //!       - **excluded** — a fluid target (see above); never gated.
 //!       - **new** — no baseline row (bless it).
+//!       - **target-changed** — the fixture's solved TARGET item differs
+//!         from the baseline's; a different KIND of drift than a
+//!         geometry change (a config edit under the same label, not the
+//!         engine/cache producing a different layout), so its own
+//!         accounting reason rather than folded into `stale`
+//!         (`BaselineRow.target` was written at bless time but never
+//!         read here until this — second-opinion review round 5).
 //!       - **stale** — entity count changed vs the baseline; not
 //!         comparable (the zone-cache hash recorded at bless time helps
 //!         tell "the SAME pin produced a different count" — a genuine
@@ -200,29 +223,36 @@
 //!         which `uncovered` alone would not make it (second-opinion
 //!         review round 3, finding #1).
 //!       - **collapsed** — baseline converged, fresh does not, AND the
-//!         fresh reading is itself below plan: FAILS (see the hard-rule
-//!         section above).
-//!       - **unsettled** — baseline converged, fresh does not, but the
-//!         fresh reading is AT-OR-ABOVE plan: printed (`UNSETTLED`), not
-//!         a failure and not `judged` either — a rising buffer, a
-//!         still-settling improvement, or jitter, none of which is
-//!         below-plan evidence (second-opinion review round 4, finding
-//!         #1 — collapse used to fire on non-convergence alone, which
-//!         also covers this shape and would fail on good news).
+//!         SAME tolerance-gated clamped comparison used for `compared`
+//!         below reads the fresh side as materially worse: FAILS,
+//!         prints `CONVERGENCE COLLAPSE` (see the hard-rule section
+//!         above for why this took three rounds to get right).
+//!       - **unsettled** — baseline converged, fresh does not, but that
+//!         same comparison does NOT read it as materially worse: printed
+//!         (`UNSETTLED`), not a failure and not `judged` either — a
+//!         rising buffer, a still-settling improvement (even while
+//!         remaining below plan), or plain jitter, none of which this
+//!         gate can tell apart from each other, only from a genuine
+//!         regression (second-opinion review rounds 4 and 5, finding #1
+//!         both times — collapse used to fire on non-convergence alone,
+//!         then on raw deficit sign with no tolerance; neither survived
+//!         contact with a below-plan-but-improving or sub-tolerance-flap
+//!         reading).
 //!       - **compared** — both converged, entities match: the clamped
-//!         below-plan comparison runs and lands in exactly one of three
-//!         buckets (second-opinion review round 3, finding #3 — the
-//!         previous single `compared` counter incremented BEFORE the
-//!         regression check, so a row that FAILED still printed as
-//!         "compared clean-or-standing" in the accounting, the exact
-//!         count/severity conflation `docs/validator-reporting.md`
-//!         polices): **at-or-above** (no below-plan reading at all — not
-//!         named "clean": that's a clearance word, and an at-or-above
-//!         reading is evidence of nothing either way, second-opinion
-//!         review round 4, finding #2), or **standing** (below-plan but
-//!         not worse than baseline — prints `STANDING BELOW-PLAN`, does
-//!         not fail), or **regressed** (below-plan AND materially worse
-//!         — FAILS, prints `BELOW-PLAN REGRESSION`).
+//!         below-plan comparison (`clamped_drop`) runs and lands in
+//!         exactly one of three buckets (second-opinion review round 3,
+//!         finding #3 — the previous single `compared` counter
+//!         incremented BEFORE the regression check, so a row that FAILED
+//!         still printed as "compared clean-or-standing" in the
+//!         accounting, the exact count/severity conflation
+//!         `docs/validator-reporting.md` polices): **at-or-above** (no
+//!         below-plan reading at all — not named "clean": that's a
+//!         clearance word, and an at-or-above reading is evidence of
+//!         nothing either way, second-opinion review round 4, finding
+//!         #2), or **standing** (below-plan but not worse than baseline
+//!         — prints `STANDING BELOW-PLAN`, does not fail), or
+//!         **regressed** (below-plan AND materially worse — FAILS,
+//!         prints `BELOW-PLAN REGRESSION`).
 //!
 //!     `check` FAILS if nothing landed in `collapsed`, `regressed`,
 //!     `standing`, or `at-or-above` — a check run that judged nothing
@@ -283,10 +313,26 @@
 //!   at -0.2%, the closest-to-plan below-plan reading in the corpus —
 //!   could in principle flap non-converged on a noisy host (CPU
 //!   contention widening the tick-timing jitter the convergence detector
-//!   samples) and read as a `CONVERGENCE COLLAPSE` failure with no real
-//!   engine change behind it (second-opinion review round 3, nit).
-//!   Known, not fixed: `check` is deliberately not CI-wired (see above),
-//!   so this is a "confusing local re-run" risk, not a false CI red.
+//!   samples) (second-opinion review round 3, nit). **Resolved by round
+//!   5's fix, not still open**: the collapse branch now shares
+//!   `compared`'s tolerance-gated `clamped_drop`, so a flap that stays
+//!   within `REGRESSION_TOLERANCE_PP` of the blessed value lands in
+//!   `UNSETTLED`, the same way it would land in `standing` if it
+//!   happened to converge — it can no longer read as a false
+//!   `CONVERGENCE COLLAPSE`. Left here, marked resolved, so the history
+//!   stays traceable rather than the bullet just vanishing.
+//! * Convergence (`factory.rs`'s checkpoints) samples TOTAL delivered
+//!   counts across every item in the factory, while a solid target's own
+//!   `deficit_pct` here watches PRODUCED for that target specifically
+//!   (see `build_and_measure`) — two different signals (second-opinion
+//!   review round 5, "absorb #3"). A genuine sink-side throttle
+//!   (delivery destabilizing while the target's own production stays
+//!   rock-steady at plan) would present as non-converged + at-or-above
+//!   plan, landing in `UNSETTLED` — which this gate never fails, by
+//!   design. Named, not fixed: closing it means watching the target's
+//!   own delivered/produced series for convergence rather than the
+//!   whole-factory delivered total, a bigger change than this
+//!   diagnostic's scope.
 
 use rustc_hash::FxHashSet;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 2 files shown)


storkme:
Absorbed all four findings from the second-opinion review on c22ac53b, in one batched commit (4e3ccc0e) per the no-double-push rule. Verified each at source before fixing — all four were real, none refuted.

**1. [major] confirmed live, fixed.** `drop = b.deficit_pct - m.deficit_pct` with no clamping: the committed baseline genuinely has `sulfuric5-chem`/`lightoil5-chem-cracking` at +239%/+200%, so any fresh reading on those moving toward or past plan would read as a 200+pp "regression." Fixed exactly as suggested: both readings are clamped to a ceiling of 0 before differencing (`min(deficit_pct, 0.0)`), which structurally guarantees the comparison can only be positive when the FRESH reading is itself below plan.

  Executed the discrimination check rather than reasoning about it: mutated `ac5-am2-ore`'s baseline row to +200% (its real fresh reading is -0.2%, converged both sides) and reran `check` — **no alarm** (`test e2e_tripwire ... ok`). Then restored the real baseline and confirmed a genuine regression (baseline faked to -2.0% against gear20-am2-plate's real -25.0%) still fails with `BELOW-PLAN REGRESSION — baseline -2.0%, now -25.0% (23.0pp worse, below-plan portion only)`.

**2. [moderate] confirmed, took the recommendation.** Demoted entity-count drift and convergence-flip to printed skips rather than test failures — and, going one step further than the finding named, did the same for "no baseline row" (a new fixture), since that's equally not a below-plan finding and would have broken the same contract. `check` mode's only failure condition is now the below-plan comparison itself; everything else prints and continues.

**3. [minor] confirmed live.** The old guard was `b.converged != m.converged` — a pure flip check — so a both-sides-unconverged reading fell through uncaught. The committed baseline has three fixtures in exactly that shape (`ec30-am2-ore`, `plastic5-chem-crude`, `sulfuric5-chem`, all `converged: false` on both sides), and running `check` against the unmodified baseline before this fix would have silently compared their deficits despite the code's own comment calling such readings untrustworthy. Now gates on `!b.converged || !m.converged` (either side), matching `corpus_replay.rs`'s own precedent of treating an unconverged reading as unusable rather than merely suspect.

**4. [minor] confirmed.** `assert_has_plan` (the renamed sanity check) is now called only from the `bless`/`check` match arms, not unconditionally before the match. Report-only mode's only possible failure is the build-failure check.

**Verification after the fix** (unpiped exit codes):
- `cargo test --manifest-path crates/meter/Cargo.toml --no-fail-fast` → exit 0, all suites ok (`e2e_tripwire` still `ignored` by default).
- `cargo clippy --manifest-path crates/meter/Cargo.toml --tests --all-targets` → exit 0, 0 warnings.
- `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` → rerunning now (no `crates/core` changes in this commit, but rerunning the full verification set as instructed); will follow up if it turns up anything, though nothing in this diff touches core.
- Bless→check round-trip against the real baseline: clean pass, numbers unchanged (only comparison logic and doc comments moved, not `build_and_measure`).
- Zone-cache pin (`crates/core/data/sat-zones-ci.bin`) restored via `git checkout --` after every run that touched it; clean in the final diff.

Not merging — that's the campaign lead's call.

storkme:
Follow-up on the verification I flagged as still running: `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` (unpiped) → exit 0, 27/27 test binaries ok, 0 failures. As expected — this commit touches only `crates/meter/tests/e2e_tripwire.rs`.

storkme:
Round 2 absorbed on 4e3ccc0e, in one batched commit (836982b8, single push). Verified every finding at source before deciding. Prioritization followed as given.

## MUST FIX — both confirmed live

**1. [major] Convergence collapse.** Confirmed: round 1's `if !b.converged || !m.converged { skip }` treated a flip identically to both-sides-unconverged, so a fixture that regressed from "converges fine" to "collapsed to ~0, no longer converges" would silently skip rather than fail — exactly the worst regression class. Fixed: baseline-converged + fresh-unconverged is now its own `CONVERGENCE COLLAPSE` failure; baseline-unconverged stays a skip (never a trustworthy anchor to regress *from*) but is now counted and printed as `UNCOVERED (baseline non-converged)` (closes finding #3 too).

**2. [major] Non-vacuous check.** Confirmed: with every guard demoted to a skip in round 1, an all-mismatch run (wrong/missing zone-cache pin, or any other reason every row becomes unjudgeable) printed 0 failures and read as clean. Fixed: `check` now fails unless `judged() > 0` (compared or collapsed), and always prints an accounting line. Executed both ways:
- Unit test `all_rows_stale_yields_zero_judged` (synthetic).
- Full-stack: mutated the real committed baseline's entity counts to 1 across the board, ran `check` against the real pipeline — `accounting: 0/13 judged ... 13/13 not judged`, test **FAILED** with the expected "compared NOTHING" message, then restored the baseline and confirmed a clean re-pass.

**3. [minor] Uncovered rows visible.** Now printed explicitly (`UNCOVERED (baseline non-converged)`) and counted in the accounting line — currently `ec30-am2-ore` and `plastic5-chem-crude`.

## ADJUDICATED

**4. [major] Fluid targets — CONFIRMED, with stronger receipts than the finding stated.** Checked `crates/meter/src/fluid.rs` and `docs/meter-divergence.md` directly. Fluid delivery is NOT unmodeled — `fluid.rs`'s Phase B pipe network populates real, non-zero `delivered_per_s` numbers for both fixtures (I'd verified this in round 1's own investigation). But `meter-divergence.md` states plainly: *"Fluid targets are untested. All 4 fluid-target fixtures... have no sim baseline in this corpus. The floor verdict is a solid-target-only result,"* and separately, under "Deliberately-not-modelled": *"the meter drains every unconsumed producer fluid unit as delivered and never stalls a producer on a full fluid output"* — exactly the mechanism behind `sulfuric5-chem`/`lightoil5-chem-cracking`'s measured +239%/+200% (a fractional solved machine count rounds up to one physical machine that isn't duty-cycled down, and the meter drains its full unthrottled output as "delivered"). So the calibrated asymmetry this whole tripwire rests on has never been established for these two targets in *either* direction. Both fixtures are now excluded from bless/check (re-blessed baseline: 13 → 11 rows) and marked `fluid (uncalibrated)` in the scoreboard regardless of sign — still measured and reported, never gated.

**5. [major] Report-only hard-fails — CONFIRMED**, same bug class as round 1's finding #4 (`no_plan`) but for `build_failures`, which I'd left unconditional. Fixed: report-only now prints build failures via `eprintln!` and asserts nothing; the assert (`assert_no_build_failures`) moved into `bless`/`check` only, alongside `assert_has_plan` (already correctly scoped since round 1 — confirmed no doc-vs-code gap remained there once this fix landed).

**6. [minor ×2] Docs overselling the ratchet — CONFIRMED.** Rewrote the module doc with a dedicated "The committed baseline is this instrument's ZERO, not a live floor" section: `check` alarms on regression relative to the blessed state only; a standing deficit (`gear20-am2-plate` at -25%, blessed that way) stays clean forever until someone deliberately re-blesses tighter; explained why (bootstrapping a regression gate on a corpus with pre-existing, known, un-actioned deficits — same design as `e2e.rs`'s `StressBaseline` ceilings). `check` now also prints a `STANDING BELOW-PLAN` line per such row so "clean" is never mistaken for "nothing here is wrong."

## REMAINING MINORS (7-9)

Absorbed the cheap part: the baseline now records a content hash of the zone-cache file, and `check` prints a note when the fresh hash differs from the blessed one — directly strengthens the stale-entity message (distinguishes "same pin, genuine engine change" from "different pin, geometry noise") and finding #2's diagnosability. Recorded as known limitations in the module doc rather than built further: structural-signature staleness detection (entity count alone can't catch a same-count-different-shape change) and engine-commit provenance (lower value here than in the sim harness, since this baseline is always compared against whatever commit is checked out, not an externally-versioned binary) — not enough marginal value in a first-cut diagnostic to justify more machinery.

## Executed, not reasoned

Extracted the comparison into a pure `evaluate_check(results, baseline, verbose) -> CheckOutcome` function and added 8 unit tests with synthetic `Measurement`/`BaselineRow` data — a live convergence flip can't be manufactured by mutating the baseline JSON alone (unlike round 1's false-alarm scenario, which only needed the baseline side to change). Proved `convergence_collapse_fails` actually discriminates: reverted the fix to the round-1 shape (`if !b.converged || !m.converged { skip }`), reran, and confirmed **only** that test failed (7/8 others still passed), then restored. Also re-ran round 1's `above_plan_baseline_moving_to_near_plan_does_not_alarm` scenario post-refactor to confirm the fix survived.

## Verification (unpiped exit codes)

- `cargo test --manifest-path crates/meter/Cargo.toml --no-fail-fast` → exit 0, all suites ok (includes the 8 new unit tests, running unignored; `e2e_tripwire` itself still `ignored`).
- `cargo clippy --manifest-path crates/meter/Cargo.toml --tests --all-targets` → exit 0, 0 warnings.
- `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` → exit 0, 27/27 binaries ok (unaffected, no core changes).
- Re-blessed baseline (schema change: wrapper object + zone-cache hash + fluid rows removed, 13 → 11 rows) with the CI zone-cache pin; clean `bless` → `check` round-trip against the real numbers.
- Zone-cache pin restored via `git checkout --` after every run that touched it; clean in the final diff.

Not merging — that's the campaign lead's call.

storkme:
Round 3 absorbed on 836982b8, in one batched commit (b44ab1be, single push). Verified each at source.

**1. [major] ec30-am2-ore collapse coverage — confirmed, and partially refuted as instructed.** Confirmed: its baseline row is `converged: false`, so `evaluate_check` hits the "baseline never converged" branch before the collapse branch is ever reached — collapse coverage is structurally unreachable on that fixture while it stays blessed non-converged. Refuted the implication that this is fixable in the comparison logic: an instrument cannot alarm on "worse than a baseline that was already producing nothing" — there is no direction left to regress FROM, and that's inherent to any blessed-baseline design (the same reason `e2e.rs`'s `StressBaseline` ceilings can't ratchet below zero errors). `ec30-am2-ore`'s real fix is the layout itself (3 `belt-dead-end` validator errors, found in round 1's investigation) — this stays queued campaign-side as the W1c contested sample, not something this diagnostic PR touches. Absorbed the real, cheap part: a `RECOVERED` branch (baseline non-converged, fresh NOW converges) is printed and counted separately from `uncovered`, with the module doc stating the flip condition explicitly (fix the layout, re-bless converged, collapse coverage arms itself automatically).

**2. [minor, real ordering bug] — confirmed.** `hash_zone_cache()` was called after the build loop (inside the `bless`/`check` match arms), and `build_and_measure` can append newly-solved zones to the cache file mid-run — so the hash described post-run bytes, not the committed pin. Now captured once at the very top of `e2e_tripwire`, before any fixture builds, and both `bless` and `check` reuse that one value. Re-blessed; the stored hash is unchanged in practice (`929b8d28938ed330` both times, since this repo's git-checkout-after-every-run discipline already meant the pin file was pristine at the start of every run I'd made) — but the ordering is now correct by construction, not by the accident of that discipline.

**3. [minor] — confirmed.** `out.compared += 1` ran unconditionally before the regression check, so a row that FAILED still printed as "compared clean-or-standing" — count/severity conflation, the exact class this instrument exists to police elsewhere. Split into `compared_clean`, `standing` (`standing_below_plan.len()`), and `regressed`, each incremented only once the row's actual disposition is known. `judged() = compared_clean + standing + regressed + collapsed`.

Also added the requested doc note on the ac5-am2-ore flap risk (a fixture blessed right at the convergence boundary could read as a collapse on a noisy host) as a known limitation, tied to "check is deliberately not CI-wired."

## Executed, not reasoned

- Added `baseline_never_converged_recovers_when_fresh_converges` and proved it discriminates: reverted the RECOVERED branch to round 2's shape, reran, watched exactly that test fail, restored.
- Strengthened `above_plan_baseline_moving_to_near_plan_does_not_alarm` and `genuine_below_plan_regression_still_fails` to assert on the new `regressed`/`compared_clean`/`standing` counters, and proved those discriminate too: reverted the split back to an unconditional `compared_clean += 1`, reran, watched exactly those two tests fail (for the right reason — count conflation), restored.
- Ran the RECOVERED scenario end-to-end against the real pipeline, not just synthetically: faked `gear10-am1-plate`'s baseline row to `converged: false` (its real fresh reading genuinely converges), ran `check` — printed `gear10-am1-plate: RECOVERED — ... re-bless to arm collapse coverage`, accounting read `1 recovered`, test still `ok`. Restored the real baseline after.

## Verification (unpiped exit codes)

- `cargo test --manifest-path crates/meter/Cargo.toml --no-fail-fast` → exit 0, all suites ok (9 unit tests now, up from 8; `e2e_tripwire` itself still `ignored`).
- `cargo clippy --manifest-path crates/meter/Cargo.toml --tests --all-targets` → exit 0, 0 warnings.
- `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` → exit 0, 27/27 binaries ok (unaffected, no core changes).
- Re-blessed baseline (hash-timing fix): committed JSON is byte-identical to before, only the `.rs` file changed in this diff.
- Zone-cache pin restored via `git checkout --` after every run; clean in the final diff.

Not merging — that's the campaign lead's call.

storkme:
Round 4 (final pass) absorbed on b44ab1be, in one batched commit (81fc6e18, single push). Verified each at source.

## Kernel fix

**1. [major] Collapse fired on non-convergence alone — confirmed, and it was a real hole introduced by our own round-2/3 fix.** `factory.rs`'s convergence detector fails on any UNSTABLE trajectory, not only a falling one — `a_filling_buffer_is_not_converged` covers a rising reading too. So the old collapse branch (`if !m.converged { collapse }`) would fail on an over-producing or still-settling factory just as readily as on a genuine stall, which violates the below-plan-only hard rule exactly as badly as round 1's original "skip everything" bug violated it in the other direction — just inverted. Fixed: collapse now requires BOTH non-convergence AND `m.deficit_pct < 0.0`. A non-converged, at-or-above-plan reading lands in a new `UNSETTLED` bucket — reported, not judged, never failed. (A genuine stall still reads unambiguously negative regardless, since `measured ≈ 0` against any `planned > 0` is `deficit_pct ≈ -100%`, so this loses no real coverage on the case that matters.)

Executed, not reasoned: added `convergence_collapse_does_not_fire_on_improvement_shape` (baseline converged, fresh non-converged but +5% above plan) and proved it discriminates — reverted the direction gate to the round-2/3 unconditional shape, reran, watched exactly that new test fail (`"a non-converged AT-OR-ABOVE-plan reading must never fail"`), restored.

## One-liners, all absorbed

2. Renamed `compared_clean`/"clean" → `compared_at_or_above`/"at-or-above" throughout (struct field, accounting print, module doc, tests) — "clean" is a clearance word for a reading the calibration says is evidence of nothing.
3. `REGRESSION_TOLERANCE_PP`'s doc no longer claims it "matches" the sim-harness's tolerance — confirmed they're different quantities (theirs: relative-to-blessed-rate; ours: fixed percentage-points-of-deficit) and rewrote the comment to justify 2.0pp on tick-window-jitter grounds alone.
4. Named the clamped comparison's blind region directly in the tolerance's doc comment: a fixture blessed comfortably above plan that drops to just-below (within tolerance of 0) won't alarm — accepted trade-off, stated why.
5. Fixed the doc/data mismatch: `plastic5-chem-crude` is also blessed non-converged, not just `ec30-am2-ore` — round 3's text named only one and read as unique. Both are now named together wherever this limitation is discussed.
6. `assert_has_plan` now skips fluid rows, matching their bless/check exclusion.

## Refuted as designed (no change)

"Entity-count guard swallows most engine changes" — this is the geometry-keyed-baseline design shared with `crates/sim-harness/src/baseline.rs`'s stress goldens, and it hasn't been silent since round 2: the accounting line prints `N judged / M not-judged` with a stale-entity breakdown, and `check` fails outright if nothing was judged. A structural-hash upgrade (two layouts with the same entity count would pass undetected) is already recorded as a known limitation in the module doc — a future hardening, not a defect in this PR.

## Verification (unpiped exit codes)

- `cargo test --manifest-path crates/meter/Cargo.toml --no-fail-fast` → exit 0, all suites ok (11 unit tests now, up from 10; `e2e_tripwire` itself still `ignored`).
- `cargo clippy --manifest-path crates/meter/Cargo.toml --tests --all-targets` → exit 0, 0 warnings.
- `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` → exit 0, 27/27 binaries ok (unaffected, no core changes).
- No baseline schema change this round (`BaselineRow`/`Baseline` untouched — only in-memory `CheckOutcome` and the collapse gate changed), so no re-bless needed. Confirmed live: ran `check` against the unmodified committed baseline with the CI zone-cache pin — clean pass, new accounting line reads `9/13 judged (5 at-or-above, 4 standing, 0 regressed, 0 collapsed), 4/13 not judged (..., 0 unsettled, ...)`, matching expectations (no fixture in today's corpus currently exercises the collapsed/unsettled paths).
- Zone-cache pin restored via `git checkout --` after every run; clean in the final diff.

Not merging — per the instructions this is the final pass and the campaign lead merges directly.

storkme:
Round 5 (the actual last pass) absorbed on 81fc6e18, in one batched commit (b2fa6d76, single push). Verified each at source.

## The kernel

Confirmed the diagnosis exactly: round 4's raw sign check (`m.deficit_pct < 0.0`) fixed "collapse fires on good news" but was still inconsistent with the compared branch's tolerance, in two concrete ways I could reproduce:
- **Below-plan-but-improving false fail**: baseline blessed at -25%, fresh reads -5% non-converged — still negative, so round 4 still failed it as a collapse, even though -5% is a large improvement over -25%.
- **Sub-tolerance flap false fail**: baseline at-plan (0%), fresh reads -0.5% non-converged — round 4 failed it, even though the identical -0.5% would land in `standing` (not fail) if it happened to converge.

Fixed exactly as specified: factored the clamped comparison out into `clamped_drop(baseline_deficit_pct, fresh_deficit_pct)` and now call the SAME function from both the collapse and compared branches. Collapse fails only when `!m.converged` AND `drop > REGRESSION_TOLERANCE_PP`; everything else non-converged is `UNSETTLED`. Only the printed label differs (`CONVERGENCE COLLAPSE` vs `BELOW-PLAN REGRESSION`).

Executed, not reasoned: renamed the stall test to `convergence_collapse_fails_on_a_true_stall` (baseline -2%, fresh -100% — drop ≈ 98pp, still caught), added `convergence_collapse_does_not_fire_on_below_plan_improvement` and `convergence_collapse_does_not_fire_on_sub_tolerance_flap`, kept the round-4 above-plan-improvement test (still passes unchanged under the unified formula — confirms the fix is additive, not a regression). Then reverted the collapse branch to round 4's raw-sign shape and reran: exactly the two new tests failed (`"an improving-while-still-below-plan reading must never fail"` and `"a sub-tolerance non-converged flap must never fail"`), the other four collapse-related tests stayed green. Restored.

## Absorbed

1. **`BaselineRow.target` dead field — confirmed.** It was written at bless time and never read in `check`. Added `m.target != b.target` as its own skip reason (`skipped_target_changed`, distinct from the entity-count `skipped_stale`), with its own accounting count and a unit test (`target_changed_is_skipped_not_failed`).
2. **STANDING message factually wrong — confirmed and reworded.** "Accepted at bless time as this fixture's zero point" is false when the baseline itself was at-or-above plan and only the fresh reading dipped sub-tolerance below it (e.g. `ac5-am2-ore`'s own -0.2% case, if it flapped). Now states the actual condition: `"within {tolerance}pp tolerance of the blessed value (b=X%)"`.
3. **Delivered-vs-produced seam — confirmed and documented.** Traced it in `factory.rs`: convergence checkpoints sample `self.delivered.values().sum()` (every item, whole-factory), while a solid target's `deficit_pct` here watches `produced_per_s` for that target specifically. A sink-side throttle that destabilizes delivery while production stays rock-steady would present as non-converged + at-or-above-plan, landing in `UNSETTLED` — invisible to this gate by design. Named as a known limitation, not fixed (bigger change than this diagnostic's scope).
4. **Nits, all confirmed and fixed:** two stale "sulfuric5-chem's committed +239%" claims (fluid targets have been excluded from the baseline since round 2 — it was never "committed," only measured); `plastic5-chem-crude` mischaracterized as having a below-plan "shortfall" when its actual reading is +0.89%, above plan (the real issue is it never converges, not a deficit); rewrote the hard-rule section's "direction gate" framing, which described round 4's since-superseded mechanism, to describe the final unified one and the three-round history plainly; noted report-only's no-plan handling (printed, never alarmed on) is deliberate.

Bonus: round 5's fix also resolves round 3's `ac5-am2-ore` flap nit that I'd previously recorded as an open "known limitation, check is deliberately not CI-wired" — with the unified tolerance, a flap that stays within 2.0pp of the blessed value now lands in `UNSETTLED`, not `COLLAPSE`. Left the bullet in place, marked resolved, rather than deleting it, so the history stays traceable.

## Refuted as recorded (no change)

The entity-count-staleness concern is round 4's finding, re-raised on its second pass. Same adjudication as before: this is the geometry-keyed-baseline design shared with `crates/sim-harness/src/baseline.rs`'s stress goldens, not silent since round 2 (the accounting line prints `N judged / M not-judged` with a stale-entity breakdown, and `check` fails outright on zero judged). The structural-hash upgrade (two layouts with the same entity count would pass undetected) is already recorded in the module doc's "Known limitations" section as a future hardening, not a defect in this PR.

## Verification (unpiped exit codes)

- `cargo test --manifest-path crates/meter/Cargo.toml --no-fail-fast` → exit 0, all suites ok (14 unit tests now, up from 11 — 3 net new after replacing 1 test with 3; `e2e_tripwire` itself still `ignored`).
- `cargo clippy --manifest-path crates/meter/Cargo.toml --tests --all-targets` → exit 0, 0 warnings.
- `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast` → exit 0, 27/27 binaries ok (unaffected, no core changes).
- No baseline schema change this round (`BaselineRow`/`Baseline` untouched again), so no re-bless needed. Confirmed live: ran `check` against the unmodified committed baseline with the CI zone-cache pin — clean pass, accounting line now shows the `target-changed` bucket and the reworded standing messages (`"ac5-am2-ore: STANDING BELOW-PLAN — -0.2%, within 2.0pp tolerance of the blessed value (-0.2%); not a new regression."`).
- Zone-cache pin restored via `git checkout --` after every run; clean in the final diff.

Not merging, per the instructions — ready for the required check and direct merge.

storkme:
**Round 6 adjudicated post-merge (closing-round protocol, declared at round 5).** Both findings are message-wording minors on the diagnostic's output: the STANDING/UNSETTLED messages print the raw blessed value where the comparison used the clamped one — misleading for improved-below-plan and above-plan-baseline shapes. Real, small, and recorded on #689 as a tripwire carry-over (print the clamped baseline or rephrase as 'no more than tol below the plan-relative baseline'). The comparison math itself is correct — the discrimination tests added in rounds 2-5 pin it.

PR #694, opened 2026-08-21
diag(w1c): parity-corpus baseline — RFC-070 Phase 0c

storkme:
## Intent

RFC-070 Phase 0c (#689 track W1c). K70-1 and K70-2 are **not re-runnable
until the parity corpus is named** — the RFC says so explicitly, and
naming it is a Phase-0c deliverable rather than an option. This names it
as code, commits the baseline as data, writes the divergence-equivalence
rule the shadow loop will be judged against, and sim-anchors the
contested sample.

## Scope

- **In:** `crates/core/tests/parity_corpus.rs` (the grid diagnostic +
  bless/check), `crates/core/tests/parity_corpus_baseline.json` (the
  committed baseline), the divergence-equivalence rule and four decision-log
  entries in `docs/rfc-070-one-selection-loop.md`, the recorded #691/#692
  carry-overs, and two sim anchors.
- **Out:** the two e2e-harness fossils stay UNFIXED (flipping either
  changes the candidate set under every regression test — a Wave-2
  campaign call, exactly as W1b recorded). The ec30 layout defect is
  reported, not fixed. No engine file is touched by this PR.
- **Size:** 1,064 added lines outside the committed baseline JSON
  (1,445 lines of pure data, excluded by the norm). Breakdown: the
  diagnostic 661 (of which **297 are comment/doc lines** and ~110 are the
  fixture table), the RFC 128, the two carry-over files 270, status.md 5.
  Why it is one PR: the corpus and its equivalence rule are a single claim
  — a committed baseline without the rule is a table nobody can adjudicate
  against, and the rule without the baseline names nothing. The sim anchors
  are what make the baseline worth reviewing rather than worth trusting.
  **The weakest part of that argument is the carry-overs** (270 lines,
  `check_firing_census.rs` + `junction_seed_census.rs`): they were bundled
  because #689 queued them onto this track, and they could defensibly have
  been their own PR. They are at least adjacent — the contract test now pins
  the three deciding stages the corpus measures live, which is a guard the
  corpus itself showed was needed.

## The corpus

**160 cells** = 12 fixtures (the #691 corpus verbatim) × permitted
machine tiers × 5 option sets. 140 decide; 20 are `no-solve`
(`assembling-machine-1`'s two ingredient slots cannot run
advanced-circuit or processing-unit — recorded, not dropped).

Option sets: `default`, `cells-off`, `e2e-harness`, `di-off`, `hs-off`.

```
status distribution:         {"decided": 140, "no-solve": 20}
deciding-stage distribution: {"best-error-free": 87, "scoped-pairwise": 26,
                              "merge-tap": 15, "best-accepted": 12}
15/32 fixture×machine rows change verdict across the option-set axis.
(10 major = winner changes, 12 minor = stage-only)
```

Four of the five stages are live. `first-produced` fires **nowhere** on
this corpus — recorded as a hole, so Phase 2a does not read corpus parity
as evidence about that arm.

### Where the winner changes (the claim surface)

<details><summary>10 major + 12 minor, full list</summary>

```
MAJOR e2e_tier2_electronic_circuit_20s_from_ore [am3] e2e-harness = native/best-error-free        default = direct-insertion/scoped-pairwise
MAJOR e2e_tier2_electronic_circuit_20s_from_ore [am3] di-off      = native/best-error-free        default = direct-insertion/scoped-pairwise
MINOR e2e_tier2_electronic_circuit_from_ore     [am1] di-off      = native/best-accepted          default = native/best-error-free
MAJOR e2e_tier2_electronic_circuit_from_ore     [am2] hs-off      = native/best-error-free        default = horizontal-stack/scoped-pairwise
MINOR e2e_tier3_plastic_bar_from_crude          [cp]  cells-off   = native/best-accepted          default = native/best-error-free
MINOR e2e_tier3_plastic_bar_from_crude          [cp]  e2e-harness = native/best-accepted          default = native/best-error-free
MAJOR e2e_tier4_advanced_circuit_from_ore_am2   [am3] hs-off      = native/best-error-free        default = horizontal-stack/scoped-pairwise
MAJOR e2e_tier5_processing_unit_from_ore_am3    [am2] hs-off      = native/best-error-free        default = horizontal-stack/scoped-pairwise
MINOR tier1_gear_am1                            [am1] cells-off   = native/best-accepted          default = native/best-error-free
MINOR tier1_gear_am1                            [am1] e2e-harness = native/best-accepted          default = native/best-error-free
MINOR tier1_gear_am1                            [am2] cells-off   = native/best-accepted          default = native/best-error-free
MINOR tier1_gear_am1                            [am2] e2e-harness = native/best-accepted          default = native/best-error-free
MINOR tier1_gear_am1                            [am3] cells-off   = native/best-accepted          default = native/best-error-free
MINOR tier1_gear_am1                            [am3] e2e-harness = native/best-accepted          default = native/best-error-free
MAJOR tier2_ec_am1_10_ore                       [am1] di-off      = horizontal-stack/best-error-free  default = direct-insertion/best-error-free
MAJOR tier2_ec_am1_10_ore                       [am3] e2e-harness = native/best-error-free        default = direct-insertion/scoped-pairwise
MAJOR tier2_ec_am1_10_ore                       [am3] di-off      = native/best-error-free        default = direct-insertion/scoped-pairwise
MINOR tier2_ec_am2_30_ore                       [am1] di-off      = native/best-accepted          default = native/best-error-free
MINOR tier3_plastic_cp_5                        [cp]  cells-off   = native/best-accepted          default = native/best-error-free
MINOR tier3_plastic_cp_5                        [cp]  e2e-harness = native/best-accepted          default = native/best-error-free
MAJOR tier4_ac_am2_5_unconstrained              [am2] hs-off      = native/best-error-free        default = horizontal-stack/scoped-pairwise
MAJOR tier4_ac_am2_5_unconstrained              [am3] hs-off      = native/best-error-free        default = horizontal-stack/scoped-pairwise
```

</details>

Every MINOR has one mechanism, and it is W1b's: with cells off only
native produces, `clean_flags` is skipped by its own `n_layouts > 1`
guard, and the decision falls from `best-error-free` to `best-accepted`.

## Finding: the e2e harness carries a SECOND fossil, and it steers selection

W1b found `run_e2e` spells `cell_composition: Default::default()` — the
ENUM default `Off`, not the struct default `Candidate`. The same struct
literal also pins **`inserter_capacity: 0`** (`tests/e2e.rs:354`). Same
cause: correct when written (`40fd48dc`, RFC-049 Phase 1, 2026-07-22, when
the struct default WAS 0), stale two days later when #383 flipped the
default to `DEFAULT_INSERTER_CAPACITY` = 2.

It is not cosmetic. `e2e-harness` (both fields) diverges from **both**
`default` and `cells-off` on two cells:

```
-- tier2_ec_am1_10_ore  [assembling-machine-3]
   default         direct-insertion/scoped-pairwise
   cells-off       direct-insertion/scoped-pairwise
   e2e-harness  != native/best-error-free            <-- capacity 0 alone flips the winner
   di-off       != native/best-error-free
   hs-off          direct-insertion/scoped-pairwise
```

(`e2e_tier2_electronic_circuit_20s_from_ore [am3]` is identical in shape.)
So the inserter ladder steers selection independently of the cells arm,
and a corpus that modelled the harness as "cells-off" — the W1c brief's
stated minimum — would have recorded the wrong winner for the
configuration the entire regression suite runs under. Both fossils stay
unfixed here; what changes is that the corpus now measures both.

## Sim anchors

Both runs `--warmup 432000 --speed 32 --timeout-secs 3600` — 2 game-hours, per
CLAUDE.md's deep-chain rule (the dim-scaled default reads buffer fill as
throughput). Both exports are **entity-count-exact against the meter
tripwire's blessed rows** (1991 / 2134), so the sim and the meter measured
the same artifacts rather than two configurations with the same name.

### `ac5-am2` — scoped-pairwise, winner `horizontal-stack`: **PASS**

```
advanced-circuit    planned 5.00   produced 4.92 (-1.6%)   delivered 5.03 (+0.6%)   PASS
copper-cable          50.00  +0.2% | copper-plate  25.00  +0.0%
electronic-circuit    10.00  -0.2% | iron-plate    10.00  +0.0%
plastic-bar           10.00  +0.0% | petroleum-gas 90.91  -0.6%
census: 136 working, 0 full_output. kit-clean. 12 checkpoints.
```

`converged: false`, but read the series before reading the flag: it
**oscillates across plan** (4.92 → 5.10 → 4.94 → 5.12 → 4.94 → 4.92 → …,
period 2) rather than decaying. The harness's "still decaying −3.9% across
the last 3 windows" test is reading a sawtooth as a trend. So the corpus's
one pairwise displacement of native is measured good.

One caveat travels with this number. The run needs
`--research-productivity plastic-bar=0.10` — the force realizes +10% on
plastic and the harness kit-FAILS a manifest declaring 0 (RFC-064 item 7;
the first attempt came back `OVERALL: NO DATA` for exactly this). The
declared-axis export is **2125 entities against the corpus cell's 2134**,
same 137×78 dims — so this anchors a near-neighbour of the cell, not the
cell. Two consequences, both recorded: `status.md`'s "AC is bit-identical
declared-or-not" is retired, and **the corpus's `default` cells are not all
directly sim-anchorable** — any chain containing a force-boosted recipe
must be anchored on a declared-axis re-export, and Phase 2a must say which
artifact a divergence's sim anchored.

### `ec30-am2` — merge-tap, winner `native`: **FAIL, and it is real**

```
electronic-circuit  planned 30.00  produced 0.00 (-100.0%)  delivered 0.00 (-100.0%)  FAIL
copper-cable 90.00 -100.0% | copper-plate 45.00 -100.0% | iron-plate 30.00 -100.0%
census: 150 full_output, 20 item_ingredient_shortage.
kit_errors {} · fluid_errors {} · 1991/1991 ghosts revived · 1 pole network
validator on the shipped layout: 3E/76W — belt-dead-end×3
```

**Zero at every stage, including the first**, across all 4 checkpoints at a
2-game-hour warmup. `full_output: 150` against `item_ingredient_shortage: 20`
says this is a **jam, not a starve** — consistent with the 3 `belt-dead-end`
errors the layout ships with. Kit-clean and fluid-clean, so the sim-kit
first-suspect is ruled out. The meter tripwire's blessed −100%/non-converged
row (#693) was right, and this is ground truth confirming it.

**This is a real shipped-layout defect. Not fixed here** — W1c is a
diagnostic track and this PR touches no engine file. Logged loudly in
`docs/status.md` under open layout-quality issues.

Three things it clears up:

1. **The premise the track was dispatched with is wrong twice over.** #689's
   session state says "the #644-era sim anchor had ec30 ≈99.4%". That anchor
   is `stress_electronic_circuit_30s_from_ore`, which forces
   `Some("transport-belt")` AND runs through `run_e2e` — i.e. under both
   harness fossils. Different fixture: 3369 entities (banked dense winner)
   or 4934 (sprawl), against this one's 1991. And 99.4% is RFC-069's
   `planning_duty: 0.6` artifact; the shipped-default anchor for even that
   fixture is **92.1%**. So instrument-vs-fixture-vs-regression resolves as:
   **the instrument was right, the two numbers were never about the same
   layout, and the layout is genuinely dead.**
2. **Why it ships green.** No e2e test covers `electronic-circuit@30` on
   `assembling-machine-2` with an engine-chosen belt tier — every ec@30/am2
   case in `e2e.rs` passes `Some("transport-belt")`. Suite greenness has
   never said anything about this configuration. This is precisely the
   coverage shape the corpus's option-set axis exists to expose.
3. **Selection shipped it knowingly.** The committed baseline records this
   fixture as `native` / `merge-tap` under all five option sets — the
   `ErrorKinds` path whose own comment says it "still returns the
   error-laden best rather than refusing". A 3-Error layout won, and it
   produces nothing. That is the miscalibration class RFC-070 exists to
   fix, measured on a live fixture.

## Verification

- [x] `cargo clippy --workspace -- -D warnings` — **exit 0** (CI's exact
      invocation). `--all-targets` additionally fails on pre-existing
      lints in `power_wires.rs`, `e2e.rs`, `junction_fixtures.rs` and two
      gitignored examples, none of them touched here; the one it flagged
      in my own file (`type_complexity` on `OPTION_SETS`) is fixed.
- [x] `cargo test --manifest-path crates/core/Cargo.toml --no-fail-fast`
      — **exit 0**, all binaries `ok`, **1185 passed / 0
      failed** / 76 ignored from one clean invocation, zone-cache pinned.
      Includes the two new contract tests.
- [x] WASM rebuild: `wasm-pack build crates/wasm-bindings --target web
      --out-dir <abs>` — exit 0, "Done in 1m 50s". No browser eyeball:
      this PR places no entities and changes no engine code.
- [x] **Byte-reproducibility, proven not asserted — SEVEN runs.** Two
      bless runs, three re-blesses forced by provenance-hash changes, and
      two `check` runs, pin restored between every one. **The 160 cells
      were byte-identical every time**; the only line that ever changed
      was `zone_cache_hash`, and only when the hash function itself
      changed.
- [x] **Discrimination checks EXECUTED, not reasoned**, four of them:
      (1) swapping the `SelectionStage` tags at
      `decomposition_search.rs:1858/1860` fails both new stage pins with
      the right cause named, while the tier-1 pin stays green;
      (2) blessing against a deliberately wrong
      `SPAGHETTIO_ZONE_CACHE_PATH` exits 101 and leaves the baseline file
      untouched; (3) a `check` run where every cell matched but the cache
      identity differed now exits 101 instead of printing green;
      (4) restoring each mutation re-verified green.
- [x] Zone cache pinned for every measurement run and `git checkout --`'d
      after each of the nine; it appears in no commit.
- [x] Junction-seed census re-run after moving both hard asserts below the
      dumps: headline unchanged, 111 seeds across 12 fixtures, 0 in the
      necessary-superset bucket.

## Review

**Six rounds absorbed** (see the thread and the RFC decision log). Four
findings were load-bearing and all four were about the **provenance
machinery or the instrument's failure messages**, not the corpus:
SHA-256 stability over `DefaultHasher`; two *different* wrong zone-source
lists (first too narrow — the legacy `.jsonl`; then too wide — I absorbed
round 3 and hashed a `#[cfg(target_arch = "wasm32")]`-only file the native
run never reads); the `None == None` provenance hole; and the
order-coupling caveat that the corpus reproduces as a whole run rather
than per cell, which is a direct trap for Phase 2a's first instinct on a
divergence.

Two of those existed only because an earlier round was absorbed too
eagerly. That is worth recording as much as the fixes are.

Refuted with receipts: the CI-gate finding (five raisings — the answer is
scheduling, not cost: a gate on this baseline today asserts only "production
has not changed"), the "make the pins looser" suggestion (winner and stage
move together on both fixtures, so there is no looser pin that still
detects a mis-tagged stage), and several hallucinated identifiers
(`NESTED_ANYWHERE`, `most_recent_selection`, `read_baseline`, a misspelled
`EXPECTED_ORDER`) — all grep to zero.

## Deviations from intent

- The brief's minimum option set was four; this ships **five**. The
  `e2e-harness` set is what found the second fossil, and dropping it
  would have made the corpus mis-describe the harness configuration.
- The dispatch premise "ec30 ≈99.4% vs meter 0.000" was
  **falsified** — see the sim section. The 99.4% belongs to a different
  fixture *and* a different artifact. The comparison could never have been
  made; recording the correction is the deliverable, not the reconciliation.
- The first `ac5` sim came back `OVERALL: NO DATA` on a research-productivity
  kit error and was re-run with the axis declared. Both runs are reported;
  the invalid one is not used for any claim.

## Models / contracts touched

- `docs/rfc-070-one-selection-loop.md` — the named parity corpus and the
  divergence-equivalence rule are new **contract** text under Verification
  plan; four decision-log entries.
- No validator severity/category/selection change, so
  `docs/validator-trust.md` is untouched (correctly — this PR changes no
  check).
- `docs/status.md` — new open layout-quality entry for
  the ec30 defect, and an in-place correction to the "AC is bit-identical
  declared-or-not" claim (2134 vs 2125, measured today).


storkme:
## Review round 1 adjudicated — 3 absorbed (one half), 2 refuted

Pushed as `diag(w1c): absorb #694 review`.

### Absorbed

**Finding 1 (`bless` overwrites with no guard) — half absorbed.** Refusing
to bless when *verdicts* differ would make `bless` unusable: re-taking the
baseline after an intentional engine change is exactly what it is for, and
the RFC's adjudication/sim-anchor obligations are on the *shadow loop's*
divergences at Phase 2a, not on a human re-taking a Phase-0 record.

The **provenance** half is real and is now fixed. The corpus is
cache-relative, and `bless` rewrites the recorded `zone_cache_hash` too — so
a mis-pinned bless leaves 160 unreproducible rows with nothing left to
notice it by. `bless` now refuses when this run's cache differs from the
committed baseline's, or when no cache resolves at all; `bless-repin` is the
named escape hatch.

Discrimination check executed rather than reasoned:

```
$ SPAGHETTIO_ZONE_CACHE_PATH=<a wrong file> SPAGHETTIO_PARITY_CORPUS=bless cargo test ...
EXIT:101
refusing to bless against a DIFFERENT zone cache than the committed baseline's:
baseline Some("929b8d28938ed330"), this run Some("dbd10605090e23b0"). ...
$ git status --short   # parity_corpus_baseline.json untouched
```

**Finding 4 (`check` prints NOTE then hard-fails) — absorbed.** The
mismatch now *prefixes* the failure message instead of trailing it as a
NOTE the reader met before the diff list and had forgotten by the time it
mattered; the success line carries it too. The comparison still runs — a
clean result under a different cache is worth knowing, and refusing to
compare would recreate the "compared nothing" hole #693 closed.

**Finding 3 (documented status vocabulary never produced) — absorbed, and
it was the good catch.** The doc listed a `refused` no branch emits. That
is precisely the records-outlive-state class this repo keeps hitting. The
doc now enumerates the five statuses the code can produce and marks
`decided-then-refused` as *unreached on the current corpus* (zero cells in
the baseline) rather than implying it is routine.

**Nit (stale test count) — absorbed**, though the numbers were off in the
other direction: the module doc said THREE tests with one non-ignored; it
is now FIVE with three.

### Refuted, with receipts

**Finding 2 — "no CI gate."** Deliberate, and the reason is *scheduling*,
not cost. A CI gate on this baseline today would assert only "production
has not changed" — which every engine PR legitimately falsifies. That is a
re-bless treadmill with no reader, and it is the failure mode
`docs/validator-reporting.md` warns about from the other direction: a check
that fires on everything discriminates nothing.

The consumer that makes a gate *mean* something is Phase 2a's shadow loop,
and RFC-070 already commits to gating exactly that — Verification plan item
2, *"the v2 loop runs beside production on the named parity corpus; winner
mismatch fails the check."* A shadow-vs-production mismatch is a real
finding; a production-vs-yesterday mismatch is a Tuesday.

Meanwhile the split the finding misses: the **instrument** is gated (three
non-ignored contract tests in `check_firing_census.rs`, which is why this
PR extends them from one deciding stage to three), and the **data** is
guarded by the next phase re-taking it. The reasoning is now written into
the module doc so it reads as a decision rather than an omission.

**Finding 5 — "the new helpers assert `decided.len()==1`; migrate them to
the corpus's nesting-robust extractor."** That assertion *is* the pin.
`assert_scoreboard_contract` is deliberately brittle — its failure message
names both readings ("the ENGINE legitimately nested a selection" vs "the
INSTRUMENTATION broke") and says how to re-point it — because a nested
selection appearing on these fixtures is a change someone must look at.
`outer_selection` is deliberately robust because it must record 160 cells
without aborting on any one of them. Different jobs; making the pin robust
would delete the detection it exists for.

The discrimination check for those two pins is in the branch: swapping the
`SelectionStage` tags at `decomposition_search.rs:1858/1860` fails both new
tests naming the right cause while `selection_scoreboard_contract` stays
green.

**Nit — "`EXPECTED_ORDER` is misspelled / the docstring says 'second
hand-maintained longhand copy'."** The review retracts the misspelling
mid-sentence, and that phrase appears nowhere in the file
(`grep -c "hand-maintained longhand" crates/core/tests/parity_corpus.rs` → 0).

### Re-verification after the changes

- `SPAGHETTIO_PARITY_CORPUS=check` over the committed baseline: **160 cells,
  0 divergences, exit 0**.
- The three contract tests: green.
- `cargo clippy` on both touched test targets with `-D warnings`: clean.

